PR #21, opened 2016-08-18
External Memcached server support + various enhancement

fvillain:
Hello,

I made some modification to use external Memcached instance

- Support for an external Memcached server
   - Added /entrypoint.sh to save docker environment variable into a file to be loaded by systemd services via EnvironmentFile parameter as it does not do it
   - Modified appliance-initialize.sh script to allow support for Dockerfile (PR to come)

- Various fix in MIQ dockerfile 
  - remove packages installation for postgresql server part (not needed i guess) : rh-postgresql95-postgresql-server rh-postgresql95-postgresql-pglogical-output rh-postgresql95-postgresql-pglogical
  - add package installation : rh-postgresql95-libs 
  - remove VOLUME declaration (for postgresql) 

Let me know your thought on this


fbladilo:
@fvillain will take a look soon and post feedback.. thanks!

fbladilo:
@fvillain Could you check yaml syntax on template, it looks like there some extra action going around line 183 : 

[root@franco-ose-eval templates (master)]# oc create -f miq-template.yaml 
yaml: line 183: did not find expected key
...
184:9     error    syntax error: expected <block end>, but found '<block mapping start>'
...
Also, the volume declaration on memcached DeploymentConfig is not necessary.

fbladilo:
@fvillain There are a few issues with the upstream memcached image we discussed internally, since this solution will eventually merged upstream and potentially drip downstream, we prefer to go with a Centos7 based memcached image (same as appliance), this is also the same path taken with the Postgres 9.5 image, this will yield good consistency on the base images, unfortunately the current official DH memcached is Debian based. 

Secondly, it looks like the latest DH memcached images were built with Docker-1.10 and OSE3.2 does not support v2 image metadata, we are forced to rebuild on docker-1.9, I will get started with the new miq-memcached image soon for testing, it should only take minimal changes on the template once this is done.

fvillain:
@fbladilo 
I removed the syntax issue plus pvc for memcached that indeed was useless.
I haven't access to an openshift cluster at the moment, so I cannot test it though

On memcached image : let me know if you need help on that part, or I can work on it on my own and push it here too.

fvillain:
@fbladilo also maybe we can re-use this image : https://hub.docker.com/r/ausnimbus/memcached-14-centos7/~/dockerfile/ ?

Seems to do what we want and is based on an official openshift/centos7 image, what do you think ?

fbladilo:
@fvillain Thanks, I'll be re-importing the template and testing again tomorrow morning. Probably more comments to come but looks good on paper.

On memcached : Good catch! :smiley:  , I took the image and simplified a few things, I re-built it and looks good, does everything we need at the moment.

images/miq-app/ose-assets/appliance-initialize.sh
```diff
@@ -1,15 +1,39 @@
...
 
+set -x
+
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 
 # Source original OSE environment, generated by lifecycle hook script.
 
 [[ -s ${APP_ROOT}/tmp/ose-env ]] && source ${APP_ROOT}/tmp/ose-env
 
+if [[ -z $DATABASE_SERVICE_NAME ]]; then
```
fbladilo: @fvillain These should never come empty if deploying from template, since they have default parameters injected.
As long as the container environment is passed correctly to systemd, we should not have to check further. Did I miss anything?

images/miq-app/ose-assets/appliance-initialize.sh
```diff
@@ -1,15 +1,39 @@
 #!/bin/sh
 
+set -x
+
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 
 # Source original OSE environment, generated by lifecycle hook script.
 
 [[ -s ${APP_ROOT}/tmp/ose-env ]] && source ${APP_ROOT}/tmp/ose-env
```
fbladilo: I don't think we need this one anymore, since EnvironmentFile=/container.env.vars is being passed with all required environment.

fvillain: Yes, the set -x is not needed

images/miq-app/ose-assets/appliance-initialize.sh
fvillain: This is for docker-compose support in case environment parameters are not provided correctly

docker-compose.yml
```diff
@@ -0,0 +1,40 @@
+version: '2'
```
fbladilo: @fvillain Could we put the docker-compose as a different PR? Maybe as a contrib/docker-compose.yml would make sense since the project mostly focuses on OpenShift.

fvillain:
@fbladilo made the modification you asked for
we still have to see what to do about the env variables check in appliance-initialize.sh

images/miq-app/ose-assets/appliance-initialize.service
```diff
@@ -1,13 +1,13 @@
 [Unit]
 Description=Initialize Appliance Database
-After=evminit.service memcached.service
+After=evminit.service
 Before=evmserverd.service
-Requires=memcached.service
 [Service]
 TimeoutStartSec=10m
 Type=oneshot
+EnvironmentFile=/container.env.vars
```
fbladilo: @fvillain I like this, now that we supply the container environment to /bin/appliance-initialize.sh , we should be able to get rid of ${APP_ROOT}/tmp/ose-env, it seems more elegant. The only reason that ose-env existed was to ensure systemd was importing the correct OSE environment variables before running the database preparation.

fvillain: @fbladilo ok so we remove ose-dump-env.sh and references to it ?

fbladilo: @fvillain I don't think is needed anymore, in addition this approach is more elegant and friendly with systemd. When you finish I go testing again and we will see what happens :smiley:

fvillain: @fbladilo i strongly agree, let's do it

fbladilo: @fvillain In regards of env variables check in appliance-initialize.sh, if those checks are to support docker-compose as a contrib, we can leave them in place and perhaps put a comment on appliance-initialize.sh to clarify why.

fvillain:
@fbladilo we souhld be all good now

fbladilo:
@fvillain One more, remove the lifecycle hook on miq-template calling ose-dump and we should be good to test again!

fvillain:
@fbladilo didn't see that one , my bad

fbladilo:
@fvillain :+1: will go to test shortly...

fbladilo:
@fvillain I made some minor image adjustments but it came up successfully, good enough to merge. I will send another cleanup commit afterwards for those.

PR #31, opened 2016-09-13
[WIP] Initial commit for MIQ upgrade infrastructure

fbladilo:


images/miq-app/ose-assets/appliance-initialize.sh
```diff
@@ -2,14 +2,47 @@
...
-# Replace memcache host in EVM configuration
-sed -i.bak -E "s/:memcache_server:.*/:memcache_server: ${!MEMCACHED_SVC_HOST}:11211/gi" ${APP_ROOT}/config/settings.yml
+# Check deployment status
+check_deployment_status
 
-echo "Initializing Appliance, please wait ..."
-appliance_console_cli --region ${DATABASE_REGION} --hostname ${!DATABASE_SVC_HOST} --username ${POSTGRESQL_USER} --password ${POSTGRESQL_PASSWORD}
+if [ -n "${UPGRADE}" ]; then
+  echo "== Starting Upgrade =="
+  restore_pv_data
```
bazulay: At this point we need to have a place holder for a pre_upgrade_hook

images/miq-app/ose-assets/appliance-initialize.sh
```diff
@@ -2,14 +2,47 @@
...
+fi
+
+if [ -n "${REDEPLOY}" ]; then
+  echo "== Starting Re-deployment =="
+  restore_pv_data
+  setup_memcached
+  migrate_db
+fi
+
+if [ -n "${NEW_DEPLOYMENT}" ]; then
```
bazulay: I would use the same variable for the different statuses ...
lets say DEPLOYMENT_STATUS is set in function check_deployment_status to one of 3 values:
"NEW_DEPLYMENT" , "REDEPLOY" & "UPGRADE"

+ there should be 'else' statement between those conditions

images/miq-app/ose-assets/appliance-initialize.sh
fbladilo: Makes sense, I can add a pre_upgrade_hook function that checks for the existance of a script file to run prior upgrading.

images/miq-app/ose-assets/appliance-initialize.sh
fbladilo: DEPLOYMENT_STATUS sounds good, I can actually use a bash "case" instead of multiple if statements, check the value of DEPLOYMENT_STATUS status and proceed accordingly, a default error message can be printed if for any reason DEPLOYMENT_STATUS does not match any (should never happen but JIC).

images/miq-app/ose-assets/ose-deploy-common.sh
```diff
@@ -0,0 +1,217 @@
...
+  # Check if we have identical EVM versions (exclude master builds)
+  if [[ $APP_VERSION == $PV_APP_VERSION && $APP_VERSION != master ]]; then
+    echo "== App version matches original deployment =="
+  # Check if we have same schema version for same EVM version
+    if [ "${SCHEMA_VERSION}" != "${PV_SCHEMA_VERSION}" ]; then
+       echo "ERROR: Something seems wrong, db schema version mismatch for the same app version: ${PV_SCHEMA_VERSION} <-> ${SCHEMA_VERSION}"
+       exit 1
+    fi
+  # Assuming redeployment (same APP_VERSION)
+  export DEPLOYMENT_STATUS=redeployment
```
bazulay: Indentation ...

images/miq-app/ose-assets/ose-deploy-common.sh
```diff
@@ -0,0 +1,217 @@
...
+  else
+  # Assuming upgrade (different APP_VERSION)
+  export DEPLOYMENT_STATUS=upgrade
```
bazulay: It is upgrade only when AP_VERSION > PV_APP_VERSION
Meaning this may hide a problem

fbladilo: So far is an upgrade when APP_VERSION != PV_APP_VERSION, we need run extended checks for UPGRADE to ensure we are not upgrading to an older APP_VERSION but given the versioning situation between upstream/downsteam, need to discuss how to handle non-numeric vs numeric...

images/miq-app/ose-assets/ose-deploy-common.sh
fbladilo: Will fix :smiley:

PR #32, opened 2016-09-20
add app pv instructions and yaml

pweil-:
The manageiq template [requires an application volume](https://github.com/fbladilo/miq-on-openshift/blob/master/templates/miq-template.yaml#L60-L69) in addition to the postgres volume but does not include the creation of the volume in the instructions.

This adds a volume for manageiq to use and updates the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,11 +61,13 @@ _**As admin:**_
 
 ```bash
 $ oc create -f miq-pv-example.yaml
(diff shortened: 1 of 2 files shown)


fbladilo:
@pweil- Paul take a look at the current WIP for Upgrade Infrastructure, it includes the mentioned PR for miq-application along with necessary settings on the deployment config.

PR #43, opened 2016-09-30
Re-work run_hook function to accept a script name argument

fbladilo:
- run_hook now accepts a script name argument, it is meant to be executed at any stage of the deployment
- two default entries were added for the upgrade procedure pre-upgrade and post-upgrade
- if a script is not found, run_hook returns 1 and the deployment continues
- if a script is found but exits abnormally, the deployment fails and exit 1

### DIFF
--- a/images/miq-app/docker-assets/appliance-initialize.sh
+++ b/images/miq-app/docker-assets/appliance-initialize.sh
@@ -22,10 +22,12 @@ case "${DEPLOYMENT_STATUS}" in
   upgrade)
   echo "== Starting Upgrade =="
   backup_pv_data
+  run_hook pre-upgrade
   restore_pv_data
   pre_upgrade_hook
   setup_memcached
   migrate_db
+  run_hook post-upgrade
   ;;
   new_deployment)
   echo "== Starting New Deployment =="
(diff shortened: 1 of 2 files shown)


fbladilo:
@bazulay

Hook script function reworked, it takes one argument pointing to the script to run, below is an example run output : 

```
== Found existing deployment configuration ==
== Restoring existing database configuration ==
== Starting Upgrade ==
== Initializing PV data backup ==
== Calling Deployment Hook ==
Hook script pre-upgrade not found, skipping
== Restoring PV data symlinks ==
/var/www/miq/vmdb/REGION symlink is already in place, skipping
/var/www/miq/vmdb/GUID symlink is already in place, skipping
/var/www/miq/vmdb/config/database.yml symlink is already in place, skipping
/var/www/miq/vmdb/certs/v2_key symlink is already in place, skipping
/var/www/miq/vmdb/log symlink is already in place, skipping
== Applying memcached config ==
== Migrating Database ==
== Calling Deployment Hook ==
== Running /opt/manageiq/container-scripts/post-upgrade ==
Do something in post-upgrade hook here

```

PR #44, opened 2016-09-30
Add APPLICATION_INIT_DELAY as a template parameter and minor fixes

fbladilo:
- Default APPLICATION_INIT_DELAY parameter is 30 seconds to allow rest of services to settle
- /bin/appliance-initialize.sh now calls sleep instead of appliance-initialize systemd service unit file
- Removed pre_upgrade_hook call from /bin/appliance-initialize.sh (function no longer valid)


fbladilo:
@bazulay

PR #51, opened 2016-10-11
Added check_version_gt function

fbladilo:
- check_version_gt will ensure that the upgrade version is actually newer than the PV recorded version
- It supports numeric and alphanumeric versioning (upstream/downstream)
- If the upgrade version is actually older than the PV version, it will abort the upgrade

### DIFF
--- a/images/miq-app/docker-assets/container-deploy-common.sh
+++ b/images/miq-app/docker-assets/container-deploy-common.sh
@@ -47,7 +47,7 @@ if [[ -f ${APP_ROOT_PERSISTENT_VMDB}/config/database.yml && -f ${PV_DEPLOY_INFO_
   APP_VERSION="$(cat ${APP_ROOT}/VERSION)"
   SCHEMA_VERSION="$(cd ${APP_ROOT} && RAILS_USE_MEMORY_STORE=true bin/rake db:version | awk '{ print $3 }')"
   # Check if we have identical EVM versions (exclude master builds)
-  if [[ $APP_VERSION == $PV_APP_VERSION && $APP_VERSION != master ]]; then
+  if [[ ${APP_VERSION} == ${PV_APP_VERSION} && ${APP_VERSION} != master ]]; then
     echo "== App version matches original deployment =="
   # Check if we have same schema version for same EVM version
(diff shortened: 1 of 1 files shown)


fbladilo:
@bazulay Please see some simulated examples, I modified versions while running the MIQ application pod to test the procedure : 
```
[root@manageiq-1-al9dq vmdb]# /bin/appliance-initialize.sh 
== Found existing deployment configuration ==
== Restoring existing database configuration ==
ERROR: Upgrade version 5.5.1.2 is older than PV version 5.6.2.1, aborting upgrade..

[root@manageiq-1-al9dq vmdb]# /bin/appliance-initialize.sh 
== Found existing deployment configuration ==
== Restoring existing database configuration ==
ERROR: Upgrade version darga is older than PV version euwe, aborting upgrade..
```

images/miq-app/docker-assets/container-deploy-common.sh
```diff
@@ -67,6 +70,22 @@ fi
 
 }
 
+function check_version_gt() { 
+# Check if upgrade version is actually greater than stored PV version
+# -V sorts alphanumeric versions within text, will always return oldest version first
+# Compare sort version result against upgrade version
+
```
bazulay: a preliminary check can/should  be added here in  order to make sure we compare upstream with upstream and downstrem with downstream.
it could be done by checking the delimiter ('.' vs '-' or '').
although far fetched it will revile the extreme case  that is very hard to debug.

bazulay: In second thought this is so far fetched so lets just leave it as is.
(in order to make it happen one must either edit the template or build a new image...)

images/miq-app/docker-assets/container-deploy-common.sh
```diff
@@ -67,6 +70,22 @@ fi
...
+if [[ "$(echo "$@" | tr " " "\n" | sort -V | head -n 1)" != "$1" ]]; then
```
carbonin: Ah I see. I missed the `$()` enclosing the string of commands. Sorry for the noise :smile:

bazulay:
LGTM

PR #53, opened 2016-10-11
Add hook sample scripts

fbladilo:
@bazulay @carbonin Gents, please review, these are sample hook scripts, we will implement logic once the use cases arise.


bazulay:
LGTM

PR #56, opened 2016-10-19
Increase miq-app deployment strategy timeout

fbladilo:
- Increased timeout to 1200 seconds on miq-app deployment strategy
- Increase liveness check timeout to 480 seconds (8mins)

This prevents issues on first time deployment where the miq-app image
is being downloaded and the default recreate strategy timeout of 600 seconds
is not long enough for a complete a successful deployment.

Tested on OCP 3.3

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -94,7 +94,7 @@ objects:
             httpGet:
               path: /
               port: 80
-            initialDelaySeconds: 300
+            initialDelaySeconds: 480
             timeoutSeconds: 3
           readinessProbe:
             httpGet:
@@ -156,6 +156,8 @@ objects:
             name: "miq-app:latest"
     strategy:
       type: "Recreate"
+      recreateParams:
+        timeoutSeconds: 1200
 - apiVersion: v1
   kind: "Service"
   metadata:


fbladilo:
@bazulay Please review, I had some timing issues while deploying miq-app from scratch on OCP 3.3, our default strategy deployment timeout of 600 seconds was too short to successfully finish a deployment since the image size is ~1.3GB and takes at least a few minutes to download.

To recap the logic and timers used during the miq-app deployment, we have : 

1. Re-create strategy timeout on deployment config to 1200 seconds (20 minutes)
If a pod is unable to finish deployment in 20 minutes, it will go on error state.

2. Liveness check Initial delay timeout of 480 seconds (8 minutes)
If a pod fails to pass liveness checks after 8 minutes of running, the pod will be terminated and RestartPolicy settings will apply, in our case the pod will be restarted.

3. Readiness check Initial delay of 200 seconds
The default readiness state before the Initial delay is always failure, we will start checking for readiness after 200 seconds, once a successful check is performed, the end-points controller will begin passing traffic to the pod.

Keep in mind that a typical first time deployment once the images have been downloaded takes around 5 minutes total, DB preparation and seeding being the most time consuming tasks.

Redeployments and upgrades usually take less time, 3 minutes on average. That's the reason for the readiness Initialdelay to be around 200 seconds.

bazulay:
The timeouts sound reasonable.
LGTM

PR #57, opened 2016-10-19
Rename rh-repmgr95 to rh-postgresql95-repmgr

simaishi:
The RPM is now hosted in a different repo.
(https://github.com/ManageIQ/manageiq/pull/11951)


fbladilo:
Good catch thanks @simaishi :smiley:

PR #58, opened 2016-10-19
Remove npm updating itself

simaishi:
With new npm (v3.10) in EPEL, updating itself is no longer needed
and 'npm install' no longer works after npm itself is updated

(https://github.com/ManageIQ/manageiq/pull/11849)


fbladilo:
LGTM :+1: Thanks @simaishi

PR #61, opened 2016-10-25
Remove rh-repmgr95 from miq-app image

fbladilo:


fbladilo:
@bazulay @simaishi Nick confirmed is not necessary on miq-app image

bazulay:
@carbonin @simaishi @fbladilo 
Than I don't understand why was it there ? can we get a more detailed explanation ?

carbonin:
I'm assuming the list of packages came from the monolithic image and it was just never removed. In fact I would think all the postgres packages could be removed from the app image, right?

simaishi:
That's my assumption as well - the original list came from monolithic. I think `rh-postgresql95-postgresql-devel` package can also be removed, but not sure about the `-libs`.  Doesn't 'pg' need it?

carbonin:
Right, I didn't see that the server packages were already removed. I would leave those two. But https://github.com/ManageIQ/manageiq-pods/pull/61/files#diff-c1f1f7358bf9afe79fd2c0e7731301a6L18 can also be removed as we are not installing anything from that repo anymore.

bazulay:
@carbonin @fbladilo @simaishi 
than let's go over the package list again and make sure we are not installing redundant packages.

fbladilo:
@bazulay @carbonin @simaishi The PG gem would at least need want rh-postgresql95-postgresql-devel in order to build properly, possibly also the runtime libs rh-postgresql95-postgresql-libs, I would need to inspect/test in order to understand also their yum dependencies as they install (one may pull more). 
The ncarboni repo layer could be removed I believe, as long as we are not pulling/needing anything pglogical on the miq-app image.

Also, it would ideal to to have the psql binary installed, as we want to be able to test/verify PGSQL connectivity to the external DB, we are not using it now during initialization but I could foresee how this can change, I thought about writing a loop function to verify connectivity to the database POD prior trying to initialize many times, inspired by : 

https://github.com/ManageIQ/manageiq/blob/master/gems/pending/appliance_console/internal_database_configuration.rb#L120

fbladilo:
@bazulay @carbonin @simaishi updated PR to remove ncarboni repo as it is safe to assume we will not need pglogical RPMs on miq-app. The PG gem will need -devel, -libs and which will pull -server as a dependency.

simaishi:
LGTM

bazulay:
LGTM

PR #63, opened 2016-10-26
Add check_db_status function to container-deploy-common

fbladilo:
- Added check_db_status to ensure readiness of postgresql POD before trying to initialize
- Function uses PG provided pg_isready to perform connectivity tests
- Database hostname is supplied by OpenShift template parameters

### DIFF
--- a/images/miq-app/docker-assets/appliance-initialize.sh
+++ b/images/miq-app/docker-assets/appliance-initialize.sh
@@ -11,6 +11,9 @@ sleep "${APPLICATION_INIT_DELAY}"
 # Prepare initialization environment
 prepare_init_env
 
+# Check DB readiness
+check_db_status
+
 # Check deployment status
 check_deployment_status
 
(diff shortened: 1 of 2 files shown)


fbladilo:
@simaishi @bazulay @carbonin Please review, this will *ensure* the postgresql POD is online and servicing before we make any attempts to interact with it. I have encountered issues when the postgresql pod takes unusually long to come online or has come online but there are SDN issues, this will help miq-app to be more tolerant of those situations.

It relies on pg_isready which comes with the rh-postgresql95-postgresql RPM.

Sequence of events is as follows : 

1. /bin/appliance-initialize.sh will call check_db_status for PG availability every 5 seconds, if sucessful it will break and continue with initialization
2. If the PG pod never comes online (whatever the reason might be), the liveness probe will take execution after 8 minutes and axe the current miq-app POD and restart it.
3. OpenShift will continue to attempt to deploy miq-app for a 20 minute cycle (deploy strategy timeout), if we reach the timeout and deployment is not ready, it will shutdown the POD and put it in Error state.

bazulay:
LGTM clean & simple

PR #65, opened 2016-10-27
Remove stale miq-monolithic image from repo

fbladilo:
- The miq-monolithic image is deprecated and stale, remove from repo along with assets


fbladilo:
@simaishi @bazulay This cleans up the old miq-monolithic image, which has been deprecated. We use the monolithic image on the main MIQ repo instead.

PR #70, opened 2016-11-17
Parametize image tags on template

fbladilo:
- Added image tag paremeters for all pods deployed, allows a user to request specific tags on deployment

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -88,9 +88,8 @@ objects:
             persistentVolumeClaim:
               claimName: ${NAME}
         containers:
-        - image: fbladilo/miq-app:${APPLICATION_IMG_TAG}
+        - name: manageiq
           imagePullPolicy: IfNotPresent
-          name: manageiq
           livenessProbe:
             httpGet:
               path: /
@@ -223,7 +222,6 @@ objects:
         containers:
           -
             name: "memcached"
-            image: "fbladilo/miq-memcached:${MEMCACHED_IMG_TAG}"
             ports:
               -
                 containerPort: 11211
@@ -311,7 +309,6 @@ objects:
         containers:
           -
             name: "postgresql"
-            image: "fbladilo/miq-postgresql:${POSTGRESQL_IMG_TAG}"
             ports:
               -
                 containerPort: 5432
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -88,7 +88,7 @@ objects:
             persistentVolumeClaim:
               claimName: ${NAME}
         containers:
-        - image: fbladilo/miq-app:latest
+        - image: fbladilo/miq-app:${APPLICATION_IMG_TAG}
           imagePullPolicy: IfNotPresent
           name: manageiq
           livenessProbe:
@@ -162,7 +162,7 @@ objects:
             - "manageiq"
           from:
             kind: "ImageStreamTag"
-            name: "miq-app:latest"
+            name: "miq-app:${APPLICATION_IMG_TAG}"
     strategy:
       type: "Recreate"
       recreateParams:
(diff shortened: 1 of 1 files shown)


fbladilo:
@bazulay @simaishi Please review, the image tags are parametized, allowing a user to request a specific tag during deployment, defaults are set to latest tag upstream.

Example use : 
```bash
oc new-app --template=manageiq -p POSTGRESQL_IMG_TAG=9.5,MEMCACHED_IMG_TAG=1.4.15
```
This will allow users upstream to also request a stable build without editing the template since latest is master upstream.

```bash
oc new-app --template=manageiq -p APPLICATION_IMG_TAG=euwe
```

bazulay:
LGTM

templates/miq-template.yaml
```diff
@@ -88,7 +88,7 @@ objects:
             persistentVolumeClaim:
               claimName: ${NAME}
         containers:
-        - image: fbladilo/miq-app:latest
+        - image: fbladilo/miq-app:${APPLICATION_IMG_TAG}
```
simaishi: @fbladilo Not really related to the change itself... but where does this value get used?  It seems the value used in the actual Deployments screen is what comes from the ImageStream 'name' and I can't seem to find where this is actually used.

fbladilo: @simaishi I'm not sure, at a glance it would seem like it's redundant with ImageStream 'name' but I need to test in order to ensure.

fbladilo: @simaishi I tested without it, but OCP does not like it if the 'image' key is missing from container : 
```bash
[root@franco-ocp-eval-master tmp]# oc new-app --template=manageiq 
error: error processing template manageiq: only encoded map or array can be decoded into a struct
```
However , an empty "image" key seems to work fine : 
```bash
[root@franco-ocp-eval-master tmp]# oc export templates manageiq | grep "image:"
          image: ""
          image: fbladilo/miq-memcached:1.4.15
          image: fbladilo/miq-postgresql:9.5
[root@franco-ocp-eval-master tmp]# oc replace -f miq-template.yaml 
template "manageiq" replaced
[root@franco-ocp-eval-master tmp]# oc new-app --template=manageiq 
--> Deploying template manageiq

     manageiq
     ---------
     ManageIQ appliance with persistent storage
```
If it sounds Ok with you I can go ahead and empty all those image keys to avoid confusion. The DC is actually replacing them with the ImageStream tag requested.

simaishi: Anything that won't be confusing (empty or just name 'miq-app') works for me 😄

fbladilo: @simaishi I removed the image key altogether for the 3 pod templates, we rely on the deployment config Image params to call the correct image from the ImageStream here : 

https://github.com/fbladilo/manageiq-pods/blob/master/templates/miq-template.yaml#L160

I tested without the container spec image key defined and worked as expected, the DC populated the pod template image key with the requested image/tag from IS : 
```bash
[root@franco-ocp-eval-master templates (template_add_version_tag_param)]# oc describe pods manageiq-2-5725m | grep Image
    Image:		docker.io/fbladilo/miq-app@sha256:395eb0dd8306b1d3c30ab0de884b7a4091746ab8c4641be4156b839f404065a9
    Image ID:		docker://sha256:c32c91aec7a6d77cb3aca608d987e4e42deb51dd4b1a61481df3a515abf8ceb1
```

simaishi:
This change doesn't work in Origin environment... discussed with @fbladilo, and another change is coming..

fbladilo:
@simaishi Regressed changes for origin compatibility..

simaishi:
LGTM

simaishi:
Euwe Backport details:
```$ git log -1
commit 7b1edb4a0f711cf02127d791724a8e8a78bd2087
Author: Satoe Imaishi <simaishi@redhat.com>
Date:   Thu Dec 1 11:06:52 2016 -0500

    Merge pull request #70 from fbladilo/template_add_version_tag_param
    
    Parametize image tags on template
    (cherry picked from commit ba1b72d59b2753c168346ab7d9f9189795d5b6b4)
```

PR #73, opened 2016-12-01
Use manageiq/manageiq-pods images

simaishi:
Use manageiq/manageiq-pods images built in Docker Hub.

All 3 images come from 'manageiq-pods' and each image will have different tag which contains the name of the image ('app', 'memcached' or 'postgresql').

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -44,11 +44,11 @@ objects:
 - apiVersion: v1
   kind: ImageStream
   metadata:
-    name: miq-app
+    name: miq-images
     annotations:
-      description: "Keeps track of changes in the ManageIQ app image"
+      description: "Keeps track of the ManageIQ image changes"
   spec:
-    dockerImageRepository: docker.io/fbladilo/miq-app
+    dockerImageRepository: docker.io/manageiq/manageiq-pods
 - apiVersion: v1
   kind: PersistentVolumeClaim
   metadata:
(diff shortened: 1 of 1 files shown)


simaishi:
@fbladilo please review - tested in Origin environment.

templates/miq-template.yaml
```diff
@@ -272,7 +272,7 @@ objects:
     annotations:
       description: "Keeps track of changes in the ManageIQ postgresql image"
   spec:
-    dockerImageRepository: docker.io/fbladilo/miq-postgresql
+    dockerImageRepository: docker.io/manageiq/manageiq-pods
```
fbladilo: @simaishi I don't think we need to keep track of three different ImageStreams pointing to the same docker repository. We could should use a single "miq-images" IS and reference that across the template. We moved away from three different repos and we now call the specific images for pods by tags, not by image names anymore.

simaishi: Oh, that's true.  IS itself isn't tied to any image.  Will update - thanks!

fbladilo: @simaishi Correct for instance : 

```
from:
          kind: ImageStreamTag
          name: miq-images:postgresql-latest

- memcached
        from:
          kind: ImageStreamTag
          name: miq-images:memcached-latest
```

fbladilo:
@simaishi LGTM :+1:

chessbyte:
@simaishi Why is this labeled `euwe/conflict`?

simaishi:
I'd need to change some values in this PR to make it work for Euwe.  Will create a separate PR.

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq-pods/pull/75

PR #74, opened 2016-12-03
Ensure DB v2_key is always restored on upgrade/redeploy

fbladilo:
- Do not rely on v2_key present on image, always restore v2_key from persistent volume on redeploy/upgrade.

### DIFF
--- a/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
+++ b/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
@@ -41,6 +41,7 @@ if [[ -f ${APP_ROOT_PERSISTENT_VMDB}/config/database.yml && -f ${PV_DEPLOY_INFO_
   echo "== Found existing deployment configuration =="
   echo "== Restoring existing database configuration =="
   ln --backup -sn ${APP_ROOT_PERSISTENT_VMDB}/config/database.yml ${APP_ROOT}/config/database.yml
+  [[ ! -f ${APP_ROOT_PERSISTENT_VMDB}/certs/v2_key ]] && echo "ERROR: Could not find ${APP_ROOT_PERSISTENT_VMDB}/certs/v2_key on upgrade/redeploy case, aborting.." && exit 1
   ln --backup -sn ${APP_ROOT_PERSISTENT_VMDB}/certs/v2_key ${APP_ROOT}/certs/v2_key
   # Source original deployment info variables from PV
   source ${PV_DEPLOY_INFO_FILE}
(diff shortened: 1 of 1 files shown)


fbladilo:
@simaishi During upgrades and redeployment we need to ensure we restore the DB v2_key along with database.yml in check_deployment_status from the persistent volumes.
 
In the current implementation, check_deployment_status, when it executes an upgrade/redeployment case, relies on the DB v2_key that comes with the upstream image instead of ensuring is using the v2_key from the PV. This logic breaks redeployments/upgrades on builds that actually create a legitimate/unique v2_key during first deployment such as downstream. (Initial deployment works fine)

The root issue comes from the fact that we actually supply a dev v2_key during image building for upstream, (bin/setup layer). 

As a different PR, we need to address bin/setup in our container builds, this is a script for development purposes, we should not supply images with the v2_dev_key installed.

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -41,6 +41,7 @@ if [[ -f ${APP_ROOT_PERSISTENT_VMDB}/config/database.yml && -f ${PV_DEPLOY_INFO_
   echo "== Found existing deployment configuration =="
   echo "== Restoring existing database configuration =="
   ln --backup -sn ${APP_ROOT_PERSISTENT_VMDB}/config/database.yml ${APP_ROOT}/config/database.yml
+  ln --backup -sn ${APP_ROOT_PERSISTENT_VMDB}/certs/v2_key ${APP_ROOT}/certs/v2_key
```
fbladilo: Yes I can add a check for v2_key existance for safety, will update. No need to cleanup keys as we should not be creating or copying DB keys during build-time to begin with, ever.

fbladilo: @bazulay Added file existance check on PV before attempt to restore symlink

simaishi:
LGTM

simaishi:
Euwe Backport details:
```
git log -1
commit e5f065cf5afbe3a3fb952b80859016763e6e080a
Author: Satoe Imaishi <simaishi@redhat.com>
Date:   Tue Dec 6 09:47:42 2016 -0500

    Merge pull request #74 from fbladilo/restore_v2_key_on_upgrade_redeploy
    
    Ensure DB v2_key is always restored on upgrade/redeploy
    (cherry picked from commit 5f52e97f458beb5d8d367a179fce8be8ba904263)
```

PR #90, opened 2017-02-07
Parametrize the container image names

simon3z:
cc @fbladilo @bazulay

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -48,7 +48,7 @@ objects:
     annotations:
       description: "Keeps track of the ManageIQ image changes"
   spec:
-    dockerImageRepository: docker.io/manageiq/manageiq-pods
+    dockerImageRepository: "${IMAGES_REGISTRY}/${APPLICATION_IMG_NAME}"
 - apiVersion: v1
   kind: PersistentVolumeClaim
   metadata:
@@ -88,7 +88,7 @@ objects:
             persistentVolumeClaim:
               claimName: ${NAME}
         containers:
-        - image: manageiq/manageiq-pods:${APPLICATION_IMG_TAG}
+        - image: "${APPLICATION_IMG_NAME}:${APPLICATION_IMG_TAG}"
           imagePullPolicy: IfNotPresent
           name: manageiq
           livenessProbe:
@@ -215,7 +215,7 @@ objects:
         containers:
           -
             name: "memcached"
-            image: "manageiq/manageiq-pods:${MEMCACHED_IMG_TAG}"
+            image: "${MEMCACHED_IMG_NAME}:${MEMCACHED_IMG_TAG}"
             ports:
               -
                 containerPort: 11211
@@ -295,7 +295,7 @@ objects:
         containers:
           -
             name: "postgresql"
-            image: "manageiq/manageiq-pods:${POSTGRESQL_IMG_TAG}"
+            image: "${POSTGRESQL_IMG_NAME}:${POSTGRESQL_IMG_TAG}"
             ports:
               -
                 containerPort: 5432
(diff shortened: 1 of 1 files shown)


simon3z:
@fbladilo @bazulay should `dockerImageRepository` be parametrized too?

bazulay:
@simon3z  in fact this can be a good idea to reduce template changes for pre release phase.

@fbladilo  IIUC there is no doc change required as the default values will behave the same , right ?

fbladilo:
@simon3z @simaishi @bazulay Sounds good, my only observation is that we need to reference the new *_IMG_NAME within the miq-images ImageStream, that's how the DC will resolve the new image name and initiate the deployment. The config/imagechange triggers actually override whatever is set on the POD spec within the DC.

We will have to pick a single variable for the IS, perhaps APPLICATION_IMG_NAME which may be counter productive if for example the postgresql IMG was attempted to be pulled from a different repo but unless we implement multiple repos for each POD image I don't see too many options around it.

simon3z:
> I guess in summary, if you plan to change the IMAGE_NAME, ensure all three images are coming from the same docker repo and the IS will have no issues there. Another alternative is to use a single variable , something like IS_IMAGE_REPO that will allow you to change it for all three at the same time.

@fbladilo is that compatible with CloudForms? IIRC this wouldn't help in minimize the changes to be maintained there.

simaishi:
If we use *_IMG_NAME for each of the 3 pods in IS, that will be compatible with CloudForms.  But if we pick a single variable for IS, that will not be compatible.

fbladilo:
@simon3z We discussed this with @simaishi and we believe the best path of action is to have 3 different IS by pointing to the same repo/image by default (same as downstream), this will give the new IMG parameters full independence from each other. At the moment we use a single IS "miq-images", which we should split into : miq-app, miq-memcached, miq-postgresql. Do you want to update the PR or I can test/submit a new one with all these changes included?

simon3z:
@fbladilo @simaishi @bazulay I added the `IMAGES_REGISTRY` parameter to tweak the registry hostname.
ATM this new addition is untested (but straightforward).

simaishi:
@simon3z We still need to create 3 separate ImageStream, with each `dockerImageRepository` pointing to different repository:
```
${IMAGES_REGISTRY}/${APPLICATION_IMG_NAME}
${IMAGES_REGISTRY}/${MEMCACHED_IMG_NAME}
${IMAGES_REGISTRY}/${POSTGRESQL_IMG_NAME}
```
Otherwise it will not be compatible with downstream, nor you will be able to get images from different registries. For the upstream, the default values will be same for all 3.  For the downstream, the default will be different.

fbladilo:
@simon3z @simaishi Something like this will do : 
```yaml
- apiVersion: v1
  kind: ImageStream
  metadata:
    name: miq-app
    annotations:
      description: "Keeps track of the ManageIQ image changes"
  spec:
    dockerImageRepository: ${IMAGES_REGISTRY}/${APPLICATION_IMG_NAME}
- apiVersion: v1
  kind: ImageStream
  metadata:
    name: miq-postgresql
    annotations:
      description: "Keeps track of the PostgreSQL image changes"
  spec:
    dockerImageRepository: ${IMAGES_REGISTRY}/${POSTGRESQL_IMG_NAME}
- apiVersion: v1
  kind: ImageStream
  metadata:
    name: miq-memcached
    annotations:
      description: "Keeps track of the Memcached image changes"
  spec:
    dockerImageRepository: ${IMAGES_REGISTRY}/${MEMCACHED_IMG_NAME}
...
from:
            kind: "ImageStreamTag"
            name: "miq-app:${APPLICATION_IMG_TAG}"
...
  from:
            kind: "ImageStreamTag"
            name: "miq-memcached:${MEMCACHED_IMG_TAG}"
...
 from:
            kind: "ImageStreamTag"
            name: "miq-postgresql:${POSTGRESQL_IMG_TAG}"
```
Please keep in mind that there is a single image registry variable, if two images were coming from different registries we will still need more parameters to satisfy fully independent registries but I believe this covers most use cases.

simon3z:
@simaishi @fbladilo I sent this to parametrize what was available already (as the PR says) but I see that you want to piggyback for "features" :blush:  (support for different registries).

fbladilo:
@simon3z Fede I know you are busy, do you want me to submit the full PR with all changes?

simon3z:
> @simon3z Fede I know you are busy, do you want me to submit the full PR with all changes?

@fbladilo it's not clear what's wrong with this PR that can't be merged and requires you to re-send it entirely instead of adding the missing feature on top of it.

fbladilo:
@simon3z :+1:  I will go ahead merge the actual PR and send a related one with the IS work.

PR #95, opened 2017-02-16
Scale MIQ appliances using PetSets

fbladilo:
- Replaced MIQ template Deployments in favor of Petsets
- miq-app image Dockerfile updated to reflect a 2-volume deployment (region/server)
- Applied necessary changes to container-deploy-common.sh to handle the 2-volume configuration
- Created function to detect a new_replica deployment on container-deploy-common.sh
- Created function to handle REGION join for new replicas on container-deploy-common.sh
- Added logic on appliance-initialize.sh to call a new_replica case
- Updated NFS PV templates to reflect new storage configuration
- Updated all documentation to reflect new procedures and added "Scale" section

### DIFF
--- a/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
+++ b/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
@@ -303,8 +303,6 @@ rsync -qavL --exclude 'v2_key' --exclude 'database.yml' --exclude 'REGION' --fil
 [ ! -f "${PV_REGION_VMDB}/certs/v2_key" ] && rsync -qavR "${APP_ROOT}/certs/v2_key" "${PV_CONTAINER_DATA_REGION_DIR}"
 [ ! -f "${PV_REGION_VMDB}/REGION" ] && rsync -qavR "${APP_ROOT}/REGION" "${PV_CONTAINER_DATA_REGION_DIR}"
 
-backup_pv_data
-
 ) 2>&1 | tee "${PV_DATA_INIT_LOG}"
 
 }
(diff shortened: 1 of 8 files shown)


fbladilo:
@bazulay @simaishi Please review, the controller for miq-app was switched from DC to PetSets, this is the major change that allows scaling. Necessary template updates, docs and script logic was added.

fbladilo:
@bazulay Rebased, ready to review.

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -75,7 +78,7 @@ if [[ -f ${APP_ROOT_PERSISTENT_VMDB}/config/database.yml && -f ${PV_DEPLOY_INFO_
     fi
   fi
 else
-  echo "No pre-existing EVM configuration found on PV"
+  echo "No pre-existing EVM configuration found on region PV"
```
bazulay: Not sure this is enough ....
If region PV data is missing but the Host PV has the data (meaning some kind of error happened and we are in an unfamiliar status).
I think this needs to also be addressed (now that we separated the state into 2 PVs)

fbladilo: We could potentially add a directory existance check for PV_CONTAINER_DATA_DIR (server data PV dir), that should never exist if we are going into a new_deployment case as nothing has been provisioned yet. We already know REGION pv is empty/blank.

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -113,6 +116,29 @@ fi
...
 }
 
+function check_if_new_replica() {
+# Description
+# Check for pre-existing server data into PV, if not found, we assume a new server/replica case
+# Always skip if we are performing a new_deployment
+
+echo "== Checking for existing data on server PV =="
+
+if [[ ! -d ${PV_CONTAINER_DATA_DIR} && ${DEPLOYMENT_STATUS} != new_deployment ]]; then
```
bazulay: Is it enough to check the directory only ? (I think it is created on the initialization of the logs ...
Shouldn't we validate the GUID is there (and just make sure the directories are there ?)
The GUID is the part that determines the appliance personality ...

fbladilo: The initialization logs for the server-data directory would have never happened on a new_deployment case or new_replica. They happen afterwards see here : 

https://github.com/ManageIQ/manageiq-pods/blob/master/images/miq-app/docker-assets/appliance-initialize.sh#L41

Keep in mind appliance-initialize.sh dictates the order of events, check_deployment_status is just a function before rest of actions happen.

Also, please don't confuse with the server-deploy logs, which are a different directory.

fbladilo: The logic for check_if_new_replica is to ensure we are not a new_deployment case and that the server data PV is empty, both need evaluate true for us to consider this is legitimately a new_replica.

bazulay:
LGTM

Issue #93, opened 2017-02-15
images built from manageiq master should not trigger upgrade seaquence every reschedule


bazulay:
While this behavior is the right behavior for development env, this is not necessarily the right way to behave on a real live system.
On one live system  based on manageiq master, the system got into a reschedule storm , and as a result it run out of space on the PV since it is backing up the data every time it thinks it is in upgrade seaquence, and on master we have always assumed a reschedule = upgrade

optional solution:
The system should base the decision whether it is in upgrade seaquence  (on images built from master) based also on the also on the persisted revision (sha) of manageiq git repo

fbladilo:
@bazulay Yes, master has been always an exception since it is considered to be for development purposes. Treating master as a regular release might induce more work from the developer side but I guess is doable.

fbladilo:
Resolved by #101

PR #101, opened 2017-02-21
Do not always force upgrade for master builds

fbladilo:
- Handle master special upgrade/redeploy case properly
- Added conditional to check for supplied master DB schema and compare against DB
- If both schemas are identical, proceed with redeploy, if master image schema is newer, proceed with upgrade


fbladilo:
@bazulay Please review, fixes #93

bazulay:
LGMT

bu I would like @carbonin  to make sure the logic we used (examining the db migration files) is actually correct.

fbladilo:
@carbonin Definitely looks like a much better fit :smile:  So check for pending, if found, then migrate/upgrade. I was not aware of that rake task, Thanks.

fbladilo:
@carbonin @bazulay Please see latest commit using suggested rake task, tested on OCP 3.4 for both cases.

bazulay:
LGTM

PR #104, opened 2017-03-01
Set and parametize resource requests and limits for all pods

fbladilo:
- Set a default min RAM resource requests for all pods
- Set a default max RAM allowed to consume for all pods
- Set a default min CPU resource requests for all pods
- All defaults are parametized and can be controlled at initial deployment
- Allow CPU time to be burstable but request a minimum allocation (expressed in millicores)

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -453,7 +453,7 @@ parameters:
     displayName: "Application Min RAM Requested"
     required: true
     description: "Minimum amount of memory the Application container will need."
-    value: "4096Mi"
+    value: "6144Mi"
   -
     name: "POSTGRESQL_MEM_REQ"
     displayName: "PostgreSQL Min RAM Requested"
@@ -471,13 +471,13 @@ parameters:
     displayName: "Application Max RAM Limit"
     required: true
     description: "Maximum amount of memory the Application container can consume."
-    value: "8192Mi"
+    value: "16384Mi"
   -
     name: "POSTGRESQL_MEM_LIMIT"
     displayName: "PostgreSQL Max RAM Limit"
     required: true
     description: "Maximum amount of memory the PostgreSQL container can consume."
-    value: "2048Mi"
+    value: "8192Mi"
   -
     name: "MEMCACHED_MEM_LIMIT"
     displayName: "Memcached Max RAM Limit"
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -160,7 +160,10 @@ objects:
               value: "${POSTGRESQL_SHARED_BUFFERS}"
           resources:
             requests:
-              memory: "${MEMORY_APPLICATION_MIN}"
+              memory: "${APPLICATION_MEM_REQ}"
+              cpu: "${APPLICATION_CPU_REQ}"
+            limits:
+              memory: "${APPLICATION_MEM_LIMIT}"
           lifecycle:
             preStop:
               exec:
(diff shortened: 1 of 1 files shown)


fbladilo:
These settings are aimed to help the kubernetes scheduler make better informed decisions about Pod placement into nodes, declaring minimum resource requests for RAM and CPU are a must in order to avoid potential crashes on the nodes due to resource starvation/overloading. Requests and limits have been parametized as follows : 

MIQ  
- 4GB RAM min request
- 8GB RAM max limit
- 1 CPU min request
- Burstable CPU max

Postgresql
- 1GB RAM min request
- 2 GB RAM max limit
- 0.5 CPU min request
- Burstable CPU max

Memcached 
- 64MB min request
- 256MB max limit
- 0.2 CPU min request
- Burstable CPU max

Users can modify/tune these defaults as desired via template parameters. I'm attaching two screenshots of a sample MIQ deployment with 1 OCP provider added. (Note the CPU/RAM consumption ramp up as I add the provider) 

screenshot from 2017-02-28 22-08-13
screenshot from 2017-02-28 22-08-54

fbladilo:
@bazulay @simaishi @simon3z Please review...

bazulay:
@fbladilo  what did you base the numbers on ?

fbladilo:
@bazulay Based on the minimum suggested cluster size we suggest, I experienced resource starvation and restarts myself on my cluster this week, the max limits heavily depends on the use of the appliance, perhaps they can be relaxed even further as defaults. I'm open to suggestions on the Max limits, the request numbers are based on a bare minimum you need available in order to bring up each appliance.

simon3z:
> MIQ
> 
> * 4GB RAM min request

FYI I recently had to bump this to 6GB (1 OpenShift provider with 8 nodes and ~20 pods with C&U enabled). Anyway this is just my personal experience.

bazulay:
@fbladilo  I would only leave the minimal requested numbers to make sure scheduler is effected,
And loose the limits (assuming when limits are omitted the system will not limit it ? )
+ I would increase to 6GB based on @simon3z  comment

fbladilo:
@bazulay I also agree with @simon3z on the 6GB request, 4GB is just enough to get you initialized. I see that myself with 1 provider Openshift cluster 3 nodes with C&U. And yes, if we don't a limit it will run until resources run out, in the case of a leak or really heavy consumption it will eventually crash the node assigned along with any other Pods in it. How does 16GB RAM limit sounds for a loose default setting?

Loicavenel:
CloudForms Appliance is coming with 8 GB of memory and 4 Processor Cores.. In latest release, Workers memory has been bumped and some providers are not just creating one Provider but multiple providers (OpenShift = 2 Providers soon). So, 6GB should be a minimum... more you add memory more you will be able to add providers..

fbladilo:
@Loicavenel Any ideas on default limits for RAM consumption? Will a limit of 16GB for a single appliance sound reasonable as a default?

Loicavenel:
Yes, usually customer will add more appliance than upgrading memory. BTW, on DB only Appliance we recommend for large customer 12 GB memory

fbladilo:
@Loicavenel @bazulay Alright, based on feedback we should certainly raise the minimum RAM request for the appliance POD to 6GB. Also set relaxed max limits for 16GB for MIQ pod and 8GB for PG Pod. These should cover most scenarios and of course they are customizable by user, these are just template defaults.

bazulay:
@fbladilo Agreed with the above.

fbladilo:
@bazulay Updated, please review.

PR #116, opened 2017-03-28
Do not migratedb on re-redeployment case

fbladilo:
- Avoid unnecessary db migration on redeployment/reschedule case

### DIFF
--- a/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
+++ b/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
@@ -235,7 +235,7 @@ echo "== Applying memcached config =="
 
 sed -i~ -E "s/:memcache_server:.*/:memcache_server: ${MEMCACHED_SERVICE_NAME}:11211/gi" "${APP_ROOT}/config/settings.yml"
 
-[ "$?" -ne "0" ] && echo "ERROR: Failed to apply memcached configuration, please check journal or PV logs" && exit 1
+[ "$?" -eq "0" ] && return 0 || echo "ERROR: Failed to apply memcached configuration, please check journal or PV logs" && exit 1
 
 }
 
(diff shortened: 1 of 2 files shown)


fbladilo:
@bazulay @carbonin Unnecessary migration for re-schedule, we already validate schema/version to be the same, this is the equivalent of a VM reboot case. It just slows down re-deployments for no benefit.

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -235,7 +235,7 @@ echo "== Applying memcached config =="
 
 sed -i~ -E "s/:memcache_server:.*/:memcache_server: ${MEMCACHED_SERVICE_NAME}:11211/gi" "${APP_ROOT}/config/settings.yml"
 
-[ "$?" -ne "0" ] && echo "ERROR: Failed to apply memcached configuration, please check journal or PV logs" && exit 1
+[ "$?" -eq "0" ] && return 0 || echo "ERROR: Failed to apply memcached configuration, please check journal or PV logs" && exit 1
```
bazulay: @fbladilo  IIUC This is not related to the redundant db migration

fbladilo: Systemd expects 0 exit value on last function run, this is an enhanced one-liner for this function.

bazulay: @fbladilo  please submit in a different PR than

fbladilo: @bazulay I can't , it will fail without this modification, the old check will return 1 (false) since in most cases the return value for the command will actually be zero. (succeeded)

bazulay: @fbladilo  now I understand ... it's only because this is the last function call in the redeploy scenario.

bazulay:
LGTM

PR #117, opened 2017-03-28
Parametize APPLICATION_REPLICA_COUNT for petset controller

fbladilo:
- Allow user to deploy desired number of MIQ replicas using parameters

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -94,7 +94,7 @@ objects:
       description: "Defines how to deploy the ManageIQ appliance"
   spec:
     serviceName: "${NAME}"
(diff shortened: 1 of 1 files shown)


fbladilo:
@bazulay @simaishi Please review..

bazulay:
@fbladilo 
Since I have never tried it ....
How does it work ? the replica is increased after the previous replica is ready ?
Otherwise it may create a race of writing region data  the region_pv

fbladilo:
@bazulay Correct, statefulsets scale orderly, only the previous replica is online/ready the next one starts.

bazulay:
LGMT

Issue #107, opened 2017-03-03
Support External PostgreSQL Instances


simon3z:
It should be possible to deploy ManageIQ podified also with an external PostgreSQL instance.

I suppose in this case you wouldn't deploy the regular posgres dc but just configure an endpoint (not even sure you need a service in this case, an endpoint would be enough IIRC) to point to the external PostgreSQL instance (configuration given by the user).

Maybe it should be a different template?

cc @fbladilo @bazulay

fbladilo:
@simon3z Likely to be a separate template as we will not need all of the PG REST objects at all in this case, also necessary logic needs to be implemented to detect/handle the external PG setup, this data will be required by user, along with proper PG role setup on the remote PG in order to create/migrate DB.

fbladilo:
@bazulay Are we Ok to close this one?

bazulay:
Per the merge of #119 
Closing this issue

PR #119, opened 2017-03-29
Deploy MIQ using external PG database

fbladilo:
- Created new template excluding unncessary PG pod objects
- Added endpoints object and re-configured PG svc
- Remote endpoints parametized via DATABASE_SERVER_IP
- Updated documentation to reflect new deployment option
- PG server must be prepared for MIQ before attempting to use this template

### DIFF
--- a/templates/miq-template-ext-db.yaml
+++ b/templates/miq-template-ext-db.yaml
@@ -74,8 +74,8 @@ objects:
     resources:
       requests:
         storage: ${APPLICATION_REGION_VOLUME_CAPACITY}
-- apiVersion: apps/v1alpha1
-  kind: "PetSet"
+- apiVersion: apps/v1beta1
+  kind: "StatefulSet"
   metadata:
     name: ${NAME}
     annotations:
@@ -88,9 +88,6 @@ objects:
         labels:
           name: ${NAME}
         name: ${NAME}
-        annotations:
-          # 'false' pauses a petset after creation of each pet, to avoid it we set it to 'true'
-          pod.alpha.kubernetes.io/initialized: "true"
       spec:
         containers:
         - name: manageiq
--- a/templates/miq-template-ext-db.yaml
+++ b/templates/miq-template-ext-db.yaml
@@ -1,0 +1,430 @@
+apiVersion: v1
+kind: Template
+labels:
+  template: manageiq-ext-db
+metadata:
+  name: manageiq-ext-db
+  annotations:
+    description: "ManageIQ appliance with persistent storage using a external DB host"
+    tags: "instant-app,manageiq,miq"
+    iconClass: "icon-rails"
+objects:
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: "${NAME}-secrets"
+  stringData:
+    pg-password: "${DATABASE_PASSWORD}"
+- apiVersion: v1
+  kind: Service
+  metadata:
+    annotations:
+      description: "Exposes and load balances ManageIQ pods"
+      service.alpha.openshift.io/dependencies: '[{"name":"${DATABASE_SERVICE_NAME}","namespace":"","kind":"Service"},{"name":"${MEMCACHED_SERVICE_NAME}","namespace":"","kind":"Service"}]'
+    name: ${NAME}
+  spec:
+    clusterIP: None
+    ports:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 80
+    - name: https
+      port: 443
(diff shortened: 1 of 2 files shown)


fbladilo:
@bazulay @simon3z Please review, attempts to resolve #107

fbladilo:
Sample MIQ deployment without DB pod : 
```
[root@franco-master-ocp34 templates (miq_use_external_pg)]# oc get pods
NAME                READY     STATUS    RESTARTS   AGE
manageiq-0          1/1       Running   0          9m
memcached-1-nlt95   1/1       Running   0          8m
[root@franco-master-ocp34 templates (miq_use_external_pg)]# oc get pvc
NAME                         STATUS    VOLUME     CAPACITY   ACCESSMODES   AGE
manageiq-region              Bound     nfs-pv05   5Gi        RWO           9m
manageiq-server-manageiq-0   Bound     nfs-pv06   5Gi        RWO           9m
[root@franco-master-ocp34 templates (miq_use_external_pg)]# oc get endpoints
NAME         ENDPOINTS                        AGE
manageiq     10.129.1.59:443,10.129.1.59:80   9m
memcached    10.130.1.191:11211               9m
postgresql   10.19.114.87:5432                9m
[root@franco-master-ocp34 templates (miq_use_external_pg)]# oc get svc
NAME         CLUSTER-IP       EXTERNAL-IP   PORT(S)          AGE
manageiq     None             <none>        80/TCP,443/TCP   9m
memcached    172.30.106.125   <none>        11211/TCP        9m
postgresql   172.30.30.209    <none>        5432/TCP         9m
```

templates/miq-template-ext-db.yaml
```diff
@@ -0,0 +1,430 @@
...
+  metadata:
+    name: "${DATABASE_SERVICE_NAME}"
+  subsets:
+    -
+      addresses:
+        -
+          ip: "${DATABASE_REMOTE_IP}"
+      ports:
+        -
+          port: 5432
```
bazulay: @fbladilo shouldn't we parameterize the external port (as we do not have control over it) ?

carbonin:
@fbladilo Just after a quick look it seems like many hundreds of lines are duplicated between this new template and the `miq-template.yaml` file that we have now.

Is there some way we can compose some of this config so that it isn't such a maintainability headache in the future?

fbladilo:
@carbonin Yes that's one the disavantages of templates today, they are pretty rigid. Unless you write your own pre-processing/scripting/composer to generate them on-the-fly but I don't think that's an area we want to get in ourselves.

templates/miq-template-ext-db.yaml
fbladilo: @bazulay Since we have no control over the external PG, yes. I will send another commit with a default port parametized, in case the user has to modify.

bazulay:
@simon3z  this is something we'll need to address in case this use case will be officially supported, currently this is only for POC.
I entierly agree that if we support more than a single template than we need to auto generate all supported templates.

fbladilo:
@bazulay @simon3z This is a recurring issue that was brought up in Phoebe meetings, not just MIQ related. This is a general issue with OpenShift templates which are too rigid, we do need a composer/tool to generate them but I don't think the best path of action is to write and maintain our own for MIQ.

templates/miq-template-ext-db.yaml
fbladilo: @bazulay More on this one, I could potentially work with targetport of the service and adjust the endpoint with a parametized custom port, so that the internal image would still talk to the default 5432 port internally and would not require miq-app image changes. (we currently fix our checks against 5432). 

Please keep in mind that this parametized external PG port cannot be carried to the OpenShift PG based deployment since the SCL image does not support binding to custom ports. This will be an external PG only parameter if it works properly.

fbladilo:
@bazulay Parametized DATABASE_PORT, you can now use an external postgresql server listening to non-default port, internally we still use 5432, only service targetPort and external endpoint change.

Default DATABASE_PORT is set of course to 5432.

bazulay:
@fbladilo 

LGTM I verified on origin 1.4 

Need to update the patch staefulset

fbladilo:
@bazulay Cool, should we merge as is? Or another commit including statefulsets?

bazulay:
@fbladilo  prefer updated patch with statefulset

fbladilo:
@bazulay Done, switched petset to statefulset for this template as well.

bazulay:
LGTM

PR #120, opened 2017-03-30
Remove unused monolithic template

fbladilo:
@bazulay This template is not being maintained and don't think it makes sense for this repo anymore.


bazulay:
LGMT

PR #125, opened 2017-04-07
Only install production dependencies for SUI

simaishi:
https://github.com/ManageIQ/manageiq-appliance-build/pull/207


fbladilo:
@simaishi LGTM

PR #127, opened 2017-04-13
Alter bash code styling

carbonin:
This looks big, but I swear it isn't.

This just changes how we use whitespace.
I've indented function bodies and case bodies, changed all indentations to two spaces and removed some extra whitespace.

The only non-whitespace change is moving the function description comments outside the function body.

/cc @bdunne


carbonin:
Use https://github.com/ManageIQ/manageiq-pods/pull/127/files?w=1 to see a much reduced diff.

fbladilo:
@carbonin 👍 we could always use better indentation, agreed on the function descriptions outside function blocks.

PR #130, opened 2017-04-28
Include default name and password

tbielawa:
Provide name and password in the 'verify setup successful' section so admins can browse the web interface without having to lookup other documentation.


tbielawa:
@fbladilo updated. What do you think now?

fbladilo:
@tbielawa 👍

Issue #132, opened 2017-05-05
move container image build process to CentOS CI


goern:
Manageiq container build process seems to be working around some Docker Hub auto build limitations. Using [CentOS CI](https://wiki.centos.org/QaWiki/CI/GettingStarted) could be a way to improve the structure of MIQ container images and to eliminate MIQ requirements like https://github.com/ManageIQ/manageiq-pods#increase-maximum-number-of-imported-images-on-imagestream

carbonin:
We removed image streams from the templates. Are there any other issues moving the build would solve?

We are debating moving the builds to a separate system which the core team would operate so we can encode dependent container builds better. Does Centos CI offer something like that?

goern:
I dont know what you mean be the second section, but on CentOS CI you can create build pipelines structures as you want....

carbonin:
We're going to build and push images nightly using [the new build script](https://github.com/ManageIQ/manageiq-pods/blob/4c7acbbeb288caaf97ff4738675517b9beb766f1/bin/build) on the same machine that builds the appliance images.

PR #136, opened 2017-05-12
Splitting unrelated dockerfiles

bdunne:
Memcached and postgresql images can stand on their own at this point.  They're independent images and should live in their own repos.  This should simplify all of the docker builds as well.


fbladilo:
@bdunne @bazulay  Where will they land? In regards of the Dockerhub auto-build limitations we also received another suggestion to use Centos CI via :

https://github.com/ManageIQ/manageiq-pods/issues/132

bdunne:
@fbladilo My intent is to have `ManageIQ/docker-memcached` on Github with automated builds as `manageiq/memcached` on Docker Hub.

carbonin:
I think we should stick with "postgresql" as the image name, that's the "correct" name of the service and is the same as the image it is based on.

Whether we need the "miq" is still up in the air for me. If it's namespaced under the manageiq docker org then I guess we can drop that.

carbonin:
> we also received another suggestion to use Centos CI

I feel like CI and where the images live are separate concerns right?
If we want to add CI couldn't we do that as a separate effort?

templates/miq-template-ext-db.yaml
```diff
@@ -265,7 +265,7 @@ objects:
   spec:
     ports:
       -
-        name: "postgresql"
+        name: "postgres"
```
bdunne: @fbladilo @carbonin I dropped the `ql` in this PR as well based on https://github.com/bdunne/docker-postgres/commit/fc01fe724886013892aea77502a9e5844e2e9ded
Is that a problem?
Do you want a separate PR before the repo split?
My goal is consistency, don't care whether we add or drop the `ql`, do you have an opinion?

carbonin: I would rather use the full name "postgresql". Is there a reason you chose to standardize on "postgres"?

bdunne: image
Just because that's the name of the official one, but either is Fine by me.

carbonin: Looks like you win this time 😉

fbladilo:
@bdunne @carbonin @bazulay A few considerations for this one : 

I think I understand the motivation for one repo per image, which is to allow DH to indepently build each image and avoid using the image tag as to identify the component (memcache,pg), is there other reasons/benefits behind the break up? I pointed the Centos CI issue open because it might offer an alternative for this, I haven't researched enough yet.

In terms of separate repos, I can already say that the memcached one will be short lived, since there is already work being done to provide a SCL image and we will not have to use our own anymore. I do believe that in the future we will only need images that are layered/customized for MIQ only.

In terms of naming for the repos, docker has changed its runtime name to moby and I don't think tying the repo name to anything docker is a good idea now, just in case we happen to go with a different container runtime in the future. Something more generic might make sense postgresql-container or miq-pg-container.

Regarding the naming change to posgres, I'm with Nick that I believe the proper name is postgresql for the service itself, postgres I have only seen it as a the user who runs the daemon. In addition, it is possible a user might have multiple deployments in a single namespace/project, so going with a generic postgresql name might cause name clashing without the prepending miq on certain objects, I need to scan the template line by line but it might also have implications on the PVCs.

templates/miq-template-ext-db.yaml
fbladilo: These are the official templates and naming from OpenShift Origin, don't recall a postgres reference for the service before, just saying... 

https://github.com/openshift/origin/blob/master/examples/db-templates/postgresql-persistent-template.json

carbonin: Naming is hard 😕

fbladilo: I know , I'm inclined over the postgresql naming since that was OpenShift/k8s is doing, after all that's the target platform.

bazulay:
@bdunne @carbonin @fbladilo @simaishi 

The image names used in this repo were to distinct these images from the "official" docker images, for several reasons:
- memcached - the official image was based on ubuntu ... we wanted one that is built from centos
- postgresql - we based on official SCL image that was missing some packages that we added specifically to manageiq

Both the images do no belong in manageiq repo to begin with, as these should be provided by other groups, there for the names reflect this fact.
SCL group  is about to release memcached generic image & there is work going on to support postgres image that will support upgrade (and hopefully HA ) out of the box for general use cases (oVirt, satellite & other groups need it as well)

I would not have a strong opinion about the names, i just want to be able to distinguish the images by their names once they get to the relevant registry.

my 0.02$ is to postpone this work till the other images will be delivered.

templates/miq-template-ext-db.yaml
bdunne: changed to `postgresql`

PR #137, opened 2017-05-12
Separate ruby container

carbonin:
The ruby part of this image doesn't change nearly as much as the miq specific bits.

If we create a separate ruby container and layer miq on top of it, we can get faster build times and make development a lot faster.

ref:
https://github.com/ManageIQ/container-ruby
https://hub.docker.com/r/manageiq/ruby/builds/

### DIFF
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:7
+FROM manageiq/ruby
 MAINTAINER ManageIQ https://github.com/ManageIQ/manageiq-appliance-build
 
 ## Set build ARGs
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -34,8 +34,11 @@ STOPSIGNAL SIGRTMIN+3
 ## Install EPEL repo, yum necessary packages for the build without docs, clean all caches
 RUN yum -y install centos-release-scl-rh && \
     yum -y install --setopt=tsflags=nodocs \
+                   cmake                   \
                    file                    \
+                   gcc-c++                 \
                    git                     \
+                   libcurl-devel           \
                    libtool                 \
                    libxslt-devel           \
                    net-tools               \
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -6,14 +6,12 @@ ARG REF=master
 
 ## Set ENV, LANG only needed if building with docker-1.8
 ENV container=docker \
-    LANG=en_US.UTF-8 \
     TERM=xterm \
     APP_ROOT=/var/www/miq/vmdb \
     APP_ROOT_PERSISTENT=/persistent \
     APP_ROOT_PERSISTENT_REGION=/persistent-region \
     APPLIANCE_ROOT=/opt/manageiq/manageiq-appliance \
     SUI_ROOT=/opt/manageiq/manageiq-ui-service \ 
-    RUBY_GEMS_ROOT=/opt/rubies/ruby-2.3.1/lib/ruby/gems/2.3.0 \
     CONTAINER_SCRIPTS_ROOT=/opt/manageiq/container-scripts \
     IMAGE_VERSION=${REF}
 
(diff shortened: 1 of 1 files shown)


images/miq-app/Dockerfile
```diff
@@ -1,19 +1,17 @@
-FROM centos:7
+FROM carbonin/manageiq-ruby-container
```
carbonin: I'm not sold on the image name...

fbladilo: I know, in principle I'm wondering why we need a special ruby for MIQ. There are plenty of ruby containers already, need to check SCL offerings.

carbonin: Yeah, I mentioned that to @simaishi and she mentioned the SCL one. Not sure if that was just a testing effort or some other issue?

simaishi: The upstream build doesn't use SCL ruby at all, so that's a separate discussion... If there are other ruby containers we can use and we don't need to build our own, that'll be great!

carbonin: Great! I'll try a few and see what we get.

carbonin: Ah I think the issue is the OS.

The official ruby containers are build on debian but we need CentOS

bdunne: Yeah, I think we want a centos base image.

I'd like to see the GH repo be `ManageIQ/docker-ruby` and the docker build be `manageiq/ruby` similar to the ones I created for memcached and postgres

fbladilo: Yeah, whatever we layer from needs to be RH aligned (Centos), is there a reason for a specific ruby version (2.3.1) for MIQ purposes? Could we live with a different ruby version coming from SCL? 

Need to investigate what is available, in general SCL builds are preferred way since we are guaranteed to have a RHEL build also available internally and maintained accordingly.

bdunne: We didn't go with SCL upstream because we didn't want to be held back waiting for newer rubies to be built.  We're close to switching to ruby 2.4 upstream, but there is no software collection for it yet.  Do you know when those packages will be available?

fbladilo: I'm not sure how often SCL releases these and it might require coordination with them. In any case, don't think maintaining our own ruby containers is something we want long term.

simaishi: Moving to SCL ruby or not isn't specific to container, as it also affects VM appliances and should be discussed with other people as well...

Can we split this out for now by maintaining our own ruby container, and consider using stock ruby container separately?

images/miq-app/Dockerfile
```diff
@@ -34,30 +32,17 @@ LABEL name="manageiq" \
...
-RUN yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm && \
-    yum -y install centos-release-scl-rh && \
+RUN yum -y install centos-release-scl-rh && \
     yum -y install --setopt=tsflags=nodocs \
-                   bison                   \
-                   bzip2                   \
-                   cmake                   \
                    file                    \
-                   gcc-c++                 \
-                   git                     \
```
carbonin: @Fryguy which of these need to be added back because they are also dependencies of ManageIQ?

carbonin:
Just finished testing this out with the new ruby docker image.

Works well.

I would like to remove the need to add the ruby path the `/etc/default/evm`, but I need to remove [this line](https://github.com/ManageIQ/manageiq-appliance/blob/master/LINK/usr/bin/evmserver.sh#L3) from `evmserverd.sh` before we can do that. Currently if we don't print the ruby path to `/etc/default/evm`, we fail because we can't find `rake`

fbladilo:
@carbonin Looks good, haven't had a chance to deploy this week... 😐

bdunne:
Another thing that changed is we no longer install chruby.  It doesn't appear to be needed, we're only expecting to ever have one ruby on the container.

Fryguy:
@carbonin @bdunne Should I update the pods repo to kick when the ruby base image is updated?

carbonin:
Ah, yeah. That would be good @Fryguy

PR #139, opened 2017-05-16
Remove duplicate APP_ROOT environment variable

carbonin:
The container scripts are the only place this is used so the Docker
ENV directive should be enough.

### DIFF
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -100,7 +100,6 @@ RUN ${APPLIANCE_ROOT}/setup && \
     mv /etc/httpd/conf.d/ssl.conf{,.orig} && \
     echo "# This file intentionally left blank. ManageIQ maintains its own SSL configuration" > /etc/httpd/conf.d/ssl.conf && \
     cp ${APP_ROOT}/config/cable.yml.sample ${APP_ROOT}/config/cable.yml && \
-    echo "export APP_ROOT=${APP_ROOT}" >> /etc/default/evm && \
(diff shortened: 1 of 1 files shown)


images/miq-app/Dockerfile
```diff
@@ -100,7 +100,6 @@ RUN ${APPLIANCE_ROOT}/setup && \
     mv /etc/httpd/conf.d/ssl.conf{,.orig} && \
     echo "# This file intentionally left blank. ManageIQ maintains its own SSL configuration" > /etc/httpd/conf.d/ssl.conf && \
     cp ${APP_ROOT}/config/cable.yml.sample ${APP_ROOT}/config/cable.yml && \
-    echo "export APP_ROOT=${APP_ROOT}" >> /etc/default/evm && \
     echo "export CONTAINER=true" >> /etc/default/evm
```
bdunne: Should this move up to be declared in the same `ENV` as the `APP_ROOT`?

carbonin: I don't think we can do that. systemd resets the environment when it starts a unit so I'm afraid moving this would prevent the application from seeing it.

I didn't test specifically with this variable, but when I moved the ruby PATH change to just be in the ruby Dockerfile the evmserverd.service failed to start because it couldn't find `rake`.

I don't think it's worth coding around that if we are trying to get rid of systemd anyway.

fbladilo: CONTAINER needs to be moved and set in the ENV variables layer, I meant to do that on the last optimization/cleanup.

PR #140, opened 2017-05-16
Remove systemd in favor of dumb-init for process management

carbonin:
This allows us to move away from some of the unpleasantness that systemd enforces when running in a container.

In particular this allows us to carry our environment through to the server process.

This PR also handles some of the fallout from this change, namely:
  - Consolidation of ENVVARs
  - Creation of the server ssl certificate (previously done in [evmserver.sh](https://github.com/ManageIQ/manageiq-appliance/blob/master/LINK/usr/bin/evmserver.sh#L10))

For this to work properly I still need to start `httpd` manually on the command line. /cc @bdunne

### DIFF
--- a/images/miq-app/docker-assets/appliance-initialize.sh
+++ b/images/miq-app/docker-assets/appliance-initialize.sh
@@ -11,6 +11,9 @@ sleep "${APPLICATION_INIT_DELAY}"
 # Prepare initialization environment
 prepare_init_env
 
+# Generate the certs if needed
+/usr/bin/generate_miq_server_cert.sh
+
 # Check Memcached readiness
 check_svc_status ${MEMCACHED_SERVICE_NAME} 11211
 
--- a/README.md
+++ b/README.md
@@ -61,13 +61,13 @@ The default service account for your namespace (project) must be added to the pr
 _**As admin**_
 
 ```bash
-$ oadm policy add-scc-to-user privileged system:serviceaccount:<your-namespace>:default
+$ oadm policy add-scc-to-user anyuid system:serviceaccount:<your-namespace>:default
 ```
 
 Verify that your default service account is now included in the privileged scc
 ```
-$ oc describe scc privileged | grep Users
-Users:					system:serviceaccount:openshift-infra:build-controller,system:serviceaccount:management-infra:management-admin,system:serviceaccount:management-infra:inspector-admin,system:serviceaccount:default:router,system:serviceaccount:default:registry,system:serviceaccount:<your-namespace>:default
+$ oc describe scc anyuid | grep Users
+Users:					system:serviceaccount:<your-namespace>:default
 ```
 
 ### Make persistent volumes to host the MIQ database and application data
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Example NFS PV templates are provided, **please skip this step you have already 
 
 For NFS backed volumes, please ensure your NFS server firewall is configured to allow traffic on port 2049 (TCP) from the OpenShift cluster.
 
-_**Note:**_ Recommended permissions for the PV volumes are 775, root uid/gid owned.
+_**Note:**_ Recommended permissions for the PV volumes are 777, root uid/gid owned.
 
 _**As admin:**_
 
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -29,9 +29,6 @@ LABEL name="manageiq" \
       io.openshift.expose-services="443:https" \
       io.openshift.tags="ManageIQ,miq,manageiq"
 
-## To cleanly shutdown systemd, use SIGRTMIN+3
-STOPSIGNAL SIGRTMIN+3
-
 ## Install EPEL repo, yum necessary packages for the build without docs, clean all caches
 RUN yum -y install centos-release-scl-rh && \
     yum -y install --setopt=tsflags=nodocs \
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -140,5 +140,5 @@ RUN wget -O /usr/local/bin/dumb-init https://github.com/Yelp/dumb-init/releases/
 RUN chmod +x /usr/local/bin/dumb-init
 
 ## Call systemd to bring up system
-ENTRYPOINT ["/usr/local/bin/dumb-init", "--"]
+ENTRYPOINT ["/usr/local/bin/dumb-init", "--single-child", "--"]
 CMD ["entrypoint"]
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -140,9 +140,9 @@ COPY docker-assets/container.data.persist /
 COPY docker-assets/appliance-initialize.sh /bin
 ADD  docker-assets/container-scripts ${CONTAINER_SCRIPTS_ROOT}
 
-## Enable services on systemd
-RUN systemctl enable evmserverd crond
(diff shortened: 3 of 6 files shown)


images/miq-app/docker-assets/appliance-initialize.sh
```diff
@@ -11,6 +11,9 @@ sleep "${APPLICATION_INIT_DELAY}"
 # Prepare initialization environment
 prepare_init_env
 
+# Generate the certs if needed
+${APPLIANCE_ROOT}/initialize_appliance.sh
+
```
fbladilo: I think there is a race on this one, might cause a new v2_key to be generated on every pod re-schedule. The image comes baked without keys/certs, we restore the original backed up certs/keys from the PV on every re-schedule. This happens [here](https://github.com/ManageIQ/manageiq-pods/blob/master/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh#L281)

carbonin: I think I'm just going to add the `--key` back to the `appliance_console_cli` command and remove it from this script.

I'll just have to also add it to https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/db_init.ks.erb#L7

carbonin:
Requires https://github.com/ManageIQ/manageiq-appliance/pull/123

carbonin:
As of https://github.com/ManageIQ/manageiq-gems-pending/pull/163 httpd is starting properly with this change.

I think it's ready for some real review.

Would appreciate input from @fbladilo and @bazulay on this one.

carbonin:
Moving this to WIP. I want to look into the signal handling offered by dumb-init wrt reschedule.

Also, I think we may have an issue with the cert generation on reschedule. I need to make sure that the cert is persisted and is the same after we restart the pod.

chessbyte:
@carbonin So this PR resolves the Privileged Container gap?

carbonin:
With this change our process tree on the container looks like this:

```plain
sh-4.2# ps -leaf --forest
F S UID        PID  PPID  C PRI  NI ADDR SZ WCHAN  STIME TTY          TIME CMD
4 S root         1     0  0  80   0 -    50 -      14:48 ?        00:00:00 /usr/local/bin/dumb-init -- entrypoint
4 S root         6     1 13  80   0 - 138642 -     14:48 ?        00:00:38 MIQ Server
1 S root       163     6  0  87   7 - 127848 -     14:50 ?        00:00:00  \_ MIQ: MiqEventHandler id: 12, queue: ems
1 S root       171     6  0  90  10 - 127848 -     14:50 ?        00:00:01  \_ MIQ: MiqGenericWorker id: 13, queue: generic
1 S root       179     6  0  90  10 - 128105 -     14:50 ?        00:00:01  \_ MIQ: MiqGenericWorker id: 14, queue: generic
1 S root       187     6  0  81   1 - 128619 -     14:50 ?        00:00:01  \_ MIQ: MiqPriorityWorker id: 15, queue: generic
1 S root       195     6  0  81   1 - 128876 -     14:50 ?        00:00:01  \_ MIQ: MiqPriorityWorker id: 16, queue: generic
1 S root       203     6  0  87   7 - 128876 -     14:50 ?        00:00:00  \_ MIQ: MiqReportingWorker id: 17, queue: reporting
1 S root       211     6  0  87   7 - 129390 -     14:50 ?        00:00:00  \_ MIQ: MiqReportingWorker id: 18, queue: reporting
1 S root       219     6  1  83   3 - 129922 -     14:50 ?        00:00:02  \_ MIQ: MiqScheduleWorker id: 19
1 S root       232     6  0  81   1 - 132230 -     14:50 ?        00:00:00  \_ puma 3.3.0 (tcp://127.0.0.1:5000) [MIQ: Web Server Worker]
1 S root       240     6  0  81   1 - 133505 -     14:50 ?        00:00:02  \_ puma 3.3.0 (tcp://127.0.0.1:3000) [MIQ: Web Server Worker]
1 S root       248     6  0  81   1 - 132488 -     14:50 ?        00:00:00  \_ puma 3.3.0 (tcp://127.0.0.1:4000) [MIQ: Web Server Worker]
1 S root         9     1  0  80   0 -  6513 -      14:48 ?        00:00:00 /usr/sbin/crond
4 S root       153     1  0  80   0 - 62464 -      14:50 ?        00:00:00 /usr/sbin/httpd -DFOREGROUND
5 S apache     154   153  0  80   0 - 62699 -      14:50 ?        00:00:00  \_ /usr/sbin/httpd -DFOREGROUND
5 S apache     155   153  0  80   0 - 62530 -      14:50 ?        00:00:00  \_ /usr/sbin/httpd -DFOREGROUND
5 S apache     156   153  0  80   0 - 62530 -      14:50 ?        00:00:00  \_ /usr/sbin/httpd -DFOREGROUND
5 S apache     157   153  0  80   0 - 62530 -      14:50 ?        00:00:00  \_ /usr/sbin/httpd -DFOREGROUND
5 S apache     158   153  0  80   0 - 62594 -      14:50 ?        00:00:00  \_ /usr/sbin/httpd -DFOREGROUND
5 S apache     346   153  0  80   0 - 62530 -      14:52 ?        00:00:00  \_ /usr/sbin/httpd -DFOREGROUND
```

So even though all of our processes (crond, httpd, and evmserver) have dumb-init as the parent process, it only tracks one child pid (https://github.com/Yelp/dumb-init/blob/master/dumb-init.c#L63).

This means that dumb-init will forward a SIGTERM to evmserver which [already implements signal handling](https://github.com/ManageIQ/manageiq/blob/master/lib/workers/evm_server.rb#L66-L71)

I think that this would leave crond and httpd in a situation where they would just be killed when the main process exits. This sounds like dumb-init isn't really making a difference.

I'm going to do some more testing with reschedules to test this theory before removing WIP.

carbonin:
> @carbonin So this PR resolves the Privileged Container gap?

I believe so, as we move from a privileged scc to [`anyuid` scc](https://docs.openshift.org/latest/admin_guide/manage_scc.html#enable-dockerhub-images-that-require-root) which should be more palatable for our users and continues to allow our app to run as root in the container.

fbladilo:
@carbonin This one is tough if dumb-init will not track and forward signals to rest of childs, it gives more ground to the option (at least as a intermediate step) of actually spawning httpd/crond from evmserver and do the signal handling there, no dumb-init would be needed.

carbonin:
From https://github.com/ManageIQ/manageiq-pods/pull/140#issuecomment-302726476

> This sounds like dumb-init isn't really making a difference.

Keeping dumb-init for zombie process reaping. We will run it in single child mode which will only send signals to the immediate child, which for us is the server process.

I'll remove WIP after https://github.com/ManageIQ/manageiq/pull/15206 is merged which fixes signal handling in the server process so that `SIGTERM` causes a graceful stop.

carbonin:
@fbladilo mentioned that the change in scc here changes the way the PG and memcached pods run. After this change they will run as the postgresql and memcached users which don't have rights to a PV set to 0770 root/root permissions, which is a regression.

@fbladilo is looking into options for setting just the app pod to use the anyuid scc.

Any progress on that @fbladilo or should we merge this now and sort out the permissions issues in a follow up?

fbladilo:
@carbonin Haven't had a chance to work on the SCC/anyuid adjustments yet, but feel free to move forward and I'll send another PR with options for the MIQ pod config. Just keep in mind to set those PG volumes to 777 in the interim.

carbonin:
@fbladilo pushed a commit to update the README to recommend 777 permissions for PVs for now.

PR #143, opened 2017-05-23
Once the DB is initialized, we shouldn't need the REGION file

bdunne:
The server should pull the region number from the database


fbladilo:
@bdunne 👍

PR #144, opened 2017-05-24
Ensure container.data.persist is kept up to date

bdunne:


bdunne:
Marking as WIP, I didn't realize there was more than 1...

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -177,8 +177,8 @@ function write_deployment_info() {
 
 # Prepare appliance initialization environment
 function prepare_init_env() {
-  # Make a copy of CONTAINER_DATA_PERSIST_FILE into PV if not present
-  [ ! -f "${PV_DATA_PERSIST_FILE}" ] && cp -a "${CONTAINER_DATA_PERSIST_FILE}" "${APP_ROOT_PERSISTENT}"
+  # Copy CONTAINER_DATA_PERSIST_FILE into PV ensuring it is kept up to date
+  cp -a "${CONTAINER_DATA_PERSIST_FILE}" "${APP_ROOT_PERSISTENT}"
```
carbonin: Is there a reason we have to have this be user modifiable? @fbladilo ?

I'm worried that if we want to add something to this file, that will never get used because we only copy over to the PV if the file doesn't already exist. So existing installations will only ever persist what was in the file on first deploy.

If we need the customer to be able to persist additional things I would rather have a separate file just for that and always read this one from the image so that we will always have the most up to date list of files to persist.

simaishi:
Nope, productization isn't affected by this.

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
fbladilo: @carbonin I have doubts about the original idea of having users modify a file and persist their data. It's not very clean and could lead to issues... 

Regarding this one, the original persist file on PV gets copied only once on normal circumstances, then it is used by the sync-data-pv function always from the copy on PV on a pre-stop event when we shutdown. If a user modifies the persist file in PV , it will always get processed and restored accordingly.

Look here on template : 
https://github.com/ManageIQ/manageiq-pods/blob/master/templates/miq-template.yaml#L181

fbladilo:
@bdunne @carbonin There are more parts to this one, if we go with the baked static persist file then sync-pv-data becomes irrelevant, as well as pre-stop hooks to call it on template. See my comment to Nick above.

carbonin:
I think the ability for us to change this file on upgrade and have our changes take effect outweighs the user modification use case (in fact, I'm not sure what a user would use this feature for ...).

I think we should come up with a different solution that doesn't override the use of this file for that use case.

fbladilo:
@carbonin Agreed, there were some initial intentions last year to have this feature help with certain things, I recall hotfixes, perhaps also custom certs, but this is no longer the case, as there are other ways to accomplish this. It would make sense also to remove the sync-pv-data function, script and openshift hook on template.

carbonin:
✂️ 🔥 🔥 ✂️ ✂️ 🔥 🔥 ✂️ ✂️ 🔥 🔥 ✂️

bdunne:
Updated to reduce some duplicate rsync logic.

I think we still want to run the `sync-pv-data` hook since server certs are backed up to the server PV and it is possible that they have been added / changed since initial creation.

PR #148, opened 2017-05-26
Only generate the server certificate on new deployments.

bdunne:
All other cases should restore the cert from the PV


carbonin:
Is the cert on the region PV? Otherwise I think we would need it for new-replica as well right?

bdunne:
@carbonin Updated for the new replica.  Previously, this was run before the cert files were restored from the PV.

PR #149, opened 2017-05-30
Set memcached server override in ENV["MEMCACHED_SERVER"] rather than sed

bdunne:
We shouldn't edit settings.yml as it's the default shipping with the product.

### DIFF
--- a/images/miq-app/docker-assets/appliance-initialize.sh
+++ b/images/miq-app/docker-assets/appliance-initialize.sh
@@ -14,8 +14,6 @@ prepare_init_env
 # Check Memcached readiness
 check_svc_status ${MEMCACHED_SERVICE_NAME} 11211
 
-setup_memcached
-
 # Check DB readiness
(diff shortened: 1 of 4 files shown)


carbonin:
Would this be better as a docker-asset? Or do you want to keep this openshift-ish stuff away from the pure docker stuff?

bdunne:
I think ERB rendering in YAML is disabled in production mode, and I don't want to ship a hard-coded value since the environment variable could change.

Fryguy:
I think it might be better to store it in config/environment/production.yml .  I like to leave settings.local.yml for users.

carbonin:
@bdunne Was there a reason to keep this file in `settings.local.yml` rather than take @Fryguy 's suggestion of `production.local.yml`?

bdunne:
With the changes merged in https://github.com/ManageIQ/manageiq/pull/15326 we can use an ENV variable instead.

Fryguy:
@bdunne Can you change the PR title to reflect the new direction?

PR #152, opened 2017-05-31
Remove systemd container var from miq-app Dockefile

fbladilo:


fbladilo:
@carbonin @bdunne Another systemd cleanup..

PR #153, opened 2017-06-01
Restrict anyuid SCC to miq pod only via service account

fbladilo:


fbladilo:
@carbonin @bdunne This prevents other pods from requesting anyuid when they don't actually need it. This was possible because we granted anyuid SCC access to the default project account. 

Now only the MIQ pod can access anyuid : 

```
[root@franco-ocp-master # oc exec manageiq-0 id
uid=0(root) gid=0(root) groups=0(root)
[root@franco-ocp-master # oc exec memcached-1-215jw id
uid=1000100000 gid=0(root) groups=0(root),1000100000
[root@franco-ocp-master # oc exec postgresql-1-25lvn id
uid=1000100000 gid=0(root) groups=0(root),1000100000
```

carbonin:
Can we create the service account as a part of the template? Or does it have to be done from the command line?

Also can we make the service account reference the permissions so we could use the same one across multiple pods (and have the naming make sense) or is it better to create one per pod even if they have the same permissions?

fbladilo:
@carbonin Yes it's possible, I thought about including the service account as part of the template. The ordering of events might be a little counter intuitive, as you need to assign the anyuid SCC to a service account that does not yet exist, need to test if OpenShift is Ok with this.

Are you talking about the naming of the service account? The service account is reusable for any other pods that we might need anyuid access in the future, permissions have not been altered as I used the already existing anyuid SCC from OpenShift without modifications.

carbonin:
> @carbonin Yes it's possible, I thought about including the service account as part of the template. The ordering of events might be a little counter intuitive, as you need to assign the anyuid SCC to a service account that does not yet exist, need to test if OpenShift is Ok with this.

I'm okay either way really. I was just thinking fewer manual steps would be better if there is no choice in the matter (we need this service account to be created unconditionally).

With this in mind, I assume we can't assign an scc to a service account in the template right? That would be the most convenient, but would also be a pretty significant security hole ...

> Are you talking about the naming of the service account?

Yeah, I was thinking we could call the service account `miq-anyuid` or something in case we need to assign it to a pod that isn't manageiq (like postgres or apache).

Additionally we may need the privileged scc for an "embedded ansible" pod I'm starting to work on. So then we could create either a `miq-privileged` service account or an `ansible` service account, but then it definitely wouldn't make sense for other pods to use the `ansible` account if they need privileged.

fbladilo:
@carbonin Renamed the service account to miq-anyuid , makes more sense if we eventually use for other pods. Also injected the service account object within the template to reduce one step, I verified that OpenShift will actually allow to set an SCC to a service account that does not yet exists. Docs updated accordingly.

PR #154, opened 2017-06-01
Simple cleanup script before redeployment

bdunne:


fbladilo:
@bdunne This is what I currently use, similar but requires a selector always : 
```
[root@franco-ocp-master manageiq-pods (master)]# cat /usr/bin/cleanup_ocp_project 
#!/bin/bash

[[ $# -lt 1 ]] && echo "ERROR: $_ requires a selector to perform cleanup" && exit 1

SELECTOR="$1"

oc delete statefulset -l app=${SELECTOR}
oc delete secret -l app=${SELECTOR}
oc delete all -l app=${SELECTOR}
oc delete pvc --all
```

PR #155, opened 2017-06-01
Remove all database environment variables in favor of DATABASE_URL

bdunne:
DATABASE_URL is built into Rails and reduces a lot of the complexity of ENV variables and maintaining database.yml in multiple locations.  It seems like the best solution for production deployments.

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -19,6 +19,7 @@ objects:
     name: "${NAME}-secrets"
   stringData:
     pg-password: "${DATABASE_PASSWORD}"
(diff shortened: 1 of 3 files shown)


templates/miq-template.yaml
```diff
@@ -149,29 +150,14 @@ objects:
               name: "DATABASE_REGION"
               value: "${DATABASE_REGION}"
             -
-              name: "DATABASE_NAME"
```
fbladilo: We need all of the database parameters for the PG pod, they are used to create the database, user and password along with tunables, check on template, PG pod objects.

carbonin:
Also, was there a PR this depended on in the core repo? @bdunne

bdunne:
Yes, https://github.com/ManageIQ/manageiq/pull/15269 was merged.

PR #156, opened 2017-06-06
[WIP] Enhancement to have httpd run in a separate pod

abellotti:
- Leveraging new manageiq/httpd container
- httpd pod is now exposing 80/443
- Application route to httpd pod
- httpd redirects to ManageIQ for classic ui (3000), api (4000) and websockets (5000)
- Creating and using miq-sysadmin service account for httpd

screen shot 2017-06-06 at 3 57 42 pm


abellotti:
/cc @bdunne @carbonin for you tire-kicking pleasures.

One question thought, for the templates/miq-sysadmin, I wonder if it might be better to remove that file and replace with instructions in README.md on how to recreate from the anyuid SCC.  Let me know.

abellotti:
/cc @simon3z one remaining unknown:

The route gets properly defined to the httpd service:

screen shot 2017-06-06 at 3 48 21 pm

But the main OpenShift overview windows does not thing a route exists:

screen shot 2017-06-06 at 3 49 25 pm

images/miq-app/Dockerfile
```diff
@@ -72,6 +72,7 @@ RUN mkdir -p ${APP_ROOT} && \
     mkdir -p ${SUI_ROOT} && \
     ln -vs ${APP_ROOT} /opt/manageiq/manageiq && \
     curl -L https://github.com/ManageIQ/manageiq-appliance/tarball/${REF} | tar vxz -C ${APPLIANCE_ROOT} --strip 1 && \
+    mkdir -p ${APP_ROOT}/public/assets && \
```
bdunne: Won't this be created for us?

templates/miq-template.yaml
```diff
@@ -367,7 +436,81 @@ objects:
...
+      metadata:
+        name: "${HTTPD_SERVICE_NAME}"
+        labels:
+          name: "${HTTPD_SERVICE_NAME}"
+      spec:
+        serviceAccount: miq-sysadmin
+        serviceAccountName: miq-sysadmin
+        volumes:
+          -
+            name: "miq-httpd-volume"
```
bdunne: Should this be `${NAME}-httpd` instead?

templates/miq-template.yaml
```diff
@@ -367,7 +436,81 @@ objects:
...
+              tcpSocket:
+                port: 443
+            livenessProbe:
+              timeoutSeconds: 1
+              initialDelaySeconds: 30
+              tcpSocket:
+                port: 443
+            volumeMounts:
+              -
+                name: "miq-httpd-volume"
```
bdunne: Should this be `${NAME}-httpd` instead?

templates/miq-template.yaml
```diff
@@ -367,7 +436,81 @@ objects:
...
+                mountPath: "/persistent"
+              -
+                name: "${NAME}-assets"
+                mountPath: "/persistent-assets"
+            env:
+              -
+                name: "APPLICATION_DOMAIN"
```
bdunne: Where is `APPLICATION_DOMAIN` used?

templates/miq-template.yaml
```diff
@@ -139,8 +195,14 @@ objects:
               -
                 name: "${NAME}-region"
                 mountPath: "/persistent-region"
+              -
+                name: "${NAME}-assets"
+                mountPath: "/persistent-assets"
           env:
             -
+              name: "APPLICATION_DOMAIN"
```
bdunne: Where is `APPLICATION_DOMAIN` used?

templates/miq-template.yaml
```diff
@@ -176,6 +241,10 @@ objects:
            name: "${NAME}-region"
            persistentVolumeClaim:
              claimName: ${NAME}-region
+         -
+           name: "${NAME}-assets"
```
carbonin: This PV is shared then? Is that because we do the compilation on this container and then use the assets on the apache one?

If so, why not do the compilation on the apache container?

templates/miq-template.yaml
abellotti: upcoming ext-auth scripts, essentially the externally seen FQDN of the ManageIQ image.

carbonin: I think this might be better as a followup if it's going to come with the whole external auth configuration.

templates/miq-template.yaml
```diff
@@ -158,6 +220,9 @@ objects:
             -
               name: "MEMCACHED_SERVICE_NAME"
               value: "${MEMCACHED_SERVICE_NAME}"
+            -
+              name: "BINDING_ADDRESS"
```
bdunne: If not specified, the rails servers listen on 127.0.0.1 and don't listen for requests on any other IP.  We need to tell it to listen on all interfaces to respond to the requests forwarded from the httpd pod.

templates/miq-template.yaml
bdunne: Then we need all of the application and its dependencies on this container... :cry:

carbonin: I think I would rather have that than another PV .. is image size that much of a concern?

Additionally I feel like assets shouldn't be this closely tied to the app runtime. Is there a way to ship them in the image rather than having to persist them? Maybe we can grab the app code, compile them then remove the app code in the docker file?

templates/miq-template.yaml
abellotti: yes and no, it is an openshift proper variable and we use it in the Route definition, does not hurt to provide that today.

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -323,3 +323,8 @@ function sync_pv_data() {
     [ "$?" -ne "0" ] && echo "WARNING: Some files might not have been copied please check logs at ${PV_DATA_SYNC_LOG}"
   ) 2>&1 | tee "${PV_DATA_SYNC_LOG}"
 }
+
+# Copy the assets built with the image into the /persistent-assets PV
+function sync_assets() {
+    rsync -avL "/var/www/miq/vmdb/public/assets" "/persistent-assets"
```
simon3z: @abellotti maybe you need `--delete` as well to make sure you don't leave dangling files?

simon3z:
@abellotti I a little bit scared about the `sync_assets` part. I understand the dependency problems of the assets but I have the feeling this would be better handled with an http container image that pre-embeds the assets.

Also you have to consider that the `ReadWriteMany` volumes are not available everywhere so it may raise a compatibility issue going forward.
See: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

cc @bazulay @fbladilo

templates/miq-template.yaml
```diff
@@ -442,7 +585,7 @@ parameters:
     displayName: "Application Min CPU Requested"
     required: true
     description: "Minimum amount of CPU time the Application container will need (expressed in millicores)."
-    value: "1000m"
+    value: "900m"
```
carbonin: Or was this decreased because we won't be using as much CPU because we removed Apache?

Does it make sense to decrease the minimum memory requested also? Maybe we need some real numbers on the idle memory usage with and without apache.

carbonin:
> Also you have to consider that the ReadWriteMany volumes are not available everywhere so it may raise a compatibility issue going forward.
> See: https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes

This is interesting, I didn't know that this mode was so uncommon.

@abellotti I think it would be worth it to try to get whatever auth configuration we need into the various ENV constructs that openshift provides rather than using a PV then trying to get rid of the PV when people complain about the lack of RWX support.

templates/miq-template.yaml
fbladilo: APPLICATION_DOMAIN is the host exposed by the router in OpenShift, by default it's blank allowing OpenShift to decide / auto-generate. The parameter intends to allow a user to customize it if needed.  

It should not be needed at all in the miq application pod anymore, since the route will now connect to the service for HTTPD and that's where APPLICATION_DOMAIN should take effect.

templates/miq-template.yaml
fbladilo: Can we ship this together with the image as Nick suggested? Reducing the storage requirements when unnecessary is a huge plus.

templates/miq-template.yaml
fbladilo: Getting a new memory usage footprint when idling should be straight forward using a Openshift cluster with metrics, we can adjust as we have hard data on it.

templates/miq-template.yaml
```diff
@@ -97,6 +131,26 @@ objects:
     resources:
       requests:
         storage: ${APPLICATION_REGION_VOLUME_CAPACITY}
+- apiVersion: v1
+  kind: PersistentVolumeClaim
+  metadata:
+    name: "${NAME}-${HTTPD_SERVICE_NAME}"
+  spec:
+    accessModes:
+      - ReadWriteOnce
```
fbladilo: What kind of data do we need to save on the PV? Is it configurations? data + configurations? Ideally we would not use PVs to save config data and use configmap/secret to externalize these.

abellotti: both data (logging) & config.

carbonin:
I think we can close this now that https://github.com/ManageIQ/manageiq-pods/pull/162 has been merged.

PR #157, opened 2017-06-07
[WIP] Remove httpd from miq-app

bdunne:
Related to changes in https://github.com/ManageIQ/manageiq-pods/pull/156

See also: https://github.com/ManageIQ/container-httpd/pull/4
cc @abellotti


carbonin:
I think most of this will be rolled back with the new arch which keeps httpd running on the application pods right?

PR #160, opened 2017-06-15
Move the v2_key from a file to a secret

carbonin:
This allows us to remove one more thing from the region PV and gives us sharing of the v2_key for free.

This gets us closer to removing the shared region PV entirely

### DIFF
--- a/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
+++ b/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
@@ -29,6 +29,15 @@ PV_LOG_TIMESTAMP="$(date +%s)"
 # VMDB shared REGION app_root directory on PV
 PV_REGION_VMDB="${PV_CONTAINER_DATA_REGION_DIR}/var/www/miq/vmdb"
 
+function write_v2_key() {
+  echo "== Writing encryption key =="
+  cat > /var/www/miq/vmdb/certs/v2_key << KEY
+---
+:algorithm: aes-256-cbc
+:key: ${V2_KEY}
+KEY
+}
+
 # Inspect PV for previous deployments, if a DB a config is present, restore
 # Source previous deployment info file from PV and compare data with current environment
 # Evaluate conditions and decide a target deployment type: redeploy,upgrade or new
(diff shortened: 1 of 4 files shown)


templates/miq-template-ext-db.yaml
```diff
@@ -20,6 +20,7 @@ objects:
   stringData:
     pg-password: "${DATABASE_PASSWORD}"
     database_url: "postgresql://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_SERVICE_NAME}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5"
+    v2-key: "${V2_KEY}"
```
bdunne: Are we able to stay consistent with `_` instead of `-` here?

carbonin: Given the issues we had with this, I would be more likely to move `database_url` to `database-url`.

Sound good?

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -29,6 +29,15 @@ PV_LOG_TIMESTAMP="$(date +%s)"
...
 
+function write_v2_key() {
+  echo "== Writing encryption key =="
+  cat > /var/www/miq/vmdb/certs/v2_key << KEY
+---
+:algorithm: aes-256-cbc
+:key: ${V2_KEY}
+KEY
+}
+
```
fbladilo: @carbonin Do we need to set special file perms on this one?

carbonin: Nope the default should be fine.

templates/miq-template-ext-db.yaml
fbladilo: 👍

templates/miq-template-ext-db.yaml
```diff
@@ -19,7 +19,8 @@ objects:
     name: "${NAME}-secrets"
   stringData:
     pg-password: "${DATABASE_PASSWORD}"
```
fbladilo: @carbonin Perhaps it is time to also place the PG username into a secret, keep all credentials contained and safe in this object. Right now the user is exposed to anyone with access to the project where MIQ is deployed.

carbonin:
> Down to one file on region volume I believe (deployment-info) 

PR incoming 😉 
Hoping to get this put together and tested today.

templates/miq-template.yaml
```diff
@@ -19,7 +19,8 @@ objects:
     name: "${NAME}-secrets"
   stringData:
     pg-password: "${DATABASE_PASSWORD}"
-    database_url: "postgresql://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_SERVICE_NAME}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5"
+    database-url: "postgresql://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_SERVICE_NAME}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5"
+    v2-key: "${V2_KEY}"
```
Fryguy: @carbonin Minor, but I wonder if the key name and ENV var name should not have "v2" in them.  If we have to bump the key format, it will go to v3, and require all of this to change.  Not sure on a better name, but maybe `manageiq-encryption-key`?

PR #161, opened 2017-06-16
Remove the shared region PV

carbonin:
The only file left in the region PV is the `.deployment_info` file which is used mainly for saving off version numbers for use when determining the deployment scenario.

It turns out that determining the deployment type does not require saving this kind of data; it can be done using a combination of rake tasks and rails runner commands.

This PR first alters the `check_deployment_status` function so that it no longer relies on version variables from the `.deployment_info` file, then removes the remaining references to the `.deployment_info` file and, finally, the PV itself.

~Leaving this as WIP until I can properly test all of the deployment cases.~
Tested everything out. Good to go. 👍


images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -38,52 +29,34 @@ function write_v2_key() {
...
-    else
-      # Handle special master case
-      # Master version remains static, check DB schema status and proceed accordingly
+  cd "${APP_ROOT}"
 
-      if [[ ${APP_VERSION} == master ]]; then
-        # Go for redeployment case unless rake task returns 1 (pending migrations)
+  SCHEMA_VERSION="$(RAILS_USE_MEMORY_STORE=true bin/rake db:version | grep "Current version" | awk '{ print $3 }')"
+  echo "Current schema version is $SCHEMA_VERSION"
+  if [ "$SCHEMA_VERSION" == "0" ]; then
```
carbonin: I'm starting to wonder if all of this logic might be better if it was a self-contained rake task.

That way we could cover it with CI.
The only issue I'm not sure about is how to communicate the result.

As @bdunne and I started looking into logging to STDOUT for the aggregated logging effort we had a few issues parsing strings from rake output. Maybe we should define a return code enum of some sort if we go that route.

Thoughts?

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -301,7 +207,7 @@ function backup_pv_data() {
   (
     echo "== Initializing PV data backup =="
 
-    rsync -av --exclude 'log' "${PV_CONTAINER_DATA_DIR}" "${PV_CONTAINER_DATA_REGION_DIR}" "${PV_BACKUP_DIR}/backup_${PV_BACKUP_TIMESTAMP}"
+    rsync -av --exclude 'log' "${PV_CONTAINER_DATA_DIR}" "${PV_BACKUP_DIR}/backup_${PV_BACKUP_TIMESTAMP}"
```
carbonin: I think I made the right edit here.

Keep me honest @fbladilo

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
bdunne: Maybe we can just write it all in ruby instead?

chessbyte: Or Crystal

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
fbladilo: @carbonin Looks good, sending everything except logs from server pv data container data dir to backup dir on the same PV. Before we were processing both PVs on the same rsync one-liner.

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
fbladilo: A rake task would be cleaner, bash is effective but this is a probably the most critical procedure during the init stage, deciding what we the status of the deployment is and how to proceed accordingly.

carbonin: Opened a PR on the core repo [here](https://github.com/ManageIQ/manageiq/pull/15402)

Will change this code to reference that instead.

PR #162, opened 2017-06-21
Add a httpd front end

bdunne:
This httpd image will be the front-end for all incoming requests.
- It will handle SSL for now (until it is moved into the route)
- It will proxy to the correct services.
- In the future it will handle all of the auth as well.

### DIFF
--- a/images/miq-app/docker-assets/appliance-initialize.sh
+++ b/images/miq-app/docker-assets/appliance-initialize.sh
@@ -24,13 +24,13 @@ restore_pv_data
 cd ${APP_ROOT}
 bin/rake evm:deployment_status
 
+# Generate httpd certificate
+/usr/bin/generate_miq_server_cert.sh
+
 # Select path of action based on DEPLOYMENT_STATUS value
 case $? in
   3) # new_deployment
     echo "== Starting New Deployment =="
-    # Generate the certs
-    /usr/bin/generate_miq_server_cert.sh
-
     # Setup logs on PV before init
     setup_logs
 
@@ -48,7 +48,6 @@ case $? in
   ;;
   4) # new_replica
     echo "== Starting New Replica =="
-    /usr/bin/generate_miq_server_cert.sh
     setup_logs
     replica_join_region
     sync_pv_data
--- a/templates/miq-template-ext-db.yaml
+++ b/templates/miq-template-ext-db.yaml
@@ -56,7 +56,7 @@ objects:
 - apiVersion: v1
   kind: Route
   metadata:
-    name: ${NAME}
+    name: "${HTTPD_SERVICE_NAME}"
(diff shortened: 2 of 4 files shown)


templates/miq-template-ext-db.yaml
```diff
@@ -265,6 +265,87 @@ objects:
...
           name: "postgresql"
+
+
+- apiVersion: v1
+  kind: ImageStream
+  metadata:
+    name: httpd
+    annotations:
+      description: "Keeps track of the httpd image changes"
+      service.alpha.openshift.io/dependencies: '[{"name":"${NAME}","namespace":"","kind":"Service"}]'
```
carbonin: I think this should live with the service, not the `ImageStream`

templates/miq-template-ext-db.yaml
```diff
@@ -265,6 +265,87 @@ objects:
...
+                path: /
+                port: 443
+                scheme: HTTPS
+              initialDelaySeconds: 10
+              timeoutSeconds: 3
+            volumeMounts: []
+            env:
+            resources:
+              requests:
+              limits:
```
carbonin: Did you want to add values here? Otherwise just remove these keys.

images/miq-app/docker-assets/appliance-initialize.sh
```diff
@@ -24,13 +24,13 @@ restore_pv_data
 cd ${APP_ROOT}
 bin/rake evm:deployment_status
 
+# Generate httpd certificate
+/usr/bin/generate_miq_server_cert.sh
```
carbonin: Shouldn't this go away entirely if we are doing it in the apache pod? Or do we still need it here for some reason?

bdunne: I think it should, but it wasn't as easy as I originally thought it would be.  So, I'd like to do that in a follow-up PR

templates/miq-template-ext-db.yaml
bdunne: Good call :)

images/miq-app/docker-assets/appliance-initialize.sh
carbonin: Isn't this change making us regenerate the cert when it isn't necessary? Now we'll generate a cert even on straight redeploy ... I'm not sure what this commit it doing for this PR.

carbonin: Nevermind it only generates the certs if they don't exist already, which they won't now that we are not saving them but we are not even checking them now that the httpd pod is providing the service attached to the route.

So this is fine for now as long as we take care of it in the future.

bdunne:
@fbladilo No special service account is required.

PR #163, opened 2017-06-21
Add ansible container

carbonin:
This will add support for "embedded ansible" as a service in our project.

In combination with https://github.com/ManageIQ/manageiq/pull/15423 this should allow the embedded ansible role to work in the podified application just as it does in the appliance.

We add the miq-sysadmin scc here so that the ansible pod can run systemd without requiring the full list of entitlements that the existing privileged scc grants.


images/miq-app/Dockerfile
```diff
@@ -76,7 +76,7 @@ RUN mkdir -p ${APP_ROOT} && \
 
 ## Add ManageIQ source from local directory (dockerfile development) or from Github (official build)
 #ADD . ${APP_ROOT}
-RUN curl -L https://github.com/ManageIQ/manageiq/tarball/${REF} | tar vxz -C ${APP_ROOT} --strip 1
+RUN curl -L https://github.com/carbonin/manageiq/tarball/use_ansible_service_in_containers | tar vxz -C ${APP_ROOT} --strip 1
```
carbonin: This will go away once https://github.com/ManageIQ/manageiq/pull/15423 is merged.

templates/miq-template-ext-db.yaml
```diff
@@ -134,6 +154,12 @@ objects:
                 secretKeyRef:
                   name: "${NAME}-secrets"
                   key: "v2-key"
+            -
+              name: "ANSIBLE_SERVICE_NAME"
+              value: "${ANSIBLE_SERVICE_NAME}"
+            -
+              name: "ANSIBLE_ADMIN_PASSWORD"
+              value: "${ANSIBLE_ADMIN_PASSWORD}"
```
bdunne: Should we be pulling this value from the Ansible secret?

templates/miq-template.yaml
```diff
@@ -155,6 +175,12 @@ objects:
                 secretKeyRef:
                   name: "${NAME}-secrets"
                   key: "v2-key"
+            -
+              name: "ANSIBLE_SERVICE_NAME"
+              value: "${ANSIBLE_SERVICE_NAME}"
+            -
+              name: "ANSIBLE_ADMIN_PASSWORD"
+              value: "${ANSIBLE_ADMIN_PASSWORD}"
```
bdunne: Same

PR #164, opened 2017-06-22
Add the db service back to the app pod in the external db template

carbonin:
This was removed in e496ce5ba305a74f3c2233414352b737e7453fc5 but is still needed


carbonin:
This fixes #166

fbladilo:
@carbonin 👍

PR #170, opened 2017-06-26
Drop references to /persistent-region volume

bdunne:


fbladilo:
@bdunne 👍 Yeah I noticed the mountpaths still being created on latest builds.

PR #173, opened 2017-06-29
Split front-end and back-end containers

bdunne:
Depends on https://github.com/ManageIQ/manageiq/pull/15470

The frontend ManageIQ pod will receive all web traffic, all frontend replicas must run the UI, API and WS roles (no change here).

Backend replicas will not receive any UI / API / WS requests, so those roles should not be enabled by default.  Backend replicas should also wait for the frontend service to become available before attempting to start the application.  Backend replicas should be set to 0 by default and users can scale it based on their needs.

This change allows us to scale with a smaller footprint since we usually need more MiqServers for operations and inventory roles than for UI/API roles.

### DIFF
--- a/images/miq-app-frontend/Dockerfile
+++ b/images/miq-app-frontend/Dockerfile
@@ -1,0 +1,69 @@
+FROM manageiq/manageiq-pods:backend-latest
+MAINTAINER ManageIQ https://github.com/ManageIQ/manageiq-appliance-build
+
+## Set build ARGs
+ARG REF=master
+
+## Set ENV, LANG only needed if building with docker-1.8
+ENV SUI_ROOT=/opt/manageiq/manageiq-ui-service
+
+## Atomic/OpenShift Labels
+LABEL name="manageiq" \
+      vendor="ManageIQ" \
+      version="Master" \
+      release=${REF} \
+      url="http://manageiq.org/" \
+      summary="ManageIQ appliance image" \
+      description="ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures." \
+      io.k8s.display-name="ManageIQ" \
+      io.k8s.description="ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures." \
+      io.openshift.expose-services="443:https" \
+      io.openshift.tags="ManageIQ,miq,manageiq"
+
+## Install EPEL repo, yum necessary packages for the build without docs, clean all caches
+RUN yum -y install centos-release-scl-rh && \
+    yum -y install --setopt=tsflags=nodocs \
+                   httpd                   \
+                   mod_auth_kerb           \
+                   mod_authnz_pam          \
+                   mod_intercept_form_submit \
+                   mod_lookup_identity     \
+                   mod_ssl                 \
+                   &&                      \
+    yum clean all
+
+## GIT clone manageiq-appliance and service UI repo (SUI)
+RUN mkdir -p ${SUI_ROOT} && \
+    curl -L https://github.com/ManageIQ/manageiq-ui-service/tarball/${REF} | tar vxz -C ${SUI_ROOT} --strip 1
+
+## Setup environment
+RUN mv /etc/httpd/conf.d/ssl.conf{,.orig} && \
+    echo "# This file intentionally left blank. ManageIQ maintains its own SSL configuration" > /etc/httpd/conf.d/ssl.conf
+
+## Change workdir to application root, build/install gems
+WORKDIR ${APP_ROOT}
+RUN source /etc/default/evm && \
+    export RAILS_USE_MEMORY_STORE="true" && \
+    rake update:bower && \
+    bin/rails log:clear tmp:clear && \
+    rake evm:compile_assets && \
+    # Cleanup install artifacts
+    npm cache clean && \
+    bower cache clean && \
+    rm -rvf ${APP_ROOT}/tmp/cache/assets && \
+    rm -vf ${APP_ROOT}/log/*.log
+
+## Build SUI
+RUN source /etc/default/evm && \
+    cd ${SUI_ROOT} && \
+    yarn install --production && \
+    yarn run build && \
+    yarn cache clean
+
+## Expose required container ports
+EXPOSE 80 443
+
+COPY docker-assets/check-dependent-services.sh /bin
+
+ENTRYPOINT ["/usr/local/bin/dumb-init", "--single-child", "--"]
+CMD ["entrypoint"]
--- a/images/miq-app/docker-assets/entrypoint
+++ b/images/miq-app/docker-assets/entrypoint
@@ -2,6 +2,7 @@
 
 /usr/sbin/crond &
 
+check-dependent-services.sh
 appliance-initialize.sh
 
 [[ -s /etc/default/evm ]] && source /etc/default/evm
--- a/templates/miq-template-ext-db.yaml
+++ b/templates/miq-template-ext-db.yaml
@@ -107,7 +107,7 @@ objects:
       spec:
         containers:
         - name: manageiq
-          image: "${APPLICATION_IMG_NAME}:${APPLICATION_IMG_TAG}"
+          image: "${APPLICATION_IMG_NAME}:${FRONTEND_APPLICATION_IMG_TAG}"
           livenessProbe:
             tcpSocket:
               port: 443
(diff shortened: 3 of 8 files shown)


images/miq-app-frontend/Dockerfile
```diff
@@ -0,0 +1,69 @@
+FROM bdunne/miq-pods-test:backend-latest
```
carbonin: I think we'll need this changed before we merge.

Also, on that note, should we change the directory in this repo to `manageiq-app-backend`?

images/miq-app-frontend/Dockerfile
```diff
@@ -0,0 +1,69 @@
...
+
+## Setup environment
+RUN mv /etc/httpd/conf.d/ssl.conf{,.orig} && \
+    echo "# This file intentionally left blank. ManageIQ maintains its own SSL configuration" > /etc/httpd/conf.d/ssl.conf
+
+## Change workdir to application root, build/install gems
+WORKDIR ${APP_ROOT}
+RUN source /etc/default/evm && \
+    export RAILS_USE_MEMORY_STORE="true" && \
+    rake update:bower && \
```
carbonin: Can this stuff come out of the other Dockerfile?

images/miq-app/docker-assets/entrypoint
```diff
@@ -2,6 +2,7 @@
 
 /usr/sbin/crond &
 
+container-initialize.sh
```
carbonin: I think we can do better naming this. Maybe `check-dependent-services.sh`?

templates/miq-template-ext-db.yaml
```diff
@@ -192,6 +192,108 @@ objects:
...
+
+
+- apiVersion: apps/v1beta1
+  kind: "StatefulSet"
+  metadata:
+    name: ${NAME}-backend
+    annotations:
+      description: "Defines how to deploy the ManageIQ appliance"
+  spec:
+    serviceName: "${NAME}-backend"
```
carbonin: Don't think this has a service right?

templates/miq-template-ext-db.yaml
```diff
@@ -107,7 +107,7 @@ objects:
       spec:
         containers:
         - name: manageiq
```
carbonin: Do we also want to change this dc and container to be named `manageiq-frontend`?

templates/miq-template-ext-db.yaml
```diff
@@ -192,6 +192,108 @@ objects:
           resources:
             requests:
               storage: "${APPLICATION_VOLUME_CAPACITY}"
+
+
+- apiVersion: apps/v1beta1
+  kind: "StatefulSet"
+  metadata:
+    name: ${NAME}-backend
```
carbonin: This is user-visible. Is "backend" something they would understand? Maybe "engine"? I don't know what would be the best name here.

templates/miq-template-ext-db.yaml
```diff
@@ -192,6 +192,108 @@ objects:
...
+    template:
+      metadata:
+        labels:
+          name: ${NAME}-backend
+        name: ${NAME}-backend
+      spec:
+        containers:
+        - name: manageiq
+          image: "${APPLICATION_IMG_NAME}:${BACKEND_APPLICATION_IMG_TAG}"
+          imagePullPolicy: Always
```
carbonin: Do we want this? We don't set this for the other app pod.

templates/miq-template-ext-db.yaml
```diff
@@ -192,6 +192,108 @@ objects:
...
+            preStop:
+              exec:
+                command:
+                  - /opt/manageiq/container-scripts/sync-pv-data
+        serviceAccount: miq-anyuid
+        serviceAccountName: miq-anyuid
+        terminationGracePeriodSeconds: 90
+    volumeClaimTemplates:
+      - metadata:
+          name: "${NAME}-server"
```
carbonin: Does this need to be different?

templates/miq-template-ext-db.yaml
```diff
@@ -672,10 +774,15 @@ parameters:
     description: "This is the Application image name requested to deploy."
     value: "docker.io/manageiq/manageiq-pods"
   -
-    name: "APPLICATION_IMG_TAG"
-    displayName: "Application Image Tag"
+    name: "FRONTEND_APPLICATION_IMG_TAG"
+    displayName: "Front end Application Image Tag"
+    description: "This is the Application image tag/version requested to deploy."
```
carbonin: I think the description should probably also change.

templates/miq-template-ext-db.yaml
```diff
@@ -192,6 +192,108 @@ objects:
...
+            # Uncomment this if using dynamic volume provisioning.
+            # https://docs.openshift.org/latest/install_config/persistent_storage/dynamically_provisioning_pvs.html
+            # volume.alpha.kubernetes.io/storage-class: anything
+        spec:
+          accessModes: [ ReadWriteOnce ]
+          resources:
+            requests:
+              storage: "${APPLICATION_VOLUME_CAPACITY}"
+
+
```
carbonin: Bonus whitespace?

templates/miq-template-ext-db.yaml
```diff
@@ -192,6 +192,108 @@ objects:
...
+                  key: "database-url"
+            -
+              name: "MIQ_SERVER_DEFAULT_ROLES"
+              value: "database_operations,event,reporting,scheduler,smartstate,ems_operations,ems_inventory,automate"
+            -
+              name: "FRONTEND_SERVICE_NAME"
+              value: "${NAME}"
+            -
+              name: "MEMCACHED_SERVER"
+              value: "${MEMCACHED_SERVICE_NAME}:11211"
```
carbonin: Do we need this in the backend?

templates/miq-template-ext-db.yaml
```diff
@@ -192,6 +192,108 @@ objects:
...
+            -
+              name: "MEMCACHED_SERVICE_NAME"
+              value: "${MEMCACHED_SERVICE_NAME}"
```
carbonin: Same, I'm not sure we need this.

templates/miq-template-ext-db.yaml
```diff
@@ -192,6 +192,108 @@ objects:
...
+          env:
+            -
+              name: "APPLICATION_INIT_DELAY"
+              value: "${APPLICATION_INIT_DELAY}"
+            -
+              name: "DATABASE_SERVICE_NAME"
+              value: "${DATABASE_SERVICE_NAME}"
+            -
+              name: "DATABASE_REGION"
+              value: "${DATABASE_REGION}"
```
carbonin: If this pod is never going to initialize the region I don't think we need to provide this. I think the region id will be detected using the existing database sequence.

images/miq-app-frontend/docker-assets/container-initialize.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# Source OpenShift scripting env
+[[ -s ${CONTAINER_SCRIPTS_ROOT}/container-deploy-common.sh ]] && source "${CONTAINER_SCRIPTS_ROOT}/container-deploy-common.sh"
+
+# Check readiness of external services
+check_svc_status ${MEMCACHED_SERVICE_NAME} 11211
+check_svc_status ${DATABASE_SERVICE_NAME} 5432
+
```
carbonin: Bonus whitespace 😉

images/miq-app-frontend/Dockerfile
bdunne: I wish, but `manageiq-ui-classic`

templates/miq-template-ext-db.yaml
bdunne: I don't think so.  I don't want to give it a name that suggests it is limited to only serving the UI / API / WS roles.

templates/miq-template-ext-db.yaml
bdunne: "engine" to me suggests that it is a required service, I don't have any better suggestions either.

templates/miq-template-ext-db.yaml
bdunne: I've been adding a break between some of these things for readability

bdunne: These files are impossible

templates/miq-template-ext-db.yaml
bdunne: Honestly, I'm not sure.

PR #174, opened 2017-06-30
Allow server to control ansible pods

carbonin:
This PR adds a new service account for the manageiq application pod which is intended to have edit and view roles for the namespace the project is running in.

This will allow the app pod itself to scale other pods within the project up and down.

The plan is to use this ability to only scale the ansible pod when it is needed (when the role is enabled).

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -426,7 +426,7 @@ objects:
     strategy:
       type: "Recreate"
     serviceName: "${ANSIBLE_SERVICE_NAME}"
-    replicas: 1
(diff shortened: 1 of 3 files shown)


templates/miq-template.yaml
```diff
@@ -149,6 +153,11 @@ objects:
                 mountPath: "/persistent"
           env:
             -
+              name: MY_POD_NAMESPACE
```
bdunne: I think you need this in the other template as well.

carbonin: Done

PR #175, opened 2017-07-03
If serviceName is not provided, it gets set to '' which is invalid

bdunne:
Without this, you get a not so helpful error:
```
pet: manageiq-backend-0, error: Pod "manageiq-backend-0" is invalid: metadata.annotations[pod.beta.kubernetes.io/subdomain]: Invalid value: "": must match the regex [a-z0-9]([-a-z0-9]*[a-z0-9])? (e.g. 'my-name' or '123-abc')
```

### DIFF
--- a/templates/miq-template-ext-db.yaml
+++ b/templates/miq-template-ext-db.yaml
@@ -203,6 +203,16 @@ objects:
               storage: "${APPLICATION_VOLUME_CAPACITY}"
 
 
+- apiVersion: v1
+  kind: Service
+  metadata:
+    annotations:
+      description: "Headless service for ManageIQ backend pods"
+    name: ${NAME}-backend
+  spec:
+    clusterIP: None
+    selector:
+      name: ${NAME}-backend
 - apiVersion: apps/v1beta1
   kind: "StatefulSet"
   metadata:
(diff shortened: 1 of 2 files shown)


carbonin:
Weird, so replicas are tied to service names? I would think we would be able to scale without having a service. (Especially because these don't have a service)

I would make sure that this doesn't route the actual `manageiq` service to these pods as well.

carbonin:
@fbladilo This seems odd to me. Can you take a look?

bdunne:
I did a little bit of clicking around in the ManageIQ UI after scaling up and it didn't seem to be trying to load balance to that server, but I don't know how to tell for sure.

templates/miq-template-ext-db.yaml
```diff
@@ -203,13 +203,25 @@ objects:
...
+- apiVersion: v1
+  kind: Service
+  metadata:
+    annotations:
+      description: "Headless service for ManageIQ backend pods"
+      service.alpha.openshift.io/dependencies: '[{"name":"${NAME}","namespace":"","kind":"Service"}]'
+    name: ${NAME}-backend
+  spec:
+    clusterIP: None
+    selector:
```
carbonin: I think we don't want selectors here.

From the doc:

>For headless services that define selectors, the endpoints controller creates Endpoints records in the API, and modifies the DNS configuration to return A records (addresses) that point directly to the Pods backing the Service.

I don't think we need any of that because we won't actually be addressing this service.

PR #176, opened 2017-07-03
Pass ENV variables for service endpoints

bdunne:
Based on changes in: https://github.com/ManageIQ/container-httpd/pull/6


carbonin:
Need to add this for the external db template as well.

bdunne:
@carbonin Done.  Also, removed the `ANSIBLE_SERVICE_NAME` as discussed.

PR #185, opened 2017-07-21
Allow PG MIQ configuration overrides via configmap

fbladilo:
@carbonin @bdunne Please review guys..

- Configmap object created with MIQ default overrides settings
- Configmap is injected as a volume into /var/lib/pgsql/conf.d by default
- POSTGRESQL_CONFIG_DIR supplied to on template to allow flexibility on dir location
- Related to PR : https://github.com/ManageIQ/container-postgresql/pull/2
- PG image will pickup and do an include_dir if the directory is found

The most important piece is that the configuration is externalized and parametized. A user for instance can oc edit configmap , change parameter, oc rollout latest postgresql and be done. Configs can be stacked and multiple *.confs can be placed in directory , PG will process them all.

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -38,6 +38,79 @@ objects:
     secret-key: "${ANSIBLE_SECRET_KEY}"
     admin-password: "${ANSIBLE_ADMIN_PASSWORD}"
 - apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: "${DATABASE_SERVICE_NAME}-configs"
+  data:
+    01_miq_overrides.conf: |
+      #------------------------------------------------------------------------------
+      # CONNECTIONS AND AUTHENTICATION
+      #------------------------------------------------------------------------------
+
+      tcp_keepalives_count = 9
+      tcp_keepalives_idle = 3
+      tcp_keepalives_interval = 75
+
+      #------------------------------------------------------------------------------
+      # RESOURCE USAGE (except WAL)
+      #------------------------------------------------------------------------------
+
+      shared_preload_libraries = 'pglogical,repmgr_funcs'
+      max_worker_processes = 10
+
+      #------------------------------------------------------------------------------
+      # WRITE AHEAD LOG
+      #------------------------------------------------------------------------------
+
+      wal_level = 'logical'
+      wal_log_hints = on
+      wal_buffers = 16MB
+      checkpoint_completion_target = 0.9
+
+      #------------------------------------------------------------------------------
+      # REPLICATION
+      #------------------------------------------------------------------------------
+
+      max_wal_senders = 10
+      wal_sender_timeout = 0
+      max_replication_slots = 10
+      hot_standby = on
+
+      #------------------------------------------------------------------------------
+      # ERROR REPORTING AND LOGGING
+      #------------------------------------------------------------------------------
+
+      log_filename = 'postgresql.log'
+      log_rotation_age = 0
+      log_min_duration_statement = 5000
+      log_connections = on
+      log_disconnections = on
+      log_line_prefix = '%t:%r:%c:%u@%d:[%p]:'
+      log_lock_waits = on
+
+      #------------------------------------------------------------------------------
+      # AUTOVACUUM PARAMETERS
+      #------------------------------------------------------------------------------
+
+      log_autovacuum_min_duration = 0
+      autovacuum_naptime = 5min
+      autovacuum_vacuum_threshold = 500
+      autovacuum_analyze_threshold = 500
+      autovacuum_vacuum_scale_factor = 0.05
+
+      #------------------------------------------------------------------------------
+      # LOCK MANAGEMENT
+      #------------------------------------------------------------------------------
+
+      deadlock_timeout = 5s
+
+      #------------------------------------------------------------------------------
+      # VERSION/PLATFORM COMPATIBILITY
+      #------------------------------------------------------------------------------
+
+      escape_string_warning = off
+      standard_conforming_strings = off
+- apiVersion: v1
   kind: Service
   metadata:
     annotations:
@@ -410,6 +483,9 @@ objects:
         - name: miq-pgdb-volume
           persistentVolumeClaim:
             claimName: "${NAME}-${DATABASE_SERVICE_NAME}"
+        - name: miq-pg-configs
+          configMap:
+            name: "${DATABASE_SERVICE_NAME}-configs"
         containers:
         - name: postgresql
           image: "${POSTGRESQL_IMG_NAME}:${POSTGRESQL_IMG_TAG}"
@@ -432,6 +508,8 @@ objects:
           volumeMounts:
           - name: miq-pgdb-volume
             mountPath: "/var/lib/pgsql/data"
+          - name: miq-pg-configs
+            mountPath: "${POSTGRESQL_CONFIG_DIR}"
           env:
           - name: POSTGRESQL_USER
             value: "${DATABASE_USER}"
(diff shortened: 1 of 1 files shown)


templates/miq-template.yaml
```diff
@@ -691,10 +771,14 @@ parameters:
   displayName: Memcached Slab Page Size
   description: Memcached size of each slab page.
   value: 1m
+- name: POSTGRESQL_CONFIG_DIR
+  displayName: PostgreSQL Configuration Overrides
+  description: Directory used to store PostgreSQL configuration overrides.
+  value: '/etc/manageiq/postgresql.conf.d'
```
carbonin: If these are changed to double quotes I think the specs will pass.

fbladilo: I'll test on openshift see it the parser likes it, I recall issues with double quotes.

templates/miq-template.yaml
```diff
@@ -691,10 +771,14 @@ parameters:
   displayName: Memcached Slab Page Size
   description: Memcached size of each slab page.
   value: 1m
+- name: POSTGRESQL_CONFIG_DIR
+  displayName: PostgreSQL Configuration Overrides
+  description: Directory used to store PostgreSQL configuration overrides.
+  value: "/etc/manageiq/postgresql.conf.d"
```
carbonin: If this directory doesn't exist, will the config map mount create the parent directory?

carbonin:
Almost forgot about these settings:

From https://github.com/ManageIQ/manageiq-gems-pending/blob/6bdd64449485857526c968a119d55a7edba6e015/lib/gems/pending/appliance_console/internal_database_configuration.rb#L168-L176
```ruby
def apply_initial_configuration
  shared_buffers = run_as_evm_server ? SHARED_DB_SHARED_BUFFERS : DEDICATED_DB_SHARED_BUFFERS
  with_pg_connection do |conn|
    conn.exec("ALTER SYSTEM SET ssl TO on") if ssl
    conn.exec("ALTER SYSTEM SET shared_buffers TO #{shared_buffers}")
  end

  restart_postgres
end
```

I don't think we're doing anything with ssl, but when an appliance runs as a database only (not being also used as a ManageIQ application server), we set `shared_buffers` to `1GB`. Do we also want to set that as one of the environment variables in the container?

It's already configurable in the [template](https://github.com/sclorg/postgresql-container/blob/master/9.5/root/usr/share/container-scripts/postgresql/openshift-custom-postgresql.conf.template#L18)

templates/miq-template.yaml
fbladilo: @carbonin Yep, the configmap volume plugin creates  the dir structure if it does not exist.

fbladilo:
@carbonin If we want to emulate what appliance_console does in these cases, we probably need to set SHARED_BUFFERS on template to 1GB by default, this will be the conservative option(IMHO the default 256MB  is already pretty low). Another alternative is alllow the user to increase SHARED_BUFFERS via configmap or environment variable whichever way suits best their case. oc set env or oc edit configmap can be used respectively, it requires rolling out PG afterwards to take the changes.

templates/miq-template.yaml
```diff
@@ -446,6 +524,8 @@ objects:
             value: "${POSTGRESQL_MAX_CONNECTIONS}"
           - name: POSTGRESQL_SHARED_BUFFERS
             value: "${POSTGRESQL_SHARED_BUFFERS}"
+          - name: POSTGRESQL_CONFIG_DIR
```
bdunne: Is this env variable needed?

fbladilo: @bdunne This allows more flexibility, as user could potentially change the default location on template and we will process accordingly via image scripts.

carbonin: I feel like this is a lot less likely to change now that we are specifying it at build time of our PG image.

@fbladilo Do you think people are likely to rebuild the image to change this?

templates/miq-template.yaml
```diff
@@ -691,10 +771,14 @@ parameters:
   displayName: Memcached Slab Page Size
   description: Memcached size of each slab page.
   value: 1m
+- name: POSTGRESQL_CONFIG_DIR
+  displayName: PostgreSQL Configuration Overrides
+  description: Directory used to store PostgreSQL configuration overrides.
+  value: "/var/lib/pgsql/conf.d"
```
carbonin: To my point about this not changing, this probably doesn't even need to be a parameter.

templates/miq-template.yaml
```diff
@@ -691,10 +771,14 @@ parameters:
...
 - name: POSTGRESQL_MAX_CONNECTIONS
   displayName: PostgreSQL Max Connections
   description: PostgreSQL maximum number of database connections allowed.
-  value: '100'
+  value: '1000'
```
carbonin: :+1:

templates/miq-template.yaml
```diff
@@ -691,10 +771,14 @@ parameters:
...
 - name: POSTGRESQL_SHARED_BUFFERS
```
carbonin: Did we want to change this in this PR also @fbladilo ?

fbladilo: @carbonin I noticed our default MIQ overrides had it at 1000 and went ahead for the change. Let me know if you prefer a different PR for this setting and I'll revert it.

templates/miq-template.yaml
fbladilo: @carbonin I could see some cases where you could want to supply your own location for your stacked configs (perhaps during troubleshoot/debug), also allow us to dynamically move to whatever directory SCL eventually decide for drop in config-customs. I don't think it hurts. My 2 cents.

templates/miq-template.yaml
carbonin: This comment was for the shared buffers, not the connections.

templates/miq-template.yaml
fbladilo: @carbonin They do not need to rebuild the image to change this, kubernetes will allow the configmap to be placed anywhere on the filesystem, it has special privileges. The baked default directory is handy when things go wrong with the configmap, when is either missing or deleted, or also older builds such as euwe which will begin using newer PG images that support conf.d but their templates are old and do not supply the configmap yet.

carbonin: > They do not need to rebuild the image to change this

I think they would.
They would need to create the new directory and set the permissions on the location like you did in the dependent PR right?

templates/miq-template.yaml
fbladilo: @carbonin 😄 Sure we can go ahead with the 1GB SHARED_BUFFERS default change for this one, I'll commit in a moment.

templates/miq-template.yaml
fbladilo: @carbonin Consider this custom deployment : 

```
oc new-app --template=manageiq -p POSTGRESQL_IMG_NAME=docker.io/fbladilo/miq-postgresql -p POSTGRESQL_CONFIG_DIR=/etc/blah

oc describe pods postgresql-1-fszjq | grep -A4 miq-pg-config
      /etc/blah from miq-pg-configs (rw)
      /var/lib/pgsql/data from miq-pgdb-volume (rw)
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-zlxmm (ro)

sh-4.2$ ls -la /etc/blah
total 8
drwxrwsrwx.  3 root 1000060000   85 Jul 25 16:01 .
drwxr-xr-x. 50 root root       4096 Jul 25 16:01 ..
drwxr-sr-x.  2 root 1000060000   34 Jul 25 16:01 ..7987_25_07_12_01_10.985121115
lrwxrwxrwx.  1 root root         31 Jul 25 16:01 ..data -> ..7987_25_07_12_01_10.985121115
lrwxrwxrwx.  1 root root         28 Jul 25 16:01 01_miq_overrides.conf -> ..data/01_miq_overrides.conf

sh-4.2$ grep include_dir /var/lib/pgsql/data/userdata/postgresql.conf 
#include_dir = 'conf.d'			# include files ending in '.conf' from
include_dir '/etc/blah'

```
Kubernetes grants the configmap volume plugin elevated privileges to do this (assuming the configmap is in place and added into the POD spec). Our default directory creation on image allows older deployments to take advantage of the dependent PR, as they are likely to not have any of the required variables set or configmaps created. It is also a safe default for us to have it created under most circumstances, things can go wrong as I mentioned with configmaps missing or troubled.

carbonin: Okay, if that deployment works then I misunderstood the reasoning for changing the location of the conf.d directory.

I'm okay with this as is then 👍

PR #187, opened 2017-07-25
Increase default httpd recreate strategy timeout

fbladilo:
- Increase default recreate strategy to 1200 (20 mins) from 600
- Allows httpd to be more patient with frontend statefulset seeding

### DIFF
--- a/templates/miq-template-ext-db.yaml
+++ b/templates/miq-template-ext-db.yaml
@@ -498,6 +498,8 @@ objects:
   spec:
     strategy:
       type: Recreate
+      recreateParams:
+        timeoutSeconds: 1200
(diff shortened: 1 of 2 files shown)


fbladilo:
@carbonin @bdunne Please review guys, more often than not, the httpd DC decides to error out as the frontend statefulset is still finishing seeding. The default strategy timeout is absolute, meaning that the deployment will error out and not even try again to restart/redeploy. The 20 minutes allows it to be more patient with frontend , thus allowing it to finish working or giving it another chance to retry finishing.

fbladilo:
@bdunne Yes sir, I dream about it faster startups...

PR #188, opened 2017-07-25
Increase PG Pod resource requirements

fbladilo:
@carbonin @bdunne Follow up on this one...

- Provide reasonable minimum default resource requests for a PG/MIQ deployment on OpenShift
- Increased shared_buffers to 1GB
- Increased minimum RAM resource request to 4GB
- References :

https://wiki.postgresql.org/wiki/Tuning_Your_PostgreSQL_Server

https://access.redhat.com/documentation/en-us/red_hat_cloudforms/4.5/html/deployment_planning_guide/introduction#virtual-hardware-requirements

The kubernetes scheduler will only attempt a pod placement on a node that satisfies the minimum request, the shared buffer calculation is taken from the PG tuning guide at 1/4 of the available RAM. In this case, we calculate from the minimum memory request (guaranteed) and not the max memory limit set by default at 8GB and enforced via cgroups.

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -784,7 +784,7 @@ parameters:
 - name: POSTGRESQL_SHARED_BUFFERS
   displayName: PostgreSQL Shared Buffer Amount
   description: Amount of memory dedicated for PostgreSQL shared memory buffers.
-  value: 1024MB
+  value: 1GB
 - name: ANSIBLE_SERVICE_NAME
   displayName: Ansible Service Name
   description: The name of the OpenShift Service exposed for the Ansible container.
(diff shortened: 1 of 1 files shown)


templates/miq-template.yaml
```diff
@@ -841,7 +841,7 @@ parameters:
   displayName: PostgreSQL Min RAM Requested
   required: true
   description: Minimum amount of memory the PostgreSQL container will need.
-  value: 1024Mi
+  value: 4096Mi
```
carbonin: Can this be `4Gi`?

templates/miq-template.yaml
```diff
@@ -784,7 +784,7 @@ parameters:
 - name: POSTGRESQL_SHARED_BUFFERS
   displayName: PostgreSQL Shared Buffer Amount
   description: Amount of memory dedicated for PostgreSQL shared memory buffers.
-  value: 256MB
+  value: 1024MB
```
carbonin: Can this be `1GB`?

carbonin:
I'm okay with this. For development I'm already clearing out all the requests and limits and letting my minishift VM swap.

fbladilo:
Adjusted size suffixes to Gigs...

PR #194, opened 2017-08-10
Update to support foundational update of container-httpd for external authentication

abellotti:
Updating manageiq-pods so that it defines an miq-sysadmin scc and service account for the container-httpd pod.

This is needed for running httpd under systemd, a requirement for enabling external authentication.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,9 +84,35 @@ $ oc adm policy add-scc-to-user privileged system:serviceaccount:<your-namespace
 ```
 
 Verify that the miq-privileged service account is now included in the privileged scc
+
 ```
 $ oc describe scc privileged | grep Users
 Users:					system:serviceaccount:<your-namespace>:miq-privileged
+```
+
+### Add the miq-sysadmin service account
+
+_**Note:**_ The application front-end Httpd container requires an anyuid scc with the SYS_ADMIN capability to support systemd and dbus.
+
+__*As admin*__
+
+Create the miq-sysadmin SCC:
+
+```bash
+$ oc create -f templates/miq-sysadmin.yaml
+```
+
+The miq-sysadmin service account must be added to the miq-sysadmin SCC before the front-end Httpd pod can run.
+
+```bash
+$ oc adm policy add-scc-to-user miq-sysadmin system:serviceaccount:<your-namespace>:miq-sysadmin
+```
+
+Verify that the miq-sysadmin service account is now included in the miq-sysadmin scc
+
+```bash
+$ oc describe scc miq-sysadmin | grep Users
+Users:              system:serviceaccount:<your-namespace>:miq-sysadmin
 ```
 
 ### Add the view and edit roles to the orchestrator service account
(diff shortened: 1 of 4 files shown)


bazulay:
Why is systemd required  in the first place ?

+ nice read about CAP_SYS_ADMIN https://lwn.net/Articles/486306/

abellotti:
When external authentication is enabled, that will depend on SSSD which requires systemd.

README.md
```diff
@@ -84,11 +84,37 @@ $ oc adm policy add-scc-to-user privileged system:serviceaccount:<your-namespace
...
+Create the miq-sysadmin SCC:
+
+```bash
+$ oc create -f templates/miq-sysadmin.yaml
+```
+
+The miq-sysadmin service account must be added to the miq-sysadmin SCC before the front-end Httpd pod can run.
+
+```bash
+$ oc adm policy add-scc-to-user miq-sysadmin system:serviceaccount:<your-namespace>:miq-sysadmin
```
Fryguy: Can we not piggyback on an existing service account?

carbonin: Then everyone using that service account would also get the SYS_ADMIN capability.

I don't think that's what we want.

templates/miq-template.yaml
```diff
@@ -22,6 +22,10 @@ objects:
   metadata:
     name: miq-privileged
 - apiVersion: v1
+  kind: ServiceAccount
```
bdunne: Don't forget to make the same changes on the external db template.

abellotti: 👍

abellotti:
@carbonin correct, if they cannot add miq-sysadmin, then a privileged SCC will also work, I ran systemd/sssd/dbus early on as such successfully.

abellotti:
Something's odd with that travis failure, bundle exec rspec works fine locally.

templates/miq-template-ext-db.yaml
```diff
@@ -473,6 +477,49 @@ objects:
   spec:
     dockerImageRepository: "${HTTPD_IMG_NAME}"
 - apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: "${HTTPD_SERVICE_NAME}-configs"
+  data:
+    redirects.conf: |
+      # The following redirects files must be included to handle most specific to least specific URLs.
```
bdunne: Is this redirects.conf file valid?  This is all outside of a `<VirtualHost>` block.

bdunne: I think we should hold off on this until https://github.com/ManageIQ/manageiq-pods/pull/197 is merged.  It will significantly simplify the httpd config.

templates/miq-template-ext-db.yaml
```diff
@@ -473,6 +477,49 @@ objects:
...
+      RewriteRule ^/ https://${NAME}%{REQUEST_URI} [P,QSA,L]
+      ProxyPassReverse / https://${NAME}/
+
+      ### WebSocket redirects
+      ProxyPass /ws ws://${NAME}/ws
+      ProxyPassReverse /ws ws://${NAME}/ws
+
+      ProxyPreserveHost on
+      RequestHeader set X_FORWARDED_PROTO 'https'
+    logging.conf: |
```
carbonin: Why are these separate files? I was thinking we would just supply the entire config in one file.

Issue #97, opened 2017-02-16
Make sure cockpit-ws is built into the PODified manageiq image


bazulay:
Per https://github.com/ManageIQ/manageiq/pull/12506
and https://github.com/ManageIQ/manageiq-appliance/pull/107

We need to make sure cockpit-ws if fully functional on the podified version.

simaishi:
Also need a new package: https://github.com/ManageIQ/manageiq-appliance-build/pull/192

bazulay:
@simaishi @fbladilo 

It is a bit more than just installing the rpm .... we need to make sure it actually works on the PODified deployment.

fbladilo:
@simaishi @bazulay A couple of things on this one : 

* We need at least the cockpit RPM included in monolithic and openshift images
* We actually git-clone appliance-build so once  https://github.com/ManageIQ/manageiq-appliance/pull/107 is merged , we should be good.

fbladilo:
@bazulay We wait on this one until cockpit related PRs are actually merged upstream in manageiq/manageiq-appliance repos.

fbladilo:
@carbonin @simaishi This one seems stale? Were the relevant cockpit PRs merged already? Any ideas how to validate this on podified?

bazulay:
@fbladilo  @gtanzillo  @carbonin @simaishi

The code was merged , but still not validated on the regular appliance,
When the feedback is provides we could validate it on the PODified.

fbladilo:
Thanks Barak, let us know so we can test and clean this one up.

chessbyte:
Seems like all the code got in, but never validated.  Closing this stale issue and will accept incoming bugs when someone finds issues.

PR #197, opened 2017-08-15
Drop all internal SSL

bdunne:
- Move SSL to the route
- Drop extra httpd modules
- Don't generate certificates

Must be merged at the same time as https://github.com/ManageIQ/container-httpd/pull/11


images/miq-app-frontend/Dockerfile
```diff
@@ -17,28 +17,21 @@ LABEL name="manageiq" \
...
     yum clean all
 
 ## GIT clone service UI repo (SUI)
 RUN mkdir -p ${SUI_ROOT} && \
     curl -L https://github.com/ManageIQ/manageiq-ui-service/tarball/${REF} | tar vxz -C ${SUI_ROOT} --strip 1
 
 ## Setup environment
-RUN mv /etc/httpd/conf.d/ssl.conf{,.orig} && \
-    echo "# This file intentionally left blank. ManageIQ maintains its own SSL configuration" > /etc/httpd/conf.d/ssl.conf
+RUN rm /etc/httpd/conf.d/ssl.conf && \
```
carbonin: `rm -f`?

images/miq-app-frontend/docker-assets/manageiq-http.conf
```diff
@@ -0,0 +1,29 @@
+## ManageIQ HTTP Virtual Host Context
+
+# Timeout: The number of seconds before receives and sends time out.
+Timeout 120
+
+# HTTP Start-up error log
+ErrorLog /var/www/miq/vmdb/log/apache/miq_apache.log
+
+# Disable this section if using HTTP only
```
carbonin: We are using HTTP only here right? Not sure what this applies to.

images/miq-app-frontend/docker-assets/manageiq-http.conf
```diff
@@ -0,0 +1,29 @@
...
+RewriteEngine On
+Options SymLinksIfOwnerMatch
+
+<VirtualHost *:80>
+  DocumentRoot /var/www/miq/vmdb/public
+  Include conf.d/manageiq-redirects-ui
+  Include conf.d/manageiq-redirects-ws
+  Include conf.d/manageiq-redirects-websocket
```
carbonin: I think we will probably still need cockpit here right?

I'm not sure if they are all tested in containers, but I think that's what https://github.com/ManageIQ/manageiq-pods/issues/97 is meant to address, so maybe we can investigate in a followup?

carbonin:
@bdunne will take care of the readme in a followup. Going to merge now so @abellotti can rebase on these changes.

PR #201, opened 2017-08-17
Enhancement to support the httpd authentication configuration map.

abellotti:
- Added new httpd-auth-configs config map to the templates
- mounted the new configmap for httpd pod as /etc/httpd/auth-conf.d
- updated teardown script to handle the new config map

### DIFF
--- a/README.md
+++ b/README.md
@@ -541,3 +541,101 @@ $ oc new-app --template=manageiq \
   -p APPLICATION_IMG_TAG=latest \
   ...
 ```
+
+## Configuring External Authentication
+Configuring the _httpd_ pod for external authentication is done by updating the _httpd-auth-configs_ configuration map to include all necessary config files and certificates. Upon startup, the _httpd_ pod overlays its files with the ones specified in the _auth-configuration.conf_ file in the configuration map. This is done by the entrypoint before _systemd_ starts so all services will have their updated configuration upon startup.
+
+The external authentication configuration file _auth-configuration.conf_ declares the following:
+
+* The authentication type, default is _internal_.
+* The list of files to overlay upon startup if type is other than _internal_.
+
+Syntax for the file is as follows:
+
+```
+# for comments
+type = internal
+file = basename1 target_path1 permission1
+file = basename2 target_path2 permission2
+```
+
+_internal_ is the default type, anything else is considered external. This could include strings like: ipa, ldap, active_directory, saml or simply custom.
+
+For the files to overlay on the _httpd_ pod, one _file_ directive is needed per file.
+
+* the _basename_ is the name of the source file in the configuration map.
+* _target\_path_ is the path of the file on the pod to over_write, i.e. _/etc/sssd/sssd.conf_
+* _permission_ is optional, by default files are copied using the pod's default umask, owner and group, so files are created as mode 644 owner root, group root.
+
+optional _permission_ can be specified as follows:
+
+* mode
+* mode:owner
+* mode:owner:group
+
+Reflecting the mode and ownership to set the copied files to.
+
+_Examples_:
+
+* 755
+* 640:root
+* 644:root:apache
+
+Binary files can be specified in the configuration map in their base64 encoded format with a basename having a _.base64_ extension. Such files are then converted back to binary as they are copied to their target path.
+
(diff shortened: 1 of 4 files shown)


carbonin:
@abellotti Can you add something to the README about how to specify the other external auth files? Should they be added to the same ConfigMap?

Also, is any guidance needed around what files should be added for which authentication types?

abellotti:
Stuff here should be generic, I can move the config map/config file syntax out of the templates into the README and maybe give a single example. Yes, they go to the same httpd-auth-configs map.

The guidance/list of files generated for each auth type will go in the new repo (scripts configuring the different ext-auth and creating the configmap that users can update here and restart their httpd pod), probably ManageIQ/container-httpd-auth-config, coming soon.

carbonin:
> will go in the new repo

Ah, I didn't know there was more. My concern was that this felt like we were leaving too much up to the user to figure out how to configure this stuff. It sounds like this repo will accomplish what I was looking for.

In that case a link from the README here to that repo with a sentence or two will probably be enough.

README.md
```diff
@@ -541,3 +541,101 @@ $ oc new-app --template=manageiq \
...
+Support for automatically generating authentication configuration maps for _httpd_ will be provided by
+[ManageIQ/container-httpd-auth-config](https://github.com/ManageIQ/container-httpd-auth-config). Please see the [README.md](https://github.com/ManageIQ/container-httpd-auth-config/blob/master/README.md) in that repo for further details.
+
+The generated authentication configuration map can then be defined in the _httpd_ pod and further customized as follows:
+
+```bash
+$ oc edit configmaps httpd-auth-configs
+```
+
+Then rebouncing the _httpd_ pod for the new authentication configuration to take effect.
```
carbonin: I think "redeploy the httpd pod" or "rollout a new httpd deployment" here rather than "rebouncing the httpd pod"

templates/miq-template.yaml
```diff
@@ -963,6 +978,10 @@ parameters:
   displayName: Apache Configuration Directory
   description: Directory used to store the Apache configuration files.
   value: "/etc/httpd/conf.d"
+- name: HTTPD_AUTH_CONFIG_DIR
+  displayName: External Authentication Configuration Directory
+  description: Directory used to store the exxternal authentication configuration files.
```
carbonin: Typo, `s/exxternal/external/`

abellotti: 👍 fixed.

abellotti:
hmm, might take a week or two to get the other repo up. I do need this merged in now though.  I could remove the reference for now until the other is live, then re-add that paragraph in the README.md then,  thoughts ?

carbonin:
Yeah, I think that's probably for the best.

When we have the other repo up and running we'll re-add the info about it.

abellotti:
👍 done.  https://github.com/ManageIQ/manageiq-pods/pull/201/commits/7d783bab26ba47c213b5e324ac85fb1d7c83fe71

abellotti:
ping @bdunne

README.md
```diff
@@ -541,3 +541,100 @@ $ oc new-app --template=manageiq \
   -p APPLICATION_IMG_TAG=latest \
   ...
 ```
+
+## Configuring External Authentication
+Configuring the _httpd_ pod for external authentication is done by updating the _httpd-auth-configs_ configuration map to include all necessary config files and certificates. Upon startup, the _httpd_ pod overlays its files with the ones specified in the _auth-configuration.conf_ file in the configuration map. This is done by the _initialize-httpd-auth_ service that runs before _httpd_.
```
Fryguy: Instead of italics can you use backticks, to make it preformatted?  I find it hard to read with italics.

templates/miq-template-ext-db.yaml
```diff
@@ -490,6 +490,16 @@ objects:
...
 - apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: "${HTTPD_SERVICE_NAME}-auth-configs"
+  data:
+    auth-type: internal
+    auth-configuration.conf: |
+      # External Authentication Configuration File
+      #
+      # For details on usage please see https://github.com/ManageIQ/manageiq-pods/blob/master/README.md
```
Fryguy: You can actually link directly to your new section with : 

`https://github.com/ManageIQ/manageiq-pods/blob/master/README.md#configuring-external-authentication`

Fryguy:
LGTM

Issue #191, opened 2017-08-02
Openshift master config changes


mberube9:
The requirement to change the Openshift master config file to deploy MIQ is not best practice.  

Any way we can avoid making this change:
**maxImagesBulkImportedPerRepository: 100**

This is an issue for automating the deployment of MIQ as it would require a cluster restart and approval for a cluster setting change.

fbladilo:
@mberube9 This one is difficult, I agree that in general restarting your master to have these settings applied is disruptive. The rationale behind the increase of maxImagesBulkImportedPerRepository is a two part : 

* We use DockerHub to auto-build upstream images, by default we build multiple MIQ releases which are expressed as tags : "latest" , "euwe" , "euwe-2" from their corresponding branches. We also different images from this main repo, an example of this are the frontend and backend images.

I'm not sure this is the case for other projects but the amount of images in our repos grows continuously as we publish new releases. As of today, the miq-app IS which sources from DH manageiq-pods repo has more than 30 images in it : 

```
[root@franco-ocp-master ~]# oc describe is miq-app | grep Unique
Unique Images:		36
```
* The default OpenShift maxImagesBulkImportedPerRepository value of 5 is low, this caused many random image pull failures for users during deployments, the IS will not import the tag that they were requesting and fail, we went ahead and documented this parameter increase.

Some possible solutions to alleviate these ocurrences : 

1. We request a default value increase for this parameter better suited for medium sized image repos. I think we should at least raise this issue. I would also point that it would be smarter to import the latest built (active) tags first, in that order. So during an import at least you will know OpenShift tried to get you the most active images up to the limit for that repo.

2. We split the application images in their own repos, pretty much the same as postgresql, ruby, etc. This will keep the number of tags/images to a more reasonable number, and also allow us to use the latest tag (which is treated specially in OpenShift) for all images that are tracking master. This is no guarantee we will not reach the import limit, eventually as we go through releases we will start populating the repos as well, perhaps maintaining up to certain release would make sense as policy.

More ideas/comments/corrections : @simaishi @carbonin @bdunne

simon3z:
@fbladilo IIUC this is "fixed" by https://github.com/ManageIQ/manageiq-pods/pull/205
Do you think we can close this or we can still do better than that?

carbonin:
I'm going to close this this as the requirement is no longer in place as of #205.

[...]

PR #205, opened 2017-08-22
Remove IS and imagechange trigger from templates

fbladilo:
@carbonin @bdunne @simon3z @mberube9 Removes the IS objects and related triggers from deployment configs. Some of controllers/deployments were not affected since they either don't currently use them or don't support them (i.e statefulsets). With the following changes image pulls will start immediately without trigger intervention. 

We never took much advantage of what IS can give us, specially considering they were configured with external docker repositories by default without actively tracking changes.

This resolves the long standing and very intrusive master node config change for maxImagesBulkImportedPerRepository.

Related issues :  #191 and https://github.com/openshift/openshift-ansible/issues/4822

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,21 +159,6 @@ miq-pv02   5Gi         RWO           Recycle         Available                  
 
 It is strongly suggested that you validate NFS share connectivity from an OpenShift node prior attemping a deployment.
 
-### Increase maximum number of imported images on ImageStream
-
-By default OpenShift will import 5 images per ImageStream, we build and use more than 5 images in our repos for MIQ deployments.
-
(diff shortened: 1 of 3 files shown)


simon3z:
> This resolves the long standing and very intrusive master node config change for maxImagesBulkImportedPerRepository.

Nice @fbladilo !
cc @joelddiaz @ilackarms

ilackarms:
lgtm.

@fbladilo for testing purposes, it's still possible to manually create build configs for custom images and then deploy from them using 

```
oc new-app --template=manageiq \
  -p APPLICATION_IMG_NAME=<your-docker-registry>:5000/<your-namespace>/<whatever> \
  -p APPLICATION_IMG_TAG=<whatever>
```

Correct?

carbonin:
@ilackarms Yeah, the parameters are still in place and are being used by the containers directly.

PR #207, opened 2017-08-25
[REARCH] Container orchestrator

carbonin:
This PR alters the pods repo such that it supports the container orchestrator model of manageiq.

One of the goals of the ManageIQ Rearch effort is to move our workers into individual pods, but have them controlled by what is now the `MiqServer` process.

This will be the only pod created by the template and the server will talk directly to OpenShift to create the workers. This PR works directly with the changes in https://github.com/ManageIQ/manageiq/pull/15884 to achieve this.

~~Don't even consider merging this :smile: I'm opening it to keep everyone on the same page about how the re-architecture is going to affect this repo after the G-release.~~ We can consider merging this now!


images/manageiq-base-worker/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM manageiq/manageiq-base:latest
+MAINTAINER ManageIQ https://manageiq.org
+
+LABEL name="manageiq-base-worker" \
```
bdunne: Can we drop the "manageiq-"?

images/manageiq-base/Dockerfile
```diff
@@ -1,25 +1,28 @@
...
     APP_ROOT=/var/www/miq/vmdb \
-    APP_ROOT_PERSISTENT=/persistent \
     APPLIANCE_ROOT=/opt/manageiq/manageiq-appliance \
+    SUI_ROOT=/opt/manageiq/manageiq-ui-service \
     CONTAINER_SCRIPTS_ROOT=/opt/manageiq/container-scripts \
     IMAGE_VERSION=${REF}
 
-## Atomic/OpenShift Labels
-LABEL name="manageiq" \
+LABEL name="manageiq-base" \
```
bdunne: Same.  Maybe replace it with "application-base" or something?

images/manageiq-base/Dockerfile
```diff
@@ -61,36 +63,28 @@ RUN yum -y install centos-release-scl-rh && \
                    &&                      \
     yum clean all
 
-## GIT clone manageiq-appliance
 RUN mkdir -p ${APPLIANCE_ROOT} && \
-    curl -L https://github.com/ManageIQ/manageiq-appliance/tarball/${REF} | tar vxz -C ${APPLIANCE_ROOT} --strip 1
+    curl -L https://github.com/${APPLIANCE_ORG}/manageiq-appliance/tarball/${APPLIANCE_REF} | tar vxz -C ${APPLIANCE_ROOT} --strip 1
+
+RUN mkdir -p ${SUI_ROOT} && \
+    curl -L https://github.com/${SUI_ORG}/manageiq-ui-service/tarball/${SUI_REF} | tar vxz -C ${SUI_ROOT} --strip 1
```
bdunne: Is this needed in this layer?

images/manageiq-webserver-worker/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM manageiq/manageiq-base-worker:latest
+MAINTAINER ManageIQ https://manageiq.org
+
+LABEL name="manageiq-webserver-worker" \
```
bdunne: Same

images/manageiq-ui-worker/Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM manageiq/manageiq-webserver-worker:latest
+MAINTAINER ManageIQ https://manageiq.org
+
+LABEL name="manageiq-ui-worker" \
```
bdunne: Same

images/manageiq-orchestrator/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM manageiq/manageiq-base:latest
+MAINTAINER ManageIQ https://manageiq.org
+
+LABEL name="manageiq-orchestrator" \
```
bdunne: Same

images/manageiq-base-worker/Dockerfile
Fryguy: What is the LABEL for anyway...OpenShift?

images/manageiq-base/Dockerfile
carbonin: Maybe not?

I was trying to get all of the source in this image though for ease of development, but if it makes a big difference we could probably move this to the webservice image.

Fryguy: It's actually quite small... 5.1MB currently for master.

images/manageiq-base-worker/Dockerfile
bdunne: We probably don't need it, but it's a way to add metadata to an image.  https://docs.docker.com/engine/reference/builder/#label
https://docs.openshift.com/enterprise/3.0/creating_images/metadata.html

carbonin: OpenShift probably interprets them in some way, but our image metadata isn't really in openshift anymore; not since we removed Image Streams anyway. Maybe we can just remove these?

As far as I can tell the only way to see the labels is to `docker inspect` an image and in this case ours shows this:
```json
            "Labels": {
                "build-date": "20170510",
                "description": "ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures.",
                "io.k8s.description": "ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures.",
                "io.k8s.display-name": "ManageIQ",
                "io.openshift.tags": "ManageIQ,miq,manageiq",
                "license": "GPLv2",
                "name": "manageiq-base-worker",
                "release": "",
                "summary": "ManageIQ Worker Image",
                "url": "https://manageiq.org/",
                "vendor": "ManageIQ",
                "version": "Master"
           }
```

There's probably some things there we want to change, but it hardly seems necessary to include all of that.

PR #210, opened 2017-09-01
Adding external authentication httpd configuration files

abellotti:
- Introduced HTTPD_AUTH_CONFIGURATION to select which authentication configuration files to include
- authentication.conf includes the configuration-* file based on HTTPD_AUTH_CONFIGURATION
- common configurations sections are shared by different types minimizing duplication.
- Introduced HTTPD_AUTH_KERBEROS_REALMS needed for kerberos authentication

### DIFF
--- a/README.md
+++ b/README.md
@@ -575,7 +575,7 @@ The config map includes the following:
 
 * The httpd configuration type `auth-configuration`, default is `internal`
 
-	This parameter drive which configuration files httpd will load upon start-up. Supported values are:
+	This parameter drives which configuration files httpd will load upon start-up. Supported values are:
 
 
 	| Value    | External-Authentication Configuration |
@@ -585,9 +585,11 @@ The config map includes the following:
 	| active-directory | Active Directory domain realm join
 	| saml | SAML based authentication (Keycloak, etc.)
 
-* The kerberos realms joined `auth-kerberos-realms`, default is `undefined`
+* The kerberos realms to join `auth-kerberos-realms`, default is `undefined`
 
 	When configuring external authentication against IPA, Active Directory or Ldap, this parameter defines the kerberos realm httpd is configured against, i.e. `example.com`
+
+	When specifying multiple Kerberos realms, they need to be space separated.
 
 * The external authentication configuration file `auth-configuration.conf` which declares the list of files to overlay upon startup if `auth-type` is other than `internal`.
 
@@ -622,6 +624,8 @@ _Examples_:
 * 644:root:apache
 
 Binary files can be specified in the configuration map in their base64 encoded format with a basename having a `.base64` extension. Such files are then converted back to binary as they are copied to their target path.
+
+When an /etc/sssd/sssd.conf file is included in the configuration map, the httpd pod automatically enables the sssd service upon startup.
 
 ### Auth-type and auth-configuration specification matrix in a configmap:
 
--- a/README.md
+++ b/README.md
@@ -623,6 +623,19 @@ _Examples_:
 
 Binary files can be specified in the configuration map in their base64 encoded format with a basename having a `.base64` extension. Such files are then converted back to binary as they are copied to their target path.
 
+### Auth-type and auth-configuration specification matrix in a configmap:
+
+* auth-type depicts the Identity Provider external authentication is being configured against.
+* auth-configuration identifies which Httpd external authentication configuration files to load.
+
+| auth-type | auth-configuration | Note |
+|-----------|--------------------|------|
+| internal  | internal           | Database / ManageIQ Ldap(s) / Amazon    |
+| ldap      | external           |     |
+| ipa       | external           |     |
+| active-directory | external | Configured against AD via SSSD as an Ldap directory |
+| active-directory | active-directory | Configured against AD domain via realm join |
+| saml | saml | Keycloak / ADFS / etc. |
 
 ### Sample external authentication configuration:
 
--- a/README.md
+++ b/README.md
@@ -571,19 +571,14 @@ The config map includes the following:
 
 * The authentication type `auth-type`, default is `internal`
 
-	`internal` is the default type, anything else is considered external. `auth-type` could include strings like: ipa, ldap, active_directory, saml or simply custom.
-
-* The httpd configuration type `auth-configuration`, default is `internal`
-
 	This parameter drives which configuration files httpd will load upon start-up. Supported values are:
-
 
 	| Value    | External-Authentication Configuration |
 	| ---------|---------------------------------------|
-	|internal | Application Based Authentication (_default_) - Database, Ldap/Ldaps, Amazon |
+	| internal | Application Based Authentication (_default_) - Database, Ldap/Ldaps, Amazon |
 	| external | IPA, IPA 2-factor authentication, IPA/AD Trust, Ldap (OpenLdap, RHDS, Active Directory, etc.)
 	| active-directory | Active Directory domain realm join
-	| saml | SAML based authentication (Keycloak, etc.)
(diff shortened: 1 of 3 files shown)


abellotti:
[WIP] Do not merge.

Need to test:
- [ ] Form based login
- [ ] Kerberos SSO
- [ ] API Authentication
- [ ] SAML / Keycloak

Also need to:
- [ ] Updated README.md to talk about the auth-config entries, auth-configuration and auth-kerberos-realms
- [ ] Duplicate to the ext-db yaml.

abellotti:
Kerberos SSO support will require https://github.com/ManageIQ/container-httpd/pull/24

abellotti:
No changes needed here for a while, I'm good with this.

ping @carbonin @bdunne can I borrow your 👀 for review/merge ?  Thanks.

README.md
```diff
@@ -638,6 +653,8 @@ Excluding the content of the files, a SAML auth-config map data section may look
 apiVersion: v1
 data:
   auth-type: saml
+  auth-configuration: saml
```
carbonin: What is the distinction between `auth-type` and `auth-configuration` in what situations will they not be the same value?

Maybe it would be worth making the example one of those cases to avoid confusion?

templates/miq-template-ext-db.yaml
```diff
@@ -433,18 +433,167 @@ objects:
...
         KeepAlive on
+        # Without ServerName mod_auth_mellon compares against http:// and not https:// from the IdP
+        ServerName https://%{REQUEST_HOST}
+
         ProxyPreserveHost on
         ProxyPass        /ws/ ws://${NAME}/ws/
         ProxyPassReverse /ws/ ws://${NAME}/ws/
-        ProxyPass        / http://${NAME}/
+
+        # For httpd, some ErrorDocuments must by served by the front-end httpd
```
carbonin: I think front-end is overloaded here. I think "httpd pod" would be better to distinguish it from the miq-frontend pod

templates/miq-template-ext-db.yaml
```diff
@@ -433,18 +433,167 @@ objects:
...
+        RewriteCond %{REQUEST_URI} !^/proxy_pages
+
+        # For SAML /saml2 is only served by front-end httpd by mod_auth_mellon
```
carbonin: Same as above, we should avoid using "front-end" to reference the auth httpd pod

README.md
abellotti: 👍 I've added an auth-type / auth-configuration specification matrix table.

README.md
```diff
@@ -573,6 +573,22 @@ The config map includes the following:
 
 	`internal` is the default type, anything else is considered external. `auth-type` could include strings like: ipa, ldap, active_directory, saml or simply custom.
 
+* The httpd configuration type `auth-configuration`, default is `internal`
+
+	This parameter drive which configuration files httpd will load upon start-up. Supported values are:
```
carbonin: Typo

"This parameter drive" -> "This parameter drives"

README.md
```diff
@@ -573,6 +573,22 @@ The config map includes the following:
...
+
+
+	| Value    | External-Authentication Configuration |
+	| ---------|---------------------------------------|
+	|internal | Application Based Authentication (_default_) - Database, Ldap/Ldaps, Amazon |
+	| external | IPA, IPA 2-factor authentication, IPA/AD Trust, Ldap (OpenLdap, RHDS, Active Directory, etc.)
+	| active-directory | Active Directory domain realm join
+	| saml | SAML based authentication (Keycloak, etc.)
+
+* The kerberos realms joined `auth-kerberos-realms`, default is `undefined`
```
carbonin: "The kerberos realms joined" seems strangely worded to me. Maybe "The kerberos realm to join"?

Also if this is indeed plural, how do you specify multiple? Comma separated?

README.md
```diff
@@ -607,7 +623,19 @@ _Examples_:
 
 Binary files can be specified in the configuration map in their base64 encoded format with a basename having a `.base64` extension. Such files are then converted back to binary as they are copied to their target path.
 
-When an /etc/sssd/sssd.conf file is included in the configuration map, the httpd pod automatically enables the sssd service upon startup.
```
carbonin: Is this no longer true? Why remove this line? I feel like the table doesn't make this point clear.

README.md
```diff
@@ -607,7 +623,19 @@ _Examples_:
...
+### Auth-type and auth-configuration specification matrix in a configmap:
+
+* auth-type depicts the Identity Provider external authentication is being configured against.
+* auth-configuration identifies which Httpd external authentication configuration files to load.
+
+| auth-type | auth-configuration | Note |
+|-----------|--------------------|------|
+| internal  | internal           | Database / ManageIQ Ldap(s) / Amazon    |
+| ldap      | external           |     |
```
carbonin: What happens if someone configures a nonsense combination of these parameters (e.g. ldap + active-directory or saml + external)? It seems like the only one with multiple combinations is `active-directory` so can we assume the `auth-configuration` value for the others?

Or even better, can we make `auth-configuration` specific to `active-directory`? Maybe something like `active-directory-config-set` or do we intend to add more `auth-configuration` options for other `auth-type`s in the future?

README.md
carbonin: Thanks, the table is really helpful. It allows me to ask better questions :wink: haha

README.md
abellotti: The auth-configuration is directly drives what Apache configurations we support.  they cannot specify anything else. As we add support for others in the future, i.e. openid, we'd be adding Apache config/include files to reflect the new auth-configuration. So the auth-configuration set is strict/concrete.  The auth-type identifies the identify provider we're targeting, note how multiple types share the same auth-configuration apache include files.  As far as the values there, if we want/need to to differentiate via the auth-config CLI a generated config map for Keycloak vs. let's say Active Directory Federated Services, we can have auth-type (saml | adfs | okta | etc.), but the auth-configuration must be saml.

Hopefully this help a bit.

carbonin: Okay, that's fine. So we can't make auth-configuration more specific to AD (even though right now that is the only auth-type that could possibly have more than one valid auth-configuration). But my first question still stands. Are we doing anything to validate against nonsense combinations of auth-type + auth-configuration?

carbonin: After discussion, it seems like the values in the left column do not have any effect on the runtime of the application.

The values in the right column can be used for the same purpose as the auth-type is currently being used for (deciding whether to copy config files listed in `auth-configuration.conf`) as well as which specific httpd auth configuration files we should include in the `authentication.conf` file, so we can remove the values on the left side (but can keep using the existing auth-type name for the field).

abellotti:
@carbonin simplified and updated.  also updated the auth-config CLI and tested end-to-end for IPA.  still good.  Thanks!!

README.md
```diff
@@ -571,7 +571,20 @@ The config map includes the following:
...
+	| internal | Application Based Authentication (_default_) - Database, Ldap/Ldaps, Amazon |
+	| external | IPA, IPA 2-factor authentication, IPA/AD Trust, Ldap (OpenLdap, RHDS, Active Directory, etc.)
+	| active-directory | Active Directory domain realm join
+	| saml | SAML based authentication (Keycloak, ADFS, etc.)
+
+* The kerberos realms to join `auth-kerberos-realms`, default is `undefined`
+
+	When configuring external authentication against IPA, Active Directory or Ldap, this parameter defines the kerberos realm httpd is configured against, i.e. `example.com`
+
+	When specifying multiple Kerberos realms, they need to be space separated.
```
carbonin: :+1:

PR #214, opened 2017-09-08
Cleanup unnecessary container-fix-permissions script from docker assets

fbladilo:
@bdunne Cleanup old assets, as a bonus should kick a new build 😄


carbonin:
Nice find @fbladilo! What was this originally used for? We were running privileged before anyuid, so I can't think of why we would have needed this.

carbonin:
Beat me to it @bdunne :laughing:

fbladilo:
@carbonin Was thinking hard about it this morning, I think it comes from old era futile attempts to run systemd user in restricted mode w/ arbitrary user  😄

Issue #211, opened 2017-09-04
Random failures in manageiq-0 pod after some time (15-30 minutes)


ilackarms:
I am noticing that the `manageiq-0` pod readiness check goes from `1/1` to `0/1` in a seemingly unpredictable pattern. It usually stays alive for about 20-30 minutes after I bring it up on the first try, then the readiness check fails.

Upon `oc rsh` into the pod, i can tell that it is `httpd` that is failing, not `manageiq`. `curl localhost:3000` replies with the login page for manageiq. `curl localhost:80` hangs indefinitely.

inspecting `log/apache/miq_apache.log`: http://termbin.com/nbob

ilackarms:
i noticed that the manageiq internal httpd (and the pod readiness check) came back up after about 60 minutes of downtime (on its own).

carbonin:
I wonder if this is caused by us using an HTTP GET for our readiness, check. The most interesting line to me in that log is definitely `[Mon Sep 04 13:28:12.242785 2017] [mpm_prefork:error] [pid 210] AH00161: server reached MaxRequestWorkers setting, consider raising the MaxRequestWorkers setting`

Can you see if that line corresponds with when the readiness checks start failing and also can you see if removing the readiness checks prevents us from failing like this?

carbonin:
Maybe we can consider using some other health checks if this is the case, maybe the TCP check can be the readiness check and we can check a pid file for liveness?

fbladilo:
@ilackarms See #218 , let me know if this helped your case.

carbonin:
I don't think this applies to the new architecture. Closing.

PR #218, opened 2017-09-13
Enhance HTTPD pod liveness/readiness probes

fbladilo:
- The httpd pod is considered alive if a httpd process is spawned
- The httpd pod is considered ready if a socket can be open on port 80


fbladilo:
@bdunne @carbonin Please have a look, seems like a more reasonable strategy for probes. Related to #211

Issue #213, opened 2017-09-07
Rename miq-sysadmin service account


carbonin:
When running on a full OpenShift cluster the miq-sysadmin scc is not needed for the httpd auth container. This is because the oci-systemd hooks take care of the necessary capabilities needed for systemd to run in the container.

These hooks do not run on minishift.

I think it would be better for the service account to be called miq-httpd; then we can document what scc should be added to the service account based on the environment you are using.

cc @Fryguy @fbladilo @bdunne

fbladilo:
@carbonin 👍 I agree, is only conditionally needed for minishift. BTW, looks like the minishift OCI systemd hooks issue that we opened is getting some traction : 

https://github.com/minishift/minishift/issues/1299

PR #220, opened 2017-09-19
Use httpd service account for all environments

carbonin:
This allows us to use the same template on minishift as well as a full OpenShift cluster. Now the change that needs to be made based on environment is just in documentation steps rather than editing the template.

Technically, this always uses the miq-httpd service account for the httpd pod and allows the user to choose the scc assigned to that service account based on their environment. For minishift that scc will be miq-sysadmin and for the full cluster, it will be anyuid.

@fbladilo This is more what I was thinking would be a good fix for #213 
Now devs using minishift don't have to change the template (and deal with a dirty git status) to run on minishift.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,9 @@ $ oc describe scc privileged | grep Users
 Users:					system:serviceaccount:<your-namespace>:miq-privileged
 ```
 
-### Minishift only development environment considerations
+### Set up the miq-httpd service account
+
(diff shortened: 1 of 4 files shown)


fbladilo:
@carbonin On another thought, what if we parametize this thingy? We can instruct minishift users to -p "HTTPD_SA=miq-sysadmin" when deploying.

JPrause:
@bdunne can you set the Milestone to Sprint 70

PR #221, opened 2017-09-19
Rename API redirects httpd conf file

carbonin:
This goes along with the changes in https://github.com/ManageIQ/manageiq-appliance/pull/131

This should be merged after that patch so that the file actually exists.

### DIFF
--- a/images/miq-app-frontend/docker-assets/manageiq-http.conf
+++ b/images/miq-app-frontend/docker-assets/manageiq-http.conf
@@ -12,7 +12,7 @@ Options SymLinksIfOwnerMatch
 <VirtualHost *:80>
   DocumentRoot /var/www/miq/vmdb/public
   Include conf.d/manageiq-redirects-ui
-  Include conf.d/manageiq-redirects-ws
+  Include conf.d/manageiq-redirects-api
   Include conf.d/manageiq-redirects-websocket
   ProxyPreserveHost on
   <Location /assets/>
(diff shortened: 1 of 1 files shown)


images/miq-app-frontend/docker-assets/manageiq-http.conf
```diff
@@ -12,7 +12,7 @@ Options SymLinksIfOwnerMatch
 <VirtualHost *:80>
   DocumentRoot /var/www/miq/vmdb/public
   Include conf.d/manageiq-redirects-ui
-  Include conf.d/manageiq-redirects-ws
+  Include conf.d/manageiq-redirects-api
```
Fryguy: In the other repo, I think there was an ordering issue that forced me to put api before ui, so that should be done here as well.

carbonin: I imagine it wouldn't matter. This is working now and you didn't change the ordering in the PR where you made the name change right?

Fryguy: I ensured the ordering was api, then ui, then websocket because that's what it was.  I remember @bdunne saying order was important (though I don't know why).

carbonin:
The manageiq-appliance PR is merged ... Container builds will now be broken until this one is also merged.

bump @bdunne @Fryguy

carbonin:
:tada: Thanks @bdunne

PR #238, opened 2017-10-25
Update external DB section with PG version requirements

fbladilo:
- We require a minimum of PG 9.4 or 9.5 if replication is needed


fbladilo:
@carbonin Forgot to send this one in 🙂

PR #246, opened 2017-11-15
Add reference to the httpd configmap generator

abellotti:
- Add reference to the httpd configmap generator in the external authentication configuration section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -679,3 +679,5 @@ $ oc replace configmaps httpd-auth-configs --filename external-auth-configmap.ya
 ```
 
 Then redeploy the httpd pod for the new authentication configuration to take effect.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -679,3 +679,5 @@ $ oc replace configmaps httpd-auth-configs --filename external-auth-configmap.ya
 ```
 
 Then redeploy the httpd pod for the new authentication configuration to take effect.
+
+Support for automatically generating authentication configuration maps for the _httpd_ pod is provided by [ManageIQ/httpd\_configmap\_generator](https://github.com/ManageIQ/httpd_configmap_generator). Please see the [README.md](https://github.com/ManageIQ/httpd_configmap_generator/blob/master/README.md) in that repo for further details.
```
carbonin: Why is httpd in italics here? Is isn't anywhere else right?

abellotti: good point, I'll update.

Issue #237, opened 2017-10-25
setting MIQ admin password during installation / template processing


joelddiaz:
Currently when using the OpenShift templates to install MIQ, there is no way to set the MIQ admin password. The MIQ pod comes up with a publicly accessible OpenShift route with the default creds.

Need a way to set the admin MIQ password before the pod is publicly accessible.

Opened a similar issue out on openshift/openshift-ansible#5849 and it was recommended I open this issue here.

moolitayer:
@gtanzillo this seems like a major flaw? I'm thinking the solution should be limited to production env? 

@joelddiaz What's the usual approach for something like this? Generating a random pass and storing it in a file an admin can access? providing an optional password during installation (`oc new-app --template=manageiq -p PASSWORD=<password>`)?

@miq-bot assign @gtanzillo

gtanzillo:
@fbladilo Can you take a  shot at addressing this issue. We were think ing that we could parameterize the admin password in the template.

joelddiaz:
@moolitayer i would prefer a system where i can pass the desired password and have it applied during installation

fbladilo:
@gtanzillo For sure, I can take it , will post something later today.

carbonin:
Closing as https://github.com/ManageIQ/manageiq-pods/pull/250 is merged

PR #250, opened 2017-11-30
Allow setting MIQ admin password during deployment

fbladilo:
* Defines new APPLICATION_ADMIN_PASSWORD param
* Adds necessary element to existing manageiq secrets object
* Adds shell function calling rails runner, seeds minimal models and changes password
* Function is called via appliance-initialize.sh, last step for new_deployment case
* By default password is set to default MIQ value but customizable as desired at deployment time
* Related to #237 

Usage : `oc new-app --template=manageiq -p APPLICATION_ADMIN_PASSWORD=mypasswd`

@carbonin @bdunne

### DIFF
--- a/images/miq-app/docker-assets/appliance-initialize.sh
+++ b/images/miq-app/docker-assets/appliance-initialize.sh
@@ -35,7 +35,6 @@ case $? in
     # Restore symlinks from PV to application rootdir
     restore_pv_data
 
-    # Set admin pwd
     set_admin_pwd
   ;;
   4) # new_replica
--- a/images/miq-app/docker-assets/appliance-initialize.sh
+++ b/images/miq-app/docker-assets/appliance-initialize.sh
@@ -34,6 +34,9 @@ case $? in
 
     # Restore symlinks from PV to application rootdir
     restore_pv_data
+
+    # Set admin pwd
+    set_admin_pwd
   ;;
   4) # new_replica
     echo "== Starting New Replica =="
--- a/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
+++ b/images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
@@ -120,6 +120,15 @@ function migrate_db() {
   ) 2>&1 | tee ${PV_MIGRATE_DB_LOG}
 }
 
+# Set EVM admin pwd
+function set_admin_pwd() {
+ echo "== Setting admin password =="
+
+   cd ${APP_ROOT} && bin/rails runner -e production "EvmDatabase.seed_primordial; user = User.find_by_userid('admin'); user.password = ENV['APPLICATION_ADMIN_PASSWORD']; user.save; exit;"
+
+   [ "$?" -ne "0" ] && echo "ERROR: Failed to set admin password, please check appliance logs"
+}
+
 # Process DATA_PERSIST_FILE which contains the desired files/dirs to store on the PV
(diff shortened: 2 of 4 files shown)


templates/miq-template.yaml
```diff
@@ -888,6 +894,12 @@ parameters:
   displayName: Application Database Region
   description: Database region that will be used for application.
   value: '0'
+- name: APPLICATION_ADMIN_PASSWORD
+  displayName: Application Admin Password
+  required: true
+  description: Admin password that will be set on the application.
+  from: "[a-zA-Z0-9]{8}"
+  generate: expression
```
carbonin: I feel like we should keep the default value here rather than using a generator.

If someone wants to change the password on deploy they can set the parameter, otherwise I think we should just keep "smartvm"

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
```diff
@@ -120,6 +120,15 @@ function migrate_db() {
   ) 2>&1 | tee ${PV_MIGRATE_DB_LOG}
 }
 
+# Set EVM admin pwd
+function set_admin_pwd() {
+ echo "== Setting admin password =="
+
+   cd ${APP_ROOT} && bin/rails runner -e production "EvmDatabase.seed_primordial; user = User.find_by_userid('admin'); user.password = ENV['APPLICATION_ADMIN_PASSWORD']; user.save; exit;"
```
carbonin: I think we can simplify the ruby here.

```ruby
EvmDatabase.seed_primordial; User.find_by_userid('admin').update_attributes!(:password => ENV['APPLICATION_ADMIN_PASSWORD'])
```

@bdunne this look good to you?

carbonin: Also, I think `RAILS_ENV=production` is exported in `/etc/default/evm` so we shouldn't need the `-e production` here.

chessbyte:
@fbladilo will you have time to respond to the comments in this PR?

images/miq-app/docker-assets/appliance-initialize.sh
```diff
@@ -34,6 +34,9 @@ case $? in
 
     # Restore symlinks from PV to application rootdir
     restore_pv_data
+
+    # Set admin pwd
```
carbonin: This comment hardly seems necessary.

images/miq-app/docker-assets/container-scripts/container-deploy-common.sh
fbladilo: @carbonin Yep , will go ahead and test with the simplied version. `/etc/default/evm` is  already sourced so should work without -e.

images/miq-app/docker-assets/appliance-initialize.sh
fbladilo: @chessbyte Surely , already commented and working on requested changes..

simaishi:
Gaprindashvili backport details:

```
$ git log -1
commit 8786454e167e9833c40ca6acc446b27af0242cd1
Date:   Thu Dec 7 09:42:57 2017 -0500

    Merge pull request #250 from fbladilo/allow_set_app_admin_pwd
    
    Allow setting MIQ admin password during deployment
    (cherry picked from commit 941a134ca488427e4830933b53af3a60f69dd806)
```

PR #252, opened 2017-12-11
Allow changing the GitHub organization when building the images

elad661:
This makes it easier to create custom builds of ManageIQ, by telling the image build to pull the sources from a fork instead of pulling them from upstream.

Our specific usecase is for integration testing of pending ManageIQ pull requests. For a while now we've had a fork of this repo with a custom dockerfile, but that got out-of-sync with upstream rather quickly. Having this as a build argument will allow us to stay in sync with upstream when we do our custom builds.

cc: @bazulay

### DIFF
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -3,6 +3,7 @@ MAINTAINER ManageIQ https://github.com/ManageIQ/manageiq-appliance-build
 
 ## Set build ARGs
 ARG REF=master
+ARG GHORG=ManageIQ
 
 ## Set ENV, LANG only needed if building with docker-1.8
 ENV TERM=xterm \
@@ -68,7 +69,7 @@ RUN mkdir -p ${APPLIANCE_ROOT} && \
 ## GIT clone manageiq
 RUN mkdir -p ${APP_ROOT} && \
     ln -vs ${APP_ROOT} /opt/manageiq/manageiq && \
-    curl -L https://github.com/ManageIQ/manageiq/tarball/${REF} | tar vxz -C ${APP_ROOT} --strip 1
+    curl -L https://github.com/${GHORG}/manageiq/tarball/${REF} | tar vxz -C ${APP_ROOT} --strip 1
 
 ## Setup environment
 RUN ${APPLIANCE_ROOT}/setup && \
--- a/images/miq-app-frontend/Dockerfile
+++ b/images/miq-app-frontend/Dockerfile
@@ -3,6 +3,7 @@ MAINTAINER ManageIQ https://github.com/ManageIQ/manageiq-appliance-build
 
 ## Set build ARGs
 ARG REF=master
+ARG GHORG=ManageIQ
 
 ## Set ENV, LANG only needed if building with docker-1.8
 ENV SUI_ROOT=/opt/manageiq/manageiq-ui-service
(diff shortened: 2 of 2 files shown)


bazulay:
@fbladilo @carbonin @gtanzillo  PTAL
This is needed for us to build & use appliance pod  with pre merged patches for the ops effort.
We use container-mgmt org at docker hub/io

carbonin:
I think we should probably make the argument specific to the repo it will be used for.

For example, this PR does not make the org configurable for the `manageiq-appliance` repo, but I think that's within the scope here.

I also don't want to require all of the repos to be present and updated in an org if you are only testing a change for one.

I think simply changing the argument name to something like `MANAGEIQ_ORG`, `APPLIANCE_ORG`, `UI_ORG` would do it. I'm all for naming suggestions though as I'm not sold on these exactly.

Thoughts @bdunne ?

carbonin:
> We're building `manageiq-pods`, not `manageiq-appliance`.

Right, but we clone the `manageiq-appliance` github repo during the pods build process. My point was that the org on [this line](https://github.com/ManageIQ/manageiq-pods/blob/master/images/miq-app/Dockerfile#L66) should also be configurable.

elad661:
Maybe it'd be better to have each of the clone URLs as variables, instead of just the org names? That way people could use local git repos, and then the arguments could be something like

```
MANAGEIQ=https://github.com/ManageIQ/manageiq/tarball/${REF} 
MANAGEIQ_APPLIANCE=https://github.com/ManageIQ/manageiq-appliance/tarball/${REF} 
```
and so on.

This feels "cleaner" than using *_ORG variables (because it's more obvious what each variable controls), but I'm not sure how it'd behave with ${REF} there

carbonin:
@elad661 I think that would cause the kind of issue I was trying to avoid. You would need a branch with the same name for both repos.

elad661:
@carbonin I'm not sure I understand.

Assuming you've changed manageiq/manageiq in your fork and called the branch "foo", you'd set `MANAGEIQ=http://github.com/<username>/manageiq/tarball/foo`
and you won't touch `MANAGEIQ_APPLIANCE` - so it'll remain the default value with the default `${REF}`. This means you only needed to fork (and change) one repo in this case. You could also set different branches for different repos by simply changing the URLs to whatever you want.

carbonin:
Ah I see, you're overriding the whole URL in that case...

Still seems a bit strange to have two levels of configuration (just change REF in the case of doing a release build vs. changing the whole URL for dev).

images/miq-app/Dockerfile
```diff
@@ -3,6 +3,7 @@ MAINTAINER ManageIQ https://github.com/ManageIQ/manageiq-appliance-build
 
 ## Set build ARGs
 ARG REF=master
+ARG GHORG=ManageIQ
```
carbonin: I think this will break the clone for `manageiq-appliance` if you provide a REF that doesn't exist on both your fork of `manageiq` and the ManageIQ fork of `manageiq-appliance`.

bdunne:
I agree that the `GHORG` arg is not enough, and `REF` is also very weak, but I don't have a better solution than spelling out the fully qualified path for each resource.  I guess it hasn't affected me enough to justify making a change.

carbonin:
So, we're working on merging in the "rearch" changes and I applied some changes that I think will solve this [here](https://github.com/ManageIQ/manageiq-pods/pull/207/commits/007c0fb3b39b948a574985d6dd52cb6389f385c6). Will that do the job for you?

elad661:
It looks like it'll work, and that's good enough for me.

PR #256, opened 2017-12-21
Remove the embedded ansible objects from the template

carbonin:
These objects will be created by the server process as they are
needed.

This way we can control what images are used and how they are deployed
from the ManageIQ code base rather than maintaining templates.

Additionally, users who are not using the embedded ansible role
will no-longer have an unused deployment config sitting in the project


carbonin:
Requires https://github.com/ManageIQ/manageiq/pull/15963

PR #269, opened 2018-02-19
Fix the external db template after merging #207

carbonin:
Missed these in #207

### DIFF
--- a/templates/miq-template-ext-db.yaml
+++ b/templates/miq-template-ext-db.yaml
@@ -36,22 +36,6 @@ objects:
     v2-key: "${V2_KEY}"
     artemis-password: "${ARTEMIS_PASSWORD}"
 - apiVersion: v1
-  kind: Service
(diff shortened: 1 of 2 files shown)


bdunne:
Can you sync up the two templates?  There are other things that are not the same.  It's hard to compare in a diff viewer

carbonin:
Yeah, noticed that too @bdunne. Wasn't sure if I should do that as a separate PR or a separate commit in this one.

bdunne:
Yeah, multiple commits in one PR would probably be best.

carbonin:
Expanded the scope of this PR to include other miscellaneous fixes to the templates caused by the previous changes.

Now you should be able to `diff` the two templates and only see database-related objects

carbonin:
Also because I reordered the object list reviewing commit-by-commit is strongly recommended :smile:

PR #275, opened 2018-03-15
Remove 'docker' references

simaishi:


images/manageiq-base-worker/Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER ManageIQ https://manageiq.org
 LABEL name="manageiq-base-worker" \
       summary="ManageIQ Worker Image"
 
-COPY docker-assets/entrypoint /usr/local/bin
-COPY docker-assets/manageiq_liveness_check /usr/local/bin
+COPY assets/entrypoint /usr/local/bin
```
Fryguy: container-assets?

simaishi: Since there is no other assets, I just shortened it... I can change to container-assets if that's preferred.

Fryguy: Not sure... @bdunne ?

bdunne: Yeah, I prefer `container-assets`

templates/miq-template-ext-db.yaml
```diff
@@ -55,7 +55,7 @@ objects:
         ProxyPass /api http://web-service:3000/api
         ProxyPassReverse /api http://web-service:3000/api
 
-        # Ensures httpd stdout/stderr are seen by docker logs.
```
bdunne: In this context `docker logs` is a command.

templates/miq-template.yaml
```diff
@@ -128,7 +128,7 @@ objects:
         ProxyPass /api http://web-service:3000/api
         ProxyPassReverse /api http://web-service:3000/api
 
-        # Ensures httpd stdout/stderr are seen by docker logs.
```
bdunne: Same here

templates/miq-template-ext-db.yaml
simaishi: Didn't realize that's a command... I'll put a quote around it then.

images/manageiq-base-worker/Dockerfile
simaishi: Ok, will change.

simaishi:
@bdunne Addressed all feedback, please review.

PR #278, opened 2018-03-27
[GAPRINDASHVILI] Fix container logrotate

carbonin:
This PR adds the logrotate configuration to the app images and alters it so that it will correctly rotate the application logs. In the first commit, the initial file was taken from the [manageiq-appliance repo](https://github.com/ManageIQ/manageiq-appliance/blob/a04b2a8a2a96f7a801e4701cd9c8cd2492079d69/COPY/etc/logrotate.d/miq_logs.conf). Subsequent commits alter the file.

Before this PR logs were not being rotated because the free space check script was failing on the container filesystem. Because filesystem space is not really the concern of the container, the check was removed from this overwrite file.

https://bugzilla.redhat.com/show_bug.cgi?id=1560210

### DIFF
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -104,6 +104,7 @@ COPY container-assets/entrypoint /usr/bin
 COPY container-assets/container.data.persist /
 COPY container-assets/appliance-initialize.sh /bin
 COPY container-assets/check-dependent-services.sh /bin
+COPY container-assets/miq_logs.conf /etc/logrotate.d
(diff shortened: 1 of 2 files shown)


carbonin:
Also we removed crond from the master appliances, so this is a gaprindashvili only PR.

@simaishi who gets to merge this one?

simaishi:
> Also we removed crond from the master pods, so this is a gaprindashvili only PR.

This means log rotation works in master, or it needs a different fix?

simaishi:
Talked to Nick, logs will be handled differently on master.

PR #182, opened 2017-07-10
Add some simple testing

bdunne:
Add testing and clean up YAML syntax

PR #281, opened 2018-04-23
Add .yamllint from provider generator

jrafanie:
Add .yamllint from provider generator

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,13 @@
+---
+ignore: |
+  /vendor/**
+  /spec/manageiq/**
+  /spec/vcr_cassettes/**
+
+extends: relaxed
+
+rules:
+  indentation:
+    indent-sequences: false
+  line-length:
+    max: 120


.yamllint
```diff
@@ -0,0 +1,13 @@
...
+  /spec/manageiq/**
+  /spec/vcr_cassettes/**
+
+extends: relaxed
+
+rules:
+  indentation:
+    indent-sequences: false
+  line-length:
+    max: 120
```
carbonin: Can we remove this restriction?

carbonin:
The CI here does a YAML load and YAML dump and then compares the file before and after. So we already have a kind of linting that will fail the PR tests.

Just want to be sure we can't get into a situation where the bot and specs tell a PR author conflicting things.

jrafanie:
> The CI here does a YAML load and YAML dump and then compares the file before and after. So we already have a kind of linting that will fail the PR tests.
> 
> Just want to be sure we can't get into a situation where the bot and specs tell a PR author conflicting things.

Sure, we can skip this PR.  It sounds like yaml linting and the existing test added in https://github.com/ManageIQ/manageiq-pods/pull/182 are different things.  I'm not sure what the test in #182 is protecting in terms of a regression.  Perhaps a psych version syntax change?  cc @bdunne 

Either way, we can close this PR if it's not needed.

PR #291, opened 2018-05-24
[HAMMER] Backport enhancements made to master since Gaprindashvili

carbonin:
Trying to put together the changes we need to make a hammer branch.

We won't merge this, but it will be easy to see like this before we actually go and create the branch.


images/miq-app-frontend/Dockerfile
```diff
@@ -1,16 +1,16 @@
-FROM manageiq/manageiq-pods:backend-latest-gaprindashvili
+FROM carbonin/manageiq-pods:backend-latest
```
bdunne: This needs to be changed

PR #292, opened 2018-05-31
[HAMMER] Add the artemis username and password to the frontend pod

carbonin:
This was missed in #291


templates/miq-template-ext-db.yaml
```diff
@@ -126,6 +126,13 @@ objects:
               secretKeyRef:
                 name: "${NAME}-secrets"
                 key: admin-password
+          - name: ARTEMIS_USER
```
bdunne: Can these be alphabetical?

PR #295, opened 2018-06-05
Add upgrade documentation

carbonin:


upgrading.md
```diff
@@ -0,0 +1,53 @@
...
+If the Embedded Ansible feature is or was in use the passwords in the ansible secret need to be moved into the application for the new version.
+Fetch the values using the command `oc export secret ansible-secrets`. There should be 3 Base64 encoded values in the secret. Substitute the values for the placeholders in angle brackets (<>) in the commands below.
+
+Apply the values to the application by accessing a manageiq pod (`oc rsh manageiq-0`, for example) and running the following commands in the rails console (`bash; vmdb; bin/rails c`)
+
+```ruby
+> db = MiqDatabase.first
+> db.ansible_secret_key = Base64.decode64("<secret-key>")
+> db.set_ansible_rabbitmq_authentication(userid: "ansible", password: Base64.decode64("<rabbit-password>"))
+> db.set_ansible_database_authentication(userid: ApplicationRecord.configurations["production"]["username"], password: ApplicationRecord.configurations["production"]["password"])
```
bdunne: This feels like it should be in a migration

upgrading.md
```diff
@@ -0,0 +1,53 @@
...
+Note: If you have *never* enabled the Embedded Ansible role, then you can skip these steps and the secrets will be created for you if you choose to use the role in the future.
+
+If the Embedded Ansible feature is or was in use the passwords in the ansible secret need to be moved into the application for the new version.
+Fetch the values using the command `oc export secret ansible-secrets`. There should be 3 Base64 encoded values in the secret. Substitute the values for the placeholders in angle brackets (<>) in the commands below.
+
+Apply the values to the application by accessing a manageiq pod (`oc rsh manageiq-0`, for example) and running the following commands in the rails console (`bash; vmdb; bin/rails c`)
+
+```ruby
+> db = MiqDatabase.first
+> db.ansible_secret_key = Base64.decode64("<secret-key>")
```
bdunne: Can this live in a data migration?

templates/miq-template.yaml
```diff
@@ -357,7 +357,7 @@ objects:
       description: Defines how to deploy the ManageIQ appliance
   spec:
     serviceName: "${NAME}"
-    replicas: "${APPLICATION_REPLICA_COUNT}"
+    replicas: "${{APPLICATION_REPLICA_COUNT}}"
```
bdunne: Can these be two separate PR's?

bdunne:
I'm also thinking it would be nice for any necessary documentation to live in the manageiq-docs repo

templates/miq-template.yaml
carbonin: This seems to have been fine outside of the context of upgrade so I think it really only applies when talking about upgrading, but I can extract it if you feel strongly about it.

upgrading.md
carbonin: That's a really good question. We would have to utilize `KubeClient` to talk to the OpenShift API during the migration to fetch the values, but if that's okay it might be possible (and definitely would be better). I'll investigate that option.

carbonin:
Marking as WIP until https://github.com/ManageIQ/manageiq-schema/pull/218 is merged.

carbonin:
@simaishi @bdunne can we merge this?

PR #301, opened 2018-08-13
[HAMMER] Changed template to add miq-log config map

rh-dluong:
Addresses RFE: https://bugzilla.redhat.com/show_bug.cgi?id=1612814

### DIFF
--- a/templates/miq-template.yaml
+++ b/templates/miq-template.yaml
@@ -55,6 +55,24 @@ objects:
 - apiVersion: v1
   kind: ConfigMap
   metadata:
+    name: miq-logs-config
+  data:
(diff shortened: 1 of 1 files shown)


templates/miq-template.yaml
```diff
@@ -52,6 +52,24 @@ objects:
...
+        missingok
+        rotate 14
+        notifempty
+        compress
+        copytruncate
+        lastaction
+          /sbin/service httpd reload > /dev/null 2>&1 || true
+        endscript
+      }
+  kind: ConfigMap
```
carbonin: Can you move the kind and metadata tags up under `apiVersion` for consistency with the rest of the file?

PR #303, opened 2018-08-31
Add a simple script to deploy MIQ on minishift

agrare:
Adds a setup script to deploy ManageIQ on minishift using a single
command.


agrare:
@miq-bot assign @carbonin

bin/setup
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+project=mynamespace
```
agrare: TODO - take this from an ENV var or command line then default to mynamespace

bin/setup
```diff
@@ -0,0 +1,67 @@
...
+then
+	echo "You must start minishift"
+	exit 1
+fi
+
+# Make sure path to the oc binary is on our PATH
+eval $(minishift oc-env)
+
+# Login as admin
+res=$(oc login -u system:admin)
```
agrare: For some reason bash is trying to execute the stdout of every command unless I set to a variable...not sure if this is just my bash env

carbonin: That's strange ... but I also think `system:admin` was broken in a recent version of minishift. I thought we had to use the admin user addon now. Was that fixed?

bin/setup
```diff
@@ -0,0 +1,67 @@
...
+
+res=$(oc describe scc anyuid | grep Users | awk '{print $2}' | grep miq-anyuid)
```
carbonin: This will return the result for all namespaces. I think this should be `grep "$project:miq-anyuid"`

carbonin: Also is it better to use `grep -q` here?

bin/setup
```diff
@@ -0,0 +1,67 @@
...
+if [ -z "$res" ];
+then
+	echo "Creating SCC miq-anyuid..."
```
carbonin: Technically this is assigning the `anyuid` SCC to the service account called `miq-anyuid`.

bin/setup
carbonin: I think an envvar is a good call for a first pass.

carbonin:
Also I think the script name shouldn't be just `setup` as that is typically not an entrypoint, but rather pre-run tasks. (For example, `bin/setup` in manageiq doesn't start the server).

I also want to make to clear from the name of the script that this is only for minishift. So maybe `deploy_on_minishift` or `minishift_quickstart`?

bin/setup
agrare: :+1: switching to that and checking the return code

bin/setup
agrare: :+1: changed

bin/setup
agrare: :+1: done
Get from env var first then default to myproject

bin/deploy_on_minishift
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
+
+project=$PROJECT
+if [ -z "$project" ];
+then
+	project=myproject
+fi
```
carbonin: Can we use something like this here?

```bash
 project="${PROJECT:-myproject}"
```

agrare: I'm sure we can, bash isn't my first, second, or third language :)

agrare:
@carbonin I assume this will require some tweaks to work on the hammer branch?

carbonin:
I don't think so... All the sccs and service accounts are named the same. Is there anything jumping out at you that won't work in hammer?

agrare:
Okay cool I assumed there was some difference on the setup.

simaishi:
Hammer backport details:

```
$ git log -1
commit 15db993f6b05f504d3e8b56818d0bb60be508de9
Date:   Fri Aug 31 14:14:13 2018 -0400

    Merge pull request #303 from agrare/add_simple_minishift_setup
    
    Add a simple script to deploy MIQ on minishift
    
    (cherry picked from commit d40b35489b7038e361fda855fb255e8ad834830e)
```

PR #307, opened 2018-09-26
[HAMMER] Make builds work

carbonin:
These changes have been made on the appliance side, but have been missing here for months.

Specifically, this PR updates to node 8, creates the `vmdb/log` directory when trying to create the `apache` log directory, and updates to bundler `>=1.16.2`

### DIFF
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -72,7 +72,7 @@ RUN mkdir -p ${APP_ROOT} && \
 
 ## Setup environment
 RUN ${APPLIANCE_ROOT}/setup && \
-    mkdir ${APP_ROOT}/log/apache && \
+    mkdir -p ${APP_ROOT}/log/apache && \
     mkdir ${APP_ROOT_PERSISTENT} && \
     mkdir -p ${CONTAINER_SCRIPTS_ROOT} && \
     cp ${APP_ROOT}/config/cable.yml.sample ${APP_ROOT}/config/cable.yml
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -82,7 +82,7 @@ WORKDIR ${APP_ROOT}
 RUN source /etc/default/evm && \
     export RAILS_USE_MEMORY_STORE="true" && \
     npm install bower yarn -g && \
-    gem install bundler -v 1.15.4 && \
+    gem install bundler -v ">=1.16.2" && \
     bundle install && \
     rake update:ui && \
     bin/rails log:clear tmp:clear && \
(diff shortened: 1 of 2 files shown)


images/miq-app/Dockerfile
```diff
@@ -82,14 +82,14 @@ WORKDIR ${APP_ROOT}
 RUN source /etc/default/evm && \
     export RAILS_USE_MEMORY_STORE="true" && \
     npm install bower yarn -g && \
-    gem install bundler -v 1.15.4 && \
+    gem install bundler -v ">=1.16.2" && \
```
simaishi: Not a big deal.. but where did this `1.16.2` come from? In appliance, we have `gem install bundler -v ">=1.8.4"`

carbonin: Well 1.15 was causing a failure to resolve so I figured it should be updated. @bdunne also mentioned that 1.16.1 had some problem, but I can't remember what it was.

The appliance should probably be updated too.

images/miq-app/Dockerfile
```diff
@@ -48,7 +49,6 @@ RUN yum -y install centos-release-scl-rh && \
                    net-tools               \
                    nmap-ncat               \
                    nodejs                  \
-                   npm                     \
```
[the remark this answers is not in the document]
carbonin: Yes

images/miq-app/Dockerfile
```diff
@@ -30,7 +30,8 @@ RUN curl -sSLko /etc/yum.repos.d/manageiq-ManageIQ-Master-epel-7.repo \
       https://copr.fedorainfracloud.org/coprs/manageiq/ManageIQ-Master/repo/epel-7/manageiq-ManageIQ-Master-epel-7.repo
 
 ## Install EPEL repo, yum necessary packages for the build without docs, clean all caches
-RUN yum -y install centos-release-scl-rh && \
+RUN yum -y install centos-release-scl-rh \
+                   https://rpm.nodesource.com/pub_8.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm && \
```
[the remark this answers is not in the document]
carbonin: Nope, I'm doing the same thing we do [in the appliance](https://github.com/ManageIQ/manageiq-appliance-build/blob/5dbfc1c558244aab0ef0659f1ccaf390145d370f/kickstarts/partials/main/repos.ks.erb#L9)

images/miq-app/Dockerfile
carbonin: Okay, but the ManageIQ appliance is at `1.16.5`

simaishi: It's not a problem to use 1.16 upstream as I don't think we really depends on bundler version for our functionality and this is how it has always been in master (now hammer).  But if we require `1.16.2` to be minimum (in Gemfile), that will break the downstream build...

carbonin: Ah okay, I didn't realize that was a response to @bdunne 's comment rather than the change on this line.

simaishi: oh sorry, that comment wasn't clear...

PR #326, opened 2019-08-02
Update ruby, node and postgresql

simaishi:
The first PR to get master/ivanchuk up-to-date... Updated ruby, node and postgresql packages and a few minor cleanups. 

Trying to make just enough changes to get monolithic (which is built on top of podified) up and running first. So podified will still not work...

### DIFF
--- a/images/miq-app/Dockerfile
+++ b/images/miq-app/Dockerfile
@@ -26,8 +26,7 @@ LABEL name="manageiq" \
       io.openshift.tags="ManageIQ,miq,manageiq"
 
 ## Install EPEL repo, yum necessary packages for the build without docs, clean all caches
-RUN yum -y install centos-release-scl-rh \
-                   https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm && \
+RUN yum -y install https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm && \
     yum -y install --setopt=tsflags=nodocs \
                    chrony                  \
                    cmake                   \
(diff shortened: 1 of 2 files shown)


images/miq-app/Dockerfile
```diff
@@ -53,8 +49,7 @@ RUN yum -y install centos-release-scl-rh \
                    openscap-scanner        \
                    patch                   \
                    psmisc                  \
-                   rh-postgresql95-postgresql-devel  \
-                   rh-postgresql95-postgresql-libs \
+                   postgresql-devel        \
```
carbonin: Is this versioned? Where are we getting this from?

simaishi: Our copr repo: https://copr.fedorainfracloud.org/coprs/manageiq/ManageIQ-Master/packages/

carbonin: Right, just found that ... do we want to rebuild it at some point? We're a few versions behind.

Oh, also we probably need to create an Ivanchuk repo, but this is fine we can merge this.

simaishi: Ivanchuk repo is already created. Didn't know we're already behind on version!!!  Will take a look.

carbonin: Yeah, think we're at 10.6 and 10.9 was released in June.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 96b196732e91d0a44b18b9021c723820e5d959f6
Date:   Fri Aug 2 15:47:03 2019 -0400

    Merge pull request #326 from simaishi/master_ivanchuk_updates
    
    Update ruby, node and postgresql
    
    (cherry picked from commit 4041a80af69ea473826d396e5f8635294a4858b3)
```

PR #334, opened 2019-10-07
OpenShift external DB template enhancements

siw36:
- Added CPU limits to the template because the deployment of the pods will fail when deployed in a NS with configured quotas and limits. If the request is configured request but the limit is left empty, OpenShift will throw an error that the limit can't be lower than the request (happend in OCP 3.11).
- Removed the postgres service and endpoint because they are not needed. You can specify the host name or IP address directly in the postgres connection string. This also enables the use of host names for the DB host, which is not possible for endpoints because they require an IP address in OpenShift.
- Added a note about URL encoding the user name. I've been using Azure DBaaS postgres, which requires the user name to be `<user>@<server_name>` format which will break the connection because of the `@` sign.

Let me know what you think about these changes.

PS: We evaluated this solution and just deployed a production instance last week. Now I wanted to give you guys my feedback about the things I needed to change to get everything running on our OCP cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,9 +247,11 @@ Import the MIQ external db template
 
 `$ oc create -f templates/miq-template-ext-db.yaml`
 
-Launch deployment, database server IP is required, rest of settings must match your remote PG server side.
+Launch deployment, database server IP or host name is required, rest of settings must match your remote PG server side.
 
-`$ oc new-app --template=manageiq-ext-db -p DATABASE_IP=<server_ip> -p DATABASE_USER=<user> -p DATABASE_PASSWORD=<password> -p DATABASE_NAME=<database_name>`
+Note: URL encode the user name, if your user name includes special characters like `@`.
+
+`$ oc new-app --template=manageiq-ext-db -p DATABASE_HOST=<server> -p DATABASE_USER=<user> -p DATABASE_PASSWORD=<password> -p DATABASE_NAME=<database_name>`
 
 ## Verifying the setup was successful
 
--- a/templates/miq-template-ext-db.yaml
+++ b/templates/miq-template-ext-db.yaml
@@ -50,7 +50,7 @@ objects:
   stringData:
     pg-password: "${DATABASE_PASSWORD}"
     admin-password: "${APPLICATION_ADMIN_PASSWORD}"
-    database-url: postgresql://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_SERVICE_NAME}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5
+    database-url: postgresql://${DATABASE_USER}:${DATABASE_PASSWORD}@${DATABASE_HOST}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5
     v2-key: "${V2_KEY}"
 - apiVersion: v1
   kind: Service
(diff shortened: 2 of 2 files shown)


siw36:
Is there something I need to do, to get someone to review this PR?

templates/miq-template-ext-db.yaml
```diff
@@ -305,28 +307,7 @@ objects:
...
-- apiVersion: v1
-  kind: Endpoints
-  metadata:
-    name: "${DATABASE_SERVICE_NAME}"
-  subsets:
-  - addresses:
-    - ip: "${DATABASE_IP}"
-    ports:
-    - port: "${{DATABASE_PORT}}"
-      name: postgresql
```
carbonin: Are we sure this isn't needed? To be honest I can't remember why we needed the service and endpoint combination rather than using the IP directly in the first place.

Maybe we were originally using the service env vars internally rather than passing in the database URL?
@bdunne do you remember?

carbonin:
@siw36 Thanks for the contribution. I'm just coming back to this repo now and will probably be making some larger scope changes in the next few weeks. Just a heads up :smile:

README.md
```diff
@@ -247,9 +247,11 @@ Import the MIQ external db template
 
 `$ oc create -f templates/miq-template-ext-db.yaml`
 
-Launch deployment, database server IP is required, rest of settings must match your remote PG server side.
+Launch deployment, database server IP or host name is required, rest of settings must match your remote PG server side.
 
-`$ oc new-app --template=manageiq-ext-db -p DATABASE_IP=<server_ip> -p DATABASE_USER=<user> -p DATABASE_PASSWORD=<password> -p DATABASE_NAME=<database_name>`
+Note: URL encode the user name, if your user name includes special characters like `@`.
```
carbonin: Good find :+1:

templates/miq-template-ext-db.yaml
siw36: I've deployed our setup without the service and endpoint so I can assure you it's working fine.
Maybe you did it like this because you are using a service in the other templates and just stuck with it for this template?

As I mentioned in the PR message: endpoints can only be created with an IP address. So if someone (like in my case) uses some DBaaS which requires a FQDN as target server rather than an IP, an endpoint can not be used.

carbonin: Actually I'm surprised this line in the startup for the container doesn't fail https://github.com/ManageIQ/manageiq-pods/blob/master/images/miq-app-frontend/container-assets/check-dependent-services.sh#L8

carbonin: If you removed the service called "postgresql" the environment variably `POSTGRESQL_SERVICE_HOST` would no longer be present on the containers in the project.

siw36: Good catch. But the failed check does not prevent the MIQ pod to start normally. There is an error message but the startup proceeds:

```
== Checking <IP>:11211 status ==
<IP>:11211 - accepting connections
Error something seems wrong, we need at least two parameters to check service status
== Writing encryption key ==
== Restoring PV data symlinks =
[...]
```

carbonin:
@siw36 so I just opened up https://github.com/ManageIQ/manageiq-pods/pull/342 which completely removes the external database template, but I think it will solve the issue you were trying to address.

If you're using one of the released branches, you can definitely open this PR against that branch and we can get it in there.

carbonin:
Closing this as the templates have been totally reworked and this has been unmergeable for 5 months.

PR #345, opened 2019-11-06
Add a build script rather than openshift build configs

carbonin:
This is a bit more flexible as it lets us parameterize the
FROM image as well as push to whatever registry we want.

Additionally if we want to try running on kubernetes we won't have
access to build configs.


bin/build
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+TAG=latest
+
+while getopts "t:d:r:nh" opt; do
```
simaishi: Can you add options for REF and ORG?

For example, if REF is specified, use that for MIQ_REF, SUI_REF and APPLIANCE_REF in manageiq-base image.

carbonin: Hm, I'm not sure I'd want to always use the same ref for all of them. When developing it's much more likely that I'll need to change the miq ref, but won't want the same value to apply to the appliance or sui ref. Maybe I'll push those six values into env vars rather than options.

simaishi: I'll take anything as long as REF can be changed 😄

carbonin:
How does this look @simaishi ?

simaishi:
Thanks for adding the options! A few things..

1) It looks like _REF and _ORG variables don't get set. I think they need to be `MIQ_REF=${MIQ_REF:-master}`.  Otherwise all those values will be empty when not set on the system and build fails.

```
 $ bin/build -h
bin/build: line 5: master: command not found
bin/build: line 6: master: command not found
bin/build: line 7: master: command not found
bin/build: line 9: ManageIQ: command not found
bin/build: line 10: ManageIQ: command not found
bin/build: line 11: ManageIQ: command not found
Usage: bin/build -d IMAGE_DIR -r IMAGE_REPOSITORY [-hnp] [ -t IMAGE_TAG ]
```

2) When an image build fails, it just continues to the next image in the list. Can we catch errors?

carbonin:
oops, just pushed a change @simaishi

bin/build
```diff
@@ -0,0 +1,68 @@
...
+      manageiq-base
+  fi
+
+  for image in $images; do
+    docker build -t $REPO/$image:$TAG --build-arg FROM_REPO=$REPO --build-arg FROM_TAG=$TAG $image
+  done
+popd
+
+if [ -n "$PUSH" ]; then
+  for image in images; do
```
simaishi: We need to push `manageiq-base` as well.

carbonin: Do we? Why? We don't run it.

simaishi: But the next image depends on this, so if we don't push the latest, people building dependent images won't have the latest if not building everything using this script.

simaishi:
LGTM, thank you for providing the script!

PR #350, opened 2019-12-04
Print out build args/tag used

simaishi:
Since build args overwritten won't be showing up in docker build log, print out the actual cmd being run.

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -36,34 +36,33 @@ build_images="manageiq-base-worker manageiq-orchestrator manageiq-webserver-work
 set -e
 
 pushd $IMAGE_DIR
+  cmd="docker build --tag $REPO/manageiq-base:$TAG \
+                    --build-arg MIQ_REF=$MIQ_REF \
+                    --build-arg SUI_REF=$SUI_REF\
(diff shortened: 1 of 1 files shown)


bin/build
```diff
@@ -36,34 +36,32 @@ build_images="manageiq-base-worker manageiq-orchestrator manageiq-webserver-work
...
+    cmd="docker build -t $REPO/$image:$TAG --build-arg FROM_REPO=$REPO --build-arg FROM_TAG=$TAG $image"
+    echo "Building $image: $cmd"
+    $cmd
   done
 popd
 
 if [ -n "$PUSH" ]; then
   push_images="manageiq-base manageiq-base-worker manageiq-orchestrator manageiq-webserver-worker manageiq-ui-worker"
   for image in $push_images; do
+    echo "Pushing $image"
```
carbonin: I think it would be helpful to make this `$REPO/$image` at least for the push, if not everywhere else as well.

carbonin: Maybe tag also

simaishi: Done.

PR #354, opened 2019-12-05
Add support to use non-default npm registry

simaishi:
Podified image version of https://github.com/ManageIQ/manageiq-appliance-build/pull/366

Scripts in container-assets/npm_registry were copied from manageiq-appliance-build and I made minor edits to make them work in this repo (e.g. no erb).

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -11,7 +11,6 @@ SUI_ORG=${SUI_ORG:-ManageIQ}
 APPLIANCE_ORG=${APPLIANCE_ORG:-ManageIQ}
 
 NPM_REGISTRY_OVERRIDE=${NPM_REGISTRY_OVERRIDE:-""}
-NPM_REGISTRY_RESET=${NPM_REGISTRY_RESET:-false}
 
 while getopts "t:d:r:nhp" opt; do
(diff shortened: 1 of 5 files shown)


bin/build
```diff
@@ -10,6 +10,9 @@ MIQ_ORG=${MIQ_ORG:-ManageIQ}
 SUI_ORG=${SUI_ORG:-ManageIQ}
 APPLIANCE_ORG=${APPLIANCE_ORG:-ManageIQ}
 
+NPM_REGISTRY_OVERRIDE=${NPM_REGISTRY_OVERRIDE:-""}
+NPM_REGISTRY_RESET=${NPM_REGISTRY_RESET:-false}
```
carbonin: Will we ever want to override and not reset? Can this just be one variable?

simaishi: "We" always want to reset, but not everyone. For example, if your build environment is same as where the images are used, you'd want to keep the non-default registry in the images so if you need to recompile assets, etc. it will continue to use non-default registry.

simaishi: @carbonin as per discussion, updated to always reset.

PR #357, opened 2020-01-07
Changes for adding ppc64le support

seth-priya:
In reference to https://github.com/ManageIQ/manageiq/issues/19676

Description of changes: Build PostgreSQL-10, phantomjs from source, install sassc gem with march-tune-native disabled. use a build arg called "ARCH"


bin/build
```diff
@@ -57,10 +60,11 @@ pushd $IMAGE_DIR
     echo "Building $image: $cmd"
     $cmd
   done
-
+
```
carbonin: Can you remove the trailing whitespace at the end of this line?

images/manageiq-base/Dockerfile
```diff
@@ -45,18 +47,43 @@ RUN yum -y install https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-re
...
+		   readline-devel   \
+		   zlib-devel	    \
+		   perl-devel	    \
+		   bison	    \
+		   flex		    \
                    &&               \
     yum clean all
 
+RUN wget  https://github.com/postgres/postgres/archive/REL_10_11.tar.gz && \
+    tar -zxvf REL_10_11.tar.gz && cd postgres-REL_10_11/ && ./configure --prefix=/usr && make && make install && \
```
carbonin: Would really prefer this rpm to be built outside of the Dockerfile.

We use copr for this. @simaishi would it be difficult to rebuild the rpms we host in copr for the different arch? Or should @seth-priya create a separate copr account for this?

images/manageiq-base/Dockerfile
```diff
@@ -45,18 +47,43 @@ RUN yum -y install https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-re
...
+# https://github.com/creationix/nvm#install-script
+RUN curl --silent -o- https://raw.githubusercontent.com/creationix/nvm/v0.31.2/install.sh | bash
+# install node and npm
+RUN source $NVM_DIR/nvm.sh \
+    && nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && nvm use default
+# add node and npm to path so the commands are available
+ENV NODE_PATH $NVM_DIR/v$NODE_VERSION/lib/node_modules
+ENV PATH $NVM_DIR/versions/node/v$NODE_VERSION/bin:$PATH
```
carbonin: Why are we installing `nvm` and `node` here? I would like to keep this as close to the appliance build as possible where we install the node rpm directly.

images/manageiq-base/Dockerfile
```diff
@@ -9,10 +9,13 @@ ARG MIQ_ORG=ManageIQ
 ARG SUI_ORG=ManageIQ
 ARG APPLIANCE_ORG=ManageIQ
 
+ARG ARCH=amd64
```
carbonin: I think I would prefer the default to be `x86_64` now that that's available in the new dumbinit release.

images/manageiq-ui-worker/Dockerfile
```diff
@@ -6,16 +6,22 @@ MAINTAINER ManageIQ https://manageiq.org
...
-RUN yum -y install httpd --setopt=tsflags=nodocs && \
+RUN yum -y install httpd sudo --setopt=tsflags=nodocs && \
     rm -f /etc/httpd/conf.d/* && \
     yum clean all
 
 COPY container-assets/npm_registry /tmp/npm_registry
 
+COPY container-assets/phantomjs_centos_7.sh /tmp/phantomjs_centos_7.sh
+
+RUN if [ "${ARCH}" == "ppc64le" ] ; then sh /tmp/phantomjs_centos_7.sh ; fi
```
carbonin: I don't really understand what we're doing here. Can you give more details around why a change like this is needed?

images/manageiq-ui-worker/container-assets/phantomjs_centos_7.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Clone and build missing dependencies from source.
+wrkdir=`pwd`
+wget http://ftp.gnu.org/pub/gnu/gperf/gperf-3.0.4.tar.gz
+tar -xzf gperf-3.0.4.tar.gz
+cd $wrkdir/gperf-3.0.4 && ./configure && make && sudo make install
+
+# Clone PhantomJS code and build it.
+cd $wrkdir
+git clone git://github.com/ariya/phantomjs.git
```
carbonin: I'm not really a javascript person ... what is phantomjs? @himdel can you help me out here? Is this something we want to be using?

images/manageiq-ui-worker/Dockerfile
```diff
@@ -6,16 +6,22 @@ MAINTAINER ManageIQ https://manageiq.org
 
 ARG NPM_REGISTRY_OVERRIDE
 
+ARG ARCH=amd64
```
carbonin: Same here, I would prefer the default to be `x86_64`

images/manageiq-ui-worker/Dockerfile
```diff
@@ -6,16 +6,22 @@ MAINTAINER ManageIQ https://manageiq.org
...
+
 LABEL name="manageiq-ui-worker" \
       summary="ManageIQ user interface worker image"
 
 # Install httpd and remove the existing httpd config from manageiq-appliance
-RUN yum -y install httpd --setopt=tsflags=nodocs && \
+RUN yum -y install httpd sudo --setopt=tsflags=nodocs && \
```
carbonin: Don't think we should need sudo during the build.

images/manageiq-base/Dockerfile
```diff
@@ -27,8 +30,7 @@ LABEL name="manageiq-base" \
       io.k8s.description="ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures." \
       io.openshift.tags="ManageIQ,miq,manageiq"
 
-RUN yum -y install https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm && \
```
carbonin: Wonder if we could contact https://github.com/nodesource/distributions about getting ppc64le support for node builds.

I would rather open an issue over there first rather than completely change how we build and install node.

images/manageiq-base/Dockerfile
simaishi: I think we should add another arch to the existing ones. I'll give it a try.

images/manageiq-ui-worker/container-assets/phantomjs_centos_7.sh
himdel: Ah, we don't need phantomjs at runtime at all and never needed it.

Until https://github.com/ManageIQ/manageiq-ui-classic/pull/5437 (in ivanchuk+), phantomjs was needed to run `rake spec:javascript` in ui-classic,
and it is still needed to run `yarn test` in ui-service.

But if you don't need to run tests, phantomjs should not be needed :).

himdel: >  what is phantomjs?

It's the predecessor to chrome-headless. Basically a js engine + a virtual dom implementation, to simulate a proper browser, with instrumentation.

bin/build
```diff
@@ -42,7 +44,8 @@ pushd $IMAGE_DIR
                     --build-arg APPLIANCE_REF=$APPLIANCE_REF \
                     --build-arg MIQ_ORG=$MIQ_ORG \
                     --build-arg SUI_ORG=$SUI_ORG \
-                    --build-arg APPLIANCE_ORG=$APPLIANCE_ORG"
+                    --build-arg APPLIANCE_ORG=$APPLIANCE_ORG \
+		    --build-arg ARCH=$ARCH"
```
Fryguy: I'm also noticing some whitespace inconsistency here and elsewhere in the file (tabs here where the rest is spaces)...can you please fix?

seth-priya:
thanks for your review @carbonin I will have a look and get back soon

images/manageiq-base/Dockerfile
simaishi: @seth-priya I built postgresql rpm. I think that's the only rpm needed. Please test the rpm using https://copr.fedorainfracloud.org/coprs/simaishi/test/repo/epel-7/simaishi-test-epel-7.repo

If it looks good, I'll make the RPM available in manageiq/ManageIQ-Master.

seth-priya: thanks @simaishi I will test the RPM

images/manageiq-base/Dockerfile
seth-priya: okay

images/manageiq-ui-worker/Dockerfile
seth-priya: okay

images/manageiq-base/Dockerfile
seth-priya: I will give it a try

images/manageiq-ui-worker/Dockerfile
seth-priya: removed, it was for executing the phantomjs script which was created to run in non-root user mode.

images/manageiq-ui-worker/Dockerfile
seth-priya: there is an internal dependency on phantomjs, binary for which is not present for ppc64le, this is how the build fails for this image if I don't build phantomjs. would it be possible to host this RPM with the PostgreSQL and ruby-install at copr? The phantomjs community itself is no longer active :(

[5/5] Building fresh packages...
error /opt/manageiq/manageiq-ui-service/node_modules/phantomjs-prebuilt: Command failed.
Exit code: 1
Command: node install.js
Arguments:
Directory: /opt/manageiq/manageiq-ui-service/node_modules/phantomjs-prebuilt
Output:
PhantomJS not found on PATH
Unexpected platform or architecture: linux/ppc64
It seems there is no binary available for your platform/architecture
Try to install PhantomJS globally
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
The command '/bin/sh -c source /etc/default/evm &&     cd ${SUI_ROOT} &&     yarn install &&     yarn run available-languages &&     yarn run build &&     yarn cache clean' returned a non-zero code: 1

bin/build
seth-priya: done

images/manageiq-base/Dockerfile
seth-priya: because this does not work https://github.com/ManageIQ/manageiq-pods/blob/42a6332eeae074c4c2aae8c87846abc80c5f5091/images/manageiq-base/Dockerfile#L48 as https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-release-el7-1.noarch.rpm is not yet available for Power

seth-priya:
@carbonin thanks again for your review and sorry for the slow response, was out sick, I have made some of the changes already on my local setup and now will do a round of testing on both Power and Intel before pushing them, I will start a discussion with nodesoure community and test the PostgreSQL rpm. I have answered the query on PhantomJS above and made a request, please do let me know your guidance, thank you!

images/manageiq-ui-worker/Dockerfile
himdel: Ah, you're right, there's currently no way to install SUI build dependencies without installing SUI test dependencies.

That said, all you really need is for a `phantomjs` binary to exist in the path, so *if* we don't want phantomjs, I think it can be replaced by `sudo ln -s /bin/true /bin/phantomjs`.

carbonin: @himdel Does this mean that we are currently installing phantomjs as a part of building the SUI in some other way? Why does that way not work for this case? Or is it related to the node install issues?

When I first saw this it looked like we were installing something new.

himdel: The normal build process does not do the building on the target machine, so we should be fine there. (On regular appliances, SUI `/var/www/miq/vmdb/public/ui/service` is the ouptut of `yarn run build` in SUI, but does not include SUI sources or `node_modules`.)

images/manageiq-base/Dockerfile
seth-priya: @simaishi I have tested the postgresql RPM and seems to work perfectly, I will update the PR with my changes, once you publish it to the main repo, thank you!!!

images/manageiq-base/Dockerfile
seth-priya: Created issue here, https://github.com/nodesource/distributions/issues/996

images/manageiq-ui-worker/container-assets/phantomjs_centos_7.sh
seth-priya: So, it seems to be coming from https://github.com/ManageIQ/manageiq-ui-service/blob/663dd18b3ae4fda4e09056f4281ce3be3818bae1/package.json#L128 a
nd https://github.com/ManageIQ/manageiq-ui-service/blob/663dd18b3ae4fda4e09056f4281ce3be3818bae1/package.json#L138, 
and is an attempt to install is made when we build the service ui from the manageiq-ui-worker Dockerfile, any thoughts on how to handle this?

himdel: As above - https://github.com/ManageIQ/manageiq-pods/pull/357#discussion_r365205107 and below, I believe there are 3 choices:

* provide a fake phantomjs binary that just knows how to react to `--version`
* build SUI elsewhere, outside the container, and only copy the `yarn build` output to the target container
* install phantomjs

himdel: (should you go with #4, `sed -i '/phantomjs/d' package.json` does the trick for me, there are no hidden dependency chains, only `phantomjs-prebuilt` and `karma-phantomjs-launcher`)

images/manageiq-base/Dockerfile
simaishi: In manageiq/ManageIQ-Master now. Thanks!

seth-priya:
I have addressed and resolved all the review comments in the latest commit, except those related to nodesource and usage of nvm to install node as I have not heard back from that community, please review and suggest, thank you

carbonin:
@seth-priya so I went looking for alternative node architecture rpms (or source rpms so we could build it) and I happened to come across http://public.dhe.ibm.com/software/ibmi/products/pase/rpms/repo/ppc64/ Is that what you're looking for? I noticed that the architecture is slightly different from the one you mention here ...

images/manageiq-base/Dockerfile
```diff
@@ -45,18 +47,32 @@ RUN yum -y install https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-re
...
                    net-tools        \
                    nmap-ncat        \
-                   nodejs           \
                    openldap-clients \
                    openscap-scanner \
                    patch            \
                    psmisc           \
-                   postgresql-devel \
                    sqlite-devel     \
+                   postgresql-devel \
```
carbonin: Why was this changed here?

images/manageiq-base/Dockerfile
```diff
@@ -9,10 +9,13 @@ ARG MIQ_ORG=ManageIQ
 ARG SUI_ORG=ManageIQ
 ARG APPLIANCE_ORG=ManageIQ
 
+ARG ARCH=x86_64
+
 ENV TERM=xterm \
     CONTAINER=true \
     APP_ROOT=/var/www/miq/vmdb \
     APPLIANCE_ROOT=/opt/manageiq/manageiq-appliance \
+    APPLIANCE_PG_DATA=/var/lib/pgsql/data \
```
carbonin: Why was this added?

images/manageiq-base/Dockerfile
```diff
@@ -78,6 +94,7 @@ WORKDIR ${APP_ROOT}
 RUN source /etc/default/evm && \
     export RAILS_USE_MEMORY_STORE="true" && \
     gem install bundler -v ">=1.16.2" && \
+    if [ ${ARCH} = "ppc64le" ] ; then gem install sassc  -- --disable-march-tune-native ; fi && \
```
carbonin: What is this used for and why is it being installed here rather than as a part of the bundle?

images/manageiq-base/Dockerfile
```diff
@@ -45,18 +47,32 @@ RUN yum -y install https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-re
...
                    which            \
                    &&               \
     yum clean all
 
+# nvm environment variables
+ENV NVM_DIR /usr/local/nvm
+ENV NODE_VERSION 10.17.0
+# install nvm
+# https://github.com/creationix/nvm#install-script
+RUN curl --silent -o- https://raw.githubusercontent.com/creationix/nvm/v0.31.2/install.sh | bash
```
carbonin: If we do have to go this direction, is there a better way to install nvm? I'm not really a fan of `curl ... | bash`

himdel:
> @simaishi @himdel do we have any other options for alternative node installations? Is this another one we would be better off building ourselves or is nvm really the choice here?

nvm is the preferred (and as far as I know, only) way if the distro doesn't provide any acceptable node version.

(I have Opinions about distributions which do not provide an up-to-date version of node as a core package these days, but let's not go into that, I assume switching distros is outside the scope of this PR :).)


> If we do have to go this direction, is there a better way to install nvm? I'm not really a fan of curl ... | bash

I get what you mean, the documentation also suggests `wget | bash` :). As far as I know, the only alernative is `git clone https://github.com/nvm-sh/nvm.git ; cd nvm ; git checkout (version tag) ./install.sh` which boils down to running exactly the same shell file, so, not sure that's worth it.


IMO the only thing we can really do about nvm is to build node for that architecture and push it to EPEL.

himdel:
One addition .. if *any* version of node exists for that platform, we could also install that version, and then use `npm install node-linux-ppc64le@10` to download a pre-built version of node from npm.

But no node means no npm, and I would not go the way of reimplementing npm here either :).

simaishi:
I tried building node RPM for ppc64le, but that will require building more than a few dependencies (and who knows what kind of dependencies they will bring...) and I don't think it's worth it.

In CentOS 8, I'll be changing to get nodejs from AppStream. So until we move to 8, we should probably just do what we can to get all arch working.

himdel:
One more option - downloading prebuilt binaries manually from https://nodejs.org/dist/v10.18.1/

snehlatamohite:
@himdel @simaishi  - I will try downloading prebuilt binaries and will check if it works.

images/manageiq-ui-worker/Dockerfile
himdel: probably resolved here already, but heads up,
dropping phantomjs from service ui in https://github.com/ManageIQ/manageiq-ui-service/pull/1619

Do let me know if backporting it is desired :)

seth-priya:
@carbonin @himdel @simaishi thanks all for your comments and suggestions - so, yes, these https://nodejs.org/dist/v10.18.1/ should work just fine - we have used these on other projects before, so I don't forsee any issues. I will make the required changes, validate them on both the architectures and get back. 

Sorry for the delayed response here, I am just getting back to work after being out sick for a couple of weeks :( , @snehlatamohite thanks for pitching in while I was away.

images/manageiq-base/Dockerfile
seth-priya: this was needed because of the difference in the way in which we were installing postgresql on ppc64le as compared to Intel, now that we have consistency there, this can be removed, I will test

images/manageiq-base/Dockerfile
seth-priya: There is more context around why this was needed here https://github.com/ManageIQ/manageiq/issues/19520
Essentially the build fails while installing (or building) this gem if the --disable-march-tune-native flag is not set for ppc64le

images/manageiq-base/Dockerfile
seth-priya: Changes atop each other caused a sequence change, sorry about it, will correct

images/manageiq-base/Dockerfile
seth-priya: Yep, will replace this by installing node from here https://nodejs.org/dist/v10.18.1/

seth-priya:
@carbonin @himdel @simaishi I have made and validated the required changes on ppc64le and x86_64 - requesting you to please review and let me know your feedback, thank you.

bin/build
```diff
@@ -55,7 +62,9 @@ pushd $IMAGE_DIR
                     --build-arg SUI_ORG=$SUI_ORG \
                     --build-arg APPLIANCE_ORG=$APPLIANCE_ORG \
                     --build-arg GIT_HOST=$GIT_HOST \
-                    --build-arg CORE_REPO_NAME=$CORE_REPO_NAME"
+                    --build-arg CORE_REPO_NAME=$CORE_REPO_NAME \
+                    --build-arg ARCH=$ARCH
```
carbonin: Think you're missing a `\` here.

images/manageiq-base/Dockerfile
```diff
@@ -62,6 +62,15 @@ RUN yum -y install https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-re
                    &&               \
     yum clean all
 
+ENV NODE_VERSION=v10.17.0
+ENV NODE_DISTRO=linux-${NODE_ARCH}
+
+RUN mkdir -p /usr/local/lib/nodejs && \
+    wget https://nodejs.org/dist/${NODE_VERSION}/node-${NODE_VERSION}-${NODE_DISTRO}.tar.xz && \
+    tar -xJvf node-${NODE_VERSION}-${NODE_DISTRO}.tar.xz -C /usr/local/lib/nodejs
```
carbonin: Can you combine these?
`wget -O - ... | tar -C ..`

images/manageiq-ui-worker/Dockerfile
```diff
@@ -6,6 +6,8 @@ MAINTAINER ManageIQ https://manageiq.org
 
 ARG NPM_REGISTRY_OVERRIDE
 
+ARG ARCH=x86_64
```
carbonin: This doesn't look used, in this Dockerfile, do we need it here?

seth-priya:
@carbonin thanks for all your support on this!

Issue #358, opened 2020-01-07
Allow use of private GitHub/GitLab


Fryguy:
- need to parameterize the images builds to change
  - https://github.com -> $GITHUB_HOST
  - manageiq- prefix -> $REPO_PREFIX
  - ability to fetch a token from host.docker.internal or something
    - $FETCH_TOKEN would drive whether or not to fetch the token, and the token would be stored in a different ENV var

Fryguy:
FYI @carbonin

carbonin:
Closed via #362

PR #362, opened 2020-01-09
Add support for github enterprise

sethpeterson:
Adds support for specifying a particular Github endpoint and Github token.

Issue: https://github.com/ManageIQ/manageiq-pods/issues/358

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -47,7 +47,7 @@ pushd $IMAGE_DIR
   
   if [[ -n "$GIT_TOKEN" ]] ; then
     SECRETS_CONTAINER=$(docker run -d --rm busybox /bin/sh -c \
-      "echo $GIT_TOKEN > git_token.txt && httpd -p 8000 -h / && sleep 600")
+      "echo $GIT_TOKEN > git_token.txt && httpd -p 8000 -h / && sleep 1200")
   fi
 
   cmd="docker build --tag $REPO/manageiq-base:$TAG \
@@ -61,7 +61,7 @@ pushd $IMAGE_DIR
                     --build-arg REPO_PREFIX=$REPO_PREFIX"
 
   if [[ -n "$GIT_TOKEN" ]] ; then
-    cmd+=" --build-arg GIT_TOKEN_AVAILABLE=true --network=container:secrets-server"
+    cmd+=" --build-arg GIT_TOKEN_AVAILABLE=true --network=container:$SECRETS_CONTAINER"
   fi
 
   if [ -n "$NO_CACHE" ]; then
--- a/bin/build
+++ b/bin/build
@@ -10,16 +10,24 @@ MIQ_ORG=${MIQ_ORG:-ManageIQ}
 SUI_ORG=${SUI_ORG:-ManageIQ}
 APPLIANCE_ORG=${APPLIANCE_ORG:-ManageIQ}
 
+REPO_PREFIX=${REPO_PREFIX:-manageiq}
+
+GIT_HOST=${GIT_HOST:-github.com}
+GIT_TOKEN=${GIT_TOKEN:-""}
+
 NPM_REGISTRY_OVERRIDE=${NPM_REGISTRY_OVERRIDE:-""}
 
-while getopts "t:d:r:nhp" opt; do
+while getopts "t:d:f:g:k:r:nhp" opt; do
   case $opt in
     d) IMAGE_DIR=$OPTARG ;;
+    f) REPO_PREFIX=$OPTARG ;;
+    g) GIT_HOST=$OPTARG ;;
+    k) GIT_TOKEN=$OPTARG ;;
     n) NO_CACHE="true" ;;
     p) PUSH="true" ;;
     r) REPO=$OPTARG ;;
     t) TAG=$OPTARG ;;
-    h) echo "Usage: $0 -d IMAGE_DIR -r IMAGE_REPOSITORY [-hnp] [ -t IMAGE_TAG ]"; exit 1
+    h) echo "Usage: $0 -d IMAGE_DIR -r IMAGE_REPOSITORY [-hfgknp] [ -t IMAGE_TAG ]"; exit 1
   esac
 done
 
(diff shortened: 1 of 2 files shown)


bin/build
```diff
@@ -36,13 +44,25 @@ fi
 set -e
 
 pushd $IMAGE_DIR
+  
+  if [[ -n "$GIT_TOKEN" ]] ; then
+    SECRETS_CONTAINER=$(docker run -d --rm busybox /bin/sh -c \
+      "echo $GIT_TOKEN > git_token.txt && httpd -p 8000 -h / && sleep 600")
```
carbonin: Why go through the effort of creating a container to serve up the token if it's going to be visible on the command line for the build script anyway?

Would to be simpler to make it an env var that the script would consume rather than an option?

bin/build
```diff
@@ -36,13 +44,25 @@ fi
...
                     --build-arg APPLIANCE_REF=$APPLIANCE_REF \
                     --build-arg MIQ_ORG=$MIQ_ORG \
                     --build-arg SUI_ORG=$SUI_ORG \
-                    --build-arg APPLIANCE_ORG=$APPLIANCE_ORG"
+                    --build-arg APPLIANCE_ORG=$APPLIANCE_ORG \
+                    --build-arg GIT_HOST=$GIT_HOST \
+                    --build-arg REPO_PREFIX=$REPO_PREFIX"
+
+  if [[ -n "$GIT_TOKEN" ]] ; then
+    cmd+=" --build-arg GIT_TOKEN_AVAILABLE=true --network=container:secrets-server"
```
carbonin: If we avoid the container above then we will also be able to just key off of the presence of $GIT_TOKEN rather than creating a separate build arg to indicate that it is available.

images/manageiq-base/Dockerfile
```diff
@@ -57,16 +62,28 @@ RUN yum -y install https://rpm.nodesource.com/pub_10.x/el/7/x86_64/nodesource-re
...
+      export GIT_HOST=${GIT_TOKEN}@${GIT_HOST} ; \
+    fi && \
+    mkdir -p ${APPLIANCE_ROOT} && \
+    curl -L https://${GIT_HOST}/${APPLIANCE_ORG}/${REPO_PREFIX}-appliance/tarball/${APPLIANCE_REF} | tar vxz -C ${APPLIANCE_ROOT} --strip 1
 
-RUN mkdir -p ${SUI_ROOT} && \
-    curl -L https://github.com/${SUI_ORG}/manageiq-ui-service/tarball/${SUI_REF} | tar vxz -C ${SUI_ROOT} --strip 1
+RUN if [[ -n "$GIT_TOKEN_AVAILABLE" ]] ; then \
+      export GIT_TOKEN=$(wget -O - -q http://localhost:8000/git_token.txt) ; \
+      export GIT_HOST=${GIT_TOKEN}@${GIT_HOST} ; \
```
carbonin: This format makes it feel like the token is a `user:pass` format rather than a single token. If that's the case maybe a better name would be `GIT_AUTH`?

bin/build
```diff
@@ -10,16 +10,24 @@ MIQ_ORG=${MIQ_ORG:-ManageIQ}
 SUI_ORG=${SUI_ORG:-ManageIQ}
 APPLIANCE_ORG=${APPLIANCE_ORG:-ManageIQ}
 
+REPO_PREFIX=${REPO_PREFIX:-manageiq}
```
carbonin: I think a better name for this would be `CORE_REPO_NAME` in the Dockerfile it would make more sense to see `$CORE_REPO_NAME-appliance` rather than an entire repo name that consists of a "prefix".

bin/build
```diff
@@ -10,16 +10,24 @@ MIQ_ORG=${MIQ_ORG:-ManageIQ}
...
+
 NPM_REGISTRY_OVERRIDE=${NPM_REGISTRY_OVERRIDE:-""}
 
-while getopts "t:d:r:nhp" opt; do
+while getopts "t:d:f:g:k:r:nhp" opt; do
   case $opt in
     d) IMAGE_DIR=$OPTARG ;;
+    f) REPO_PREFIX=$OPTARG ;;
+    g) GIT_HOST=$OPTARG ;;
+    k) GIT_TOKEN=$OPTARG ;;
```
carbonin: I'm struggling to come up with meaningful option letters for these ... maybe they should all be driven by environment variables like `NPM_REGISTRY_OVERRIDE`.

bin/build
```diff
@@ -10,16 +10,24 @@ MIQ_ORG=${MIQ_ORG:-ManageIQ}
 SUI_ORG=${SUI_ORG:-ManageIQ}
 APPLIANCE_ORG=${APPLIANCE_ORG:-ManageIQ}
 
+REPO_PREFIX=${REPO_PREFIX:-manageiq}
+
+GIT_HOST=${GIT_HOST:-github.com}
```
Fryguy: This should probably include the `https://` bit as well, in case people want to use something else like `ssh://`

bin/build
Fryguy: do env vars get baked into the image history?  if so, then we can't use a defined ENV.  What this does is expose a sidecar service during build only that the docker build can call into (and thus store into temporary memory)

images/manageiq-base/Dockerfile
Fryguy: yeah i like GIT_AUTH

bin/build
sethpeterson: The purpose of the sidecar container is to hide the token from image history.
Looking at the image history after build, this is what I see for that layer:
```sha256:ab871f0a6bbcecdeb54c7e0a5f27aa22c786a777245e044fcb4f04cec982e83e   12 minutes ago      |9 APPLIANCE_ORG=IBMPrivateCloud APPLIANCE_REF=master GIT_HOST=github.ibm.com GIT_TOKEN_AVAILABLE=true MIQ_ORG=IBMPrivateCloud MIQ_REF=master REPO_PREFIX=bluecf SUI_ORG=IBMPrivateCloud SUI_REF=master /bin/sh -c if [[ -n "$GIT_TOKEN_AVAILABLE" ]] ; then       export GIT_TOKEN=$(wget -O - -q http://localhost:8000/git_token.txt) ;       export GIT_HOST=${GIT_TOKEN}@${GIT_HOST} ;     fi &&     echo $GIT_TOKEN &&     mkdir -p ${APPLIANCE_ROOT} &&     echo "https://${GIT_HOST}/${APPLIANCE_ORG}/${REPO_PREFIX}-appliance/tarball/${APPLIANCE_REF}" &&     curl -L https://${GIT_HOST}/${APPLIANCE_ORG}/${REPO_PREFIX}-appliance/tarball/${APPLIANCE_REF} | tar vxz -C ${APPLIANCE_ROOT} --strip 1 ```

I think that keeps the token hidden

bin/build
sethpeterson: If we include the `https://` bit then building the url gets significantly more complicated. Maybe we could leave it like this for now and if we need to add support for a ssh key, we could refactor it then?

carbonin:
So regarding the secrets and the sidecar container...

I think this is fine to get you guys working, but I would rather investigate something like the built-in [`--secret` option to `docker build`](https://docs.docker.com/develop/develop-images/build_enhancements/#new-docker-build-secret-information) or possibly using multi-stage builds as described in https://dev.to/lirantal/are-you-building-docker-images-here-s-how-to-avoid-leaking-sensitive-information-into-docker-images-99g

Personally, if multi-stage builds solves this problem, I would prefer to go that route to keep the build process compatible with both `docker` and `buildah` as [`buildah` doesn't support the `--secret` option yet](https://github.com/containers/buildah/issues/1684). Either way I think we would keep the interface to the build script the same as it's defined in this patch so hopefully it wouldn't be too disruptive of a change.

carbonin:
Thoughts on :point_up: https://github.com/ManageIQ/manageiq-pods/pull/362#issuecomment-575294949 @Fryguy @sethpeterson ?

sethpeterson:
@carbonin I prefer multi stage build to docker --secret because the --secret argument isn't as widely supported yet. I'm open to doing that vs this method of a sidecar container.

carbonin:
:+1: I agree @sethpeterson. I'll leave it up to you about timing. Do you want to take a look at solving this with multi-stage builds for this patch or a follow-up?

sethpeterson:
I think we should do it in a follow up so that we can continue making progress here

PR #366, opened 2020-01-30
Create the ui, api, and remote console services in templates

carbonin:
Nothing in these services will change at runtime and the name
maps directly to the contents of the httpd config.


carbonin:
Merge with https://github.com/ManageIQ/manageiq/pull/19785

PR #367, opened 2020-02-17
Use multi-stage builds to remove need for secrets container

carbonin:
By using a multi-stage build we can include our secret in a build
arg but then only copy over the files we need, throwing away the
layers that contain the reference to the secret and leaving no trace
on the filesystem.

@sethrpeterson @bennett-white does this handle the case you were trying to solve?


Fryguy:
LGTM

Issue #360, opened 2020-01-07
Allow for productization of image builds


Fryguy:
This is possibly done by insert the following lines in the manageiq-base Dockerfile after curling the 3 repos...maybe 😝 

```
ENV PRODUCTIZATION_ROOT=/opt/manageiq/miq_productization
RUN mkdir -p ${PRODUCTIZATION_ROOT} && \
    curl -L https://github.com/${PRODUCTIZATION_ORG}/miq_productization/tarball/${PRODUCTUZATION_REF} | tar vxz -C ${PRODUCTIZATION_ROOT} --strip 1
RUN cd ${PRODUCTIZATION_ROOT} && \
    ./link.rb
```

Fryguy:
Part of productization also includes changing the product name.  However, product name is coupled to the name of the objects being deployed [1], which causes issues, since the template and the generated names don't match.  We should avoid coupling those 2 things, by introducing an ENV var, such as PRODUCT_PREFIX, which would be used as the prefix instead hardcoding [2] or relying on the product name [3].

[1] https://github.com/ManageIQ/manageiq/blob/master/lib/container_orchestrator/object_definition.rb#L91-L93
[2] https://github.com/ManageIQ/manageiq/blob/master/lib/container_orchestrator/object_definition.rb#L17
[3] https://github.com/ManageIQ/manageiq/blob/master/lib/container_orchestrator/object_definition.rb#L75

cc @carbonin

carbonin:
@Fryguy So I can implement what you put in the first comment if we create a `https://github.com/ManageIQ/miq_productization` repo with a no-op `link` scrip for now (thinking `/bin/true`). Then anyone who wants can fork that repo and commit whatever they want.

Fryguy:
Oh, like, for example purposes?  Or you literally want to clone that repo?  I was thinking it would be ENV var driven, and in the abscense of the ENV var we wouldn't clone anything.

chessbyte:
@carbonin @Fryguy let's keep this discussion moving forward, as I would like this to make it into the early timeframe of the Jansa release (beta)

PR #369, opened 2020-02-18
Allow for changing the app name using a parameter

carbonin:
This PR adds a new parameter (`APP_NAME`) to the templates which allows for changing the name of the deployed application.

This solves part of https://github.com/ManageIQ/manageiq-pods/issues/360

### DIFF
--- a/templates/app/app-secrets.yaml
+++ b/templates/app/app-secrets.yaml
@@ -1,12 +1,12 @@
 apiVersion: v1
 kind: Template
 metadata:
-  name: manageiq-secrets
+  name: app-secrets
 objects:
 - apiVersion: v1
   kind: Secret
   metadata:
-    name: manageiq-secrets
+    name: app-secrets
     labels:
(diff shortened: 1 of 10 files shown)


templates/app/app-secrets.yaml
```diff
@@ -1,18 +1,20 @@
 apiVersion: v1
 kind: Template
 metadata:
-  name: manageiq-secrets
+  name: app-secrets
```
bdunne: Should these be `${APP_NAME}-secrets` as well?

carbonin: Nope, explained that in the first commit message actually.

> It's hard to productize based on a configurable app name when
we're referencing the object name in the deploy script.
This avoids the problem by not using the app name in the secret name
at all.

carbonin:
Sorry, should have mentioned that this should be merged with https://github.com/ManageIQ/manageiq/pull/19847

PR #370, opened 2020-02-20
Add IMAGE_PULL_SECRET parameter

carbonin:
This PR adds a parameter for a secret name to be used for the orchestrator and worker images.

This secret should be pre-created and contain the credentials for the registry specified in the `ORCHESTRATOR_IMAGE_NAMESPACE` parameter.

Additionally it removes the app name prefix for the orchestrator deployment.

https://github.com/ManageIQ/manageiq/issues/19699


carbonin:
Merge with https://github.com/ManageIQ/manageiq/pull/19856

Issue #371, opened 2020-02-26
Base containers on ubi8


carbonin:

carbonin:
Fixed in https://github.com/ManageIQ/manageiq-pods/pull/372

PR #372, opened 2020-02-28
Base images on ubi8

carbonin:
This commit substitutes dnf for yum, installs ruby in the manageiq-base
image, and adds some packages that were missing from the base image.

Specifically:
hostname for the "hostname" command
iproute for the "ip" command
libxml2-devel for the nokogiri gem
make for building gems
openssl-devel for the nokogiri gem
redhat-rpm-config for building gems
wget for the "wget" command
xz for use with tar -J flag

### DIFF
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:7
+FROM registry.access.redhat.com/ubi8/ubi
 
 ARG MIQ_REF=master
 ARG SUI_REF=master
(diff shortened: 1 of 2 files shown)


images/manageiq-base/Dockerfile
```diff
@@ -48,33 +48,42 @@ LABEL name="manageiq-base" \
...
-                   openscap-scanner \
-                   patch            \
-                   psmisc           \
-                   postgresql-devel \
-                   sqlite-devel     \
-                   sysvinit-tools   \
-                   which            \
-                   &&               \
-    yum clean all
+RUN dnf -y --disableplugin=subscription-manager install http://mirror.centos.org/centos/8/BaseOS/${ARCH}/os/Packages/centos-repos-8.1-1.1911.0.8.el8.${ARCH}.rpm http://mirror.centos.org/centos/8/BaseOS/${ARCH}/os/Packages/centos-gpg-keys-8.1-1.1911.0.8.el8.noarch.rpm && \
```
simaishi: A note... because centos-repos is being added here, most of the packages installed below is coming from CentOS repos, not UBI8 repos, even if available in UBI8. It's the same version, but different binaries built by CentOS. If a newer version is available in UBI8 repo, that will be picked up though.

simaishi:
@carbonin @seth-priya I think we can switch to using nodejs from AppStream now. (We're using v12 in master appliance build).

carbonin:
This fixes https://github.com/ManageIQ/manageiq-pods/issues/371

PR #376, opened 2020-03-03
[JANSA] Update tags to latest-jansa

Fryguy:
Now that we have the jansa branch, this branch needs to build from the correct sources and tag those builds accordingly.

@carbonin @simaishi Please review.


simaishi:
We should also change postgresql image to use tagged version not "latest", so when we move 'master' to later version, branches will continue using the same version.

Fryguy:
@carbonin For monolithic I pushed directly to ManageIQ/manageiq jansa branch because I have a script that does it (though I did find an issue).  Here is the change - https://github.com/ManageIQ/manageiq/blob/21b6d9cc6676f278e8cc83b9e92579d6d1e98fae/Dockerfile#L1

Issue #381, opened 2020-03-05
Add ansible/ansible-runner and python dependency modules


simaishi:
Related PR: https://github.com/ManageIQ/manageiq-appliance-build/pull/341

ansible comes from: epel ~~https://mirrors.fedoraproject.org/mirrorlist?repo=playground-epel8&arch=x86_64~~

ansible-runner comes from: https://releases.ansible.com/ansible-runner/rpm/epel-8-x86_64/

PR #382, opened 2020-03-06
Align package list with appliance

carbonin:
Fixes https://github.com/ManageIQ/manageiq-pods/issues/379 by adding the needed rpm packages
Fixes https://github.com/ManageIQ/manageiq-pods/issues/381 by installing the required python modules directly from https://raw.githubusercontent.com/ManageIQ/manageiq-appliance-build/master/kickstarts/partials/post/python_modules.ks.erb

This also removes some packages that are not needed in the container image:
```
chrony      - NTP client
cronie      - crond
initscripts - support for System V init
net-tools   - obsolete networking package replaced by iproute
psmisc      - process management utilities
which
```

I've created separate commits with justifications for each package, so this is probably easier to review looking at them separately.

~~Leaving this as WIP until I get the chance to test embedded ansible~~

I tried testing embedded ansible, but ran into issues that I think will need to be dealt with on the application side as well as some current problems with the master branch.

Will follow up with testing this after this PR is built into the image.


images/manageiq-base/Dockerfile
```diff
@@ -45,46 +45,65 @@ LABEL name="manageiq-base" \
...
+      sqlite-devel                \
+      v2v-conversion-host-ansible \
+      wget                        \
+      wmi                         \
+      xz                          \
+      &&                          \
     dnf clean all
 
+# Install python packages the same way the appliance does
+RUN curl -L https://raw.githubusercontent.com/ManageIQ/manageiq-appliance-build/master/kickstarts/partials/post/python_modules.ks.erb | bash -
```
simaishi: `master` needs to be replaced with appropriate branch as we cut branches. Not sure if it makes more sense to use ARG...  If not, one more place for @Fryguy to change.

Also, do we need to support other ARGs (GIT_AUTH, _ORG, etc.) here?

images/manageiq-base/Dockerfile
```diff
@@ -45,46 +45,65 @@ LABEL name="manageiq-base" \
...
+      openldap-clients            \
+      openscap-scanner            \
+      openssl-devel               \
+      OpenIPMI                    \
+      patch                       \
+      postfix                     \
+      postgresql-devel            \
+      procps-ng                   \
+      python3-devel               \
+      qpid-proton-c-devel         \
```
simaishi: If we're adding the `qpid-proton-c-devel` library, we should run bundle with `--with qpid_proton` option.

images/manageiq-base/Dockerfile
carbonin: Hm, possibly. Right now we don't actually pull anything from appliance-build so this is net-new.

I guess our options are to either treat it the same as the other repos including ARGs, AUTH, etc. Or just copy the content like we do for the packages. What do you think?

simaishi:
Jansa backport details (replaced reference to 'master' with `jansa`) :

```
$ git log -1
commit 61c49e8f5f6ac8d0563bc33c81f0f29b7ee1acfa
Date:   Thu Mar 12 18:10:02 2020 -0400

    Merge pull request #382 from carbonin/add_missing_packages

    Align package list with appliance

    (cherry picked from commit aa14ff0279305070789d3aa40a691efd6e6cda7b)
```

PR #385, opened 2020-03-12
Pull latest ubi8 image for build

simaishi:
Without `--pull`, it uses the image on local machine if exists, and doesn't check to see if a newer version is available.

I only added the option for manageiq-base build so it will pick up the latest ubi8 image.  For manageiq-* images, I thought you'd want to use local copy if exists, even if that's older than what's on Docker hub.

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -43,6 +43,7 @@ set -e
 
 pushd $IMAGE_DIR
   cmd="docker build --tag $REPO/manageiq-base:$TAG \
+                    --pull \
                     --build-arg MIQ_REF=$MIQ_REF \
                     --build-arg SUI_REF=$SUI_REF\
                     --build-arg APPLIANCE_REF=$APPLIANCE_REF \


carbonin:
Would it be better to add `:latest` to https://github.com/ManageIQ/manageiq-pods/blob/master/images/manageiq-base/Dockerfile#L28 ?

Would that even accomplish the same thing?

simaishi:
> Would it be better to add `:latest` to https://github.com/ManageIQ/manageiq-pods/blob/master/images/manageiq-base/Dockerfile#L28 ?
> 
> Would that even accomplish the same thing?

No, it won't... even without `:latest`, it already pulls the image as `:latest`, so it's considered as the image exists.

simaishi:
Jansa backport details:

```
$ git log -1
commit db0ab68cb49d21ebe746788cd65a16d4f5b7889b
Date:   Thu Mar 12 11:37:43 2020 -0400

    Merge pull request #385 from simaishi/pull_latest

    Pull latest ubi8 image for build

    (cherry picked from commit dd70f313279375a7add63964929569537788edec)
```

PR #389, opened 2020-03-23
added operator for manageiq-pods

### LINKED ISSUE #355: [Podification] Create Operator for Lifecycle Management
The expectation here is for a single operator that will pretty much handle what the deploy script does today.

I would expect the following parameters to continue to be configurable from the CR with the current default values to still be the defaults if not provided in the CR:
```
APP_NAME
APPLICATION_ADMIN_PASSWORD
APPLICATION_DOMAIN
DATABASE_HOSTNAME
DATABASE_NAME
DATABASE_PORT
DATABASE_USER
DATABASE_PASSWORD
DATABASE_REGION
DATABASE_VOLUME_CAPACITY
ENCRYPTION_KEY
GUID
HTTPD_CPU_REQ
HTTPD_IMAGE_NAME
HTTPD_IMAGE_TAG
HTTPD_MEM_LIMIT
HTTPD_MEM_REQ
IMAGE_PULL_SECRET
MEMCACHED_CPU_REQ
MEMCACHED_IMAGE_NAME
MEMCACHED_IMAGE_TAG
MEMCACHED_MEM_LIMIT
MEMCACHED_MEM_REQ
ORCHESTRATOR_CPU_REQ
ORCHESTRATOR_IMAGE_NAME
ORCHESTRATOR_IMAGE_NAMESPACE
ORCHESTRATOR_IMAGE_TAG
ORCHESTRATOR_MEM_LIMIT
ORCHESTRATOR_MEM_REQ
POSTGRESQL_CPU_REQ
POSTGRESQL_IMG_NAME
POSTGRESQL_IMG_TAG
POSTGRESQL_MAX_CONNECTIONS
POSTGRESQL_MEM_LIMIT
POSTGRESQL_MEM_REQ
```

The operator should handle deploying, upgrading, and removing installations of manageiq by deploying the base dependencies (memcached, httpd, postgresql) and the orchestrator. It should not handle any use of manageiq such as adding or removing providers or changing settings.

---

- [x] Initial operator creation: [PR](https://github.com/ManageIQ/manageiq-pods/pull/389)
- [x] Build in the Dockerfile directly: [Issue](https://github.com/ManageIQ/manageiq-pods/issues/403) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397314395)
- [x] Add gofmt to CI: [PR](https://github.com/ManageIQ/manageiq-pods/pull/395) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#issuecomment-602876631)
- [x] Add build to CI - [Issue](https://github.com/ManageIQ/manageiq-pods/issues/404) [PR](https://github.com/ManageIQ/manageiq-pods/pull/412)
- [x] Set up CI to at least verify build ~~and gofmt~~: gofmt [PR](https://github.com/ManageIQ/manageiq-pods/pull/395) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#issuecomment-602876631)
- [x] Switch to go modules: [PR](https://github.com/ManageIQ/manageiq-pods/pull/400) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#issuecomment-603300089)
- [x] Update to latest boilerplate for operator-sdk: [Issue](https://github.com/ManageIQ/manageiq-pods/issues/405) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#issuecomment-603447748) in [PR](https://github.com/ManageIQ/manageiq-pods/pull/432)
- [x] Change "kind" from "Manageiq" to "ManageIQ": [Issue](https://github.com/ManageIQ/manageiq-pods/issues/406) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397324661) in [PR](https://github.com/ManageIQ/manageiq-pods/pull/444)
- [x] Change the boilerplate roles to something more restrictive: [Issue](https://github.com/ManageIQ/manageiq-pods/issues/407) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397319716) in [PR](https://github.com/ManageIQ/manageiq-pods/pull/413)
- [x] Fully spelled out keys: [Issue](https://github.com/ManageIQ/manageiq-pods/issues/408) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397322227) in [PR](https://github.com/ManageIQ/manageiq-pods/pull/416)
- [x] Certs available outside of the build: [Issue](https://github.com/ManageIQ/manageiq-pods/issues/409) originally discussed [here](https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397344335) in [PR1](https://github.com/ManageIQ/manageiq-pods/pull/431) and [PR2](https://github.com/ManageIQ/manageiq-pods/pull/445)
- [x] Do we need productizable kind names based on APP_NAME ?  Not even sure that's possible. [Issue](https://github.com/ManageIQ/manageiq-pods/issues/410)
- [x] Some other minor formatting cleanup and variable naming

mali-chainzee:
added operator for manageiq-pods.

Fixes #355


Fryguy:
@carbonin  please review.

Fryguy:
@mali-chainzee Can you please make it a proper PR as opposed to a submodule?  We want to live *in* this repo.  As is, it can't be reviewed.

mali-chainzee:
> @mali-chainzee Can you please make it a proper PR as opposed to a submodule? We want to live _in_ this repo. As is, it can't be reviewed.

i see, so the whole repo only contains the operator?

Fryguy:
well, the manageiq-operator directory would contain the source, but yes, the entire thing would live here.

mali-chainzee:
> well, the manageiq-operator directory would contain the source, but yes, the entire thing would live here.

i don't know how the review works, but do you mean the root directory of the repo should reflect the root directory of the operator? or i need to build the operator into image and put it on root dir?

carbonin:
I think @Fryguy wants all the code you developed for the operator to live in the `/manageiq-operator` directory (including the image). At least so we can take a look at the content.

carbonin:
Okay, given the size of the subtree, can you commit it in multiple pieces? Right now it's still impossible to review.

I think things that we won't really be reviewing, specifically dependencies should be one commit that we can mostly ignore. Then we would want another commit for operator and project scaffolding (I'm thinking things that declare our required dependencies or other files that we will hand-edit, but are not necessarily part of the operator itself). Then make a final commit for the actual operator source (CRD, image, operator app, etc).

Given what you know about how the project is structured, feel free to break it up further, but each commit should have a clear purpose and anything that we should actually be reviewing should have a reasonable change-set (at least a number of changes that GitHub will actually render).

mali-chainzee:
> Okay, given the size of the subtree, can you commit it in multiple pieces? Right now it's still impossible to review.
> 
> I think things that we won't really be reviewing, specifically dependencies should be one commit that we can mostly ignore. Then we would want another commit for operator and project scaffolding (I'm thinking things that declare our required dependencies or other files that we will hand-edit, but are not necessarily part of the operator itself). Then make a final commit for the actual operator source (CRD, image, operator app, etc).
> 
> Given what you know about how the project is structured, feel free to break it up further, but each commit should have a clear purpose and anything that we should actually be reviewing should have a reasonable change-set (at least a number of changes that GitHub will actually render).

yeah since the image has been built and pushed already, guess we can just keep the controller and cr code for reference.

mali-chainzee:
> > well, the manageiq-operator directory would contain the source, but yes, the entire thing would live here.
> 
> i don't know how the review works, but do you mean the root directory of the repo should reflect the root directory of the operator? or i need to build the operator into image and put it on root dir?

my bad, I didn't realize I did not push the actual code.

carbonin:
> yeah since the image has been built and pushed already, guess we can just keep the controller and cr code for reference.

What do you mean by this? We're going to continue to iterate on the operator in this project. If there's something that isn't needed for development and can be fetched only when we build the image, that's fine, but we still need everything that is required for continued development to be committed.

Fryguy:
@mali-chainzee Can you squash your commits, since the history of the project will still contain all those dependencies?

manageiq-operator/build/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM registry.access.redhat.com/ubi7/ubi-minimal:latest
```
Fryguy: Follow up PR - This should be ubi8

Fryguy:
Also,the binary files in build/_output should not be committed (should the entire directory just be .gitignored?)

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: manageiq.example.com/v1alpha1
+kind: Manageiq
+metadata:
+  name: miq
+spec:
+  # Add fields here
+  appName:  "manageiq"
+  applicationAdminPassword: "smartvm"
```
Fryguy: Minor, but there is trailing whitespace in a number of locations in the various files (here is one example), if you could clean that up.

manageiq-operator/pkg/helpers/httpd_conf/application.conf
```diff
@@ -0,0 +1,38 @@
+# Timeout: The number of seconds before receives and sends time out.
```
Fryguy: @carbonin Just an FYI, but this is creating yet another copy (might be required, so maybe we change templates-style to pull from here instead)

manageiq-operator/pkg/helpers/tlstools/generate_crt.go
```diff
@@ -0,0 +1,69 @@
...
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"flag"
+	"log"
+	"math/big"
+	"time"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var logs = logf.Log.WithName("controller_cf")
```
Fryguy: still says `cf` here or is this a different `cf`?

manageiq-operator/README.md
```diff
@@ -0,0 +1,64 @@
+# ManageIQ Golang Operator
+
+## Deploying ManageIQ resources:
+1. **Start Operator**
+    In project root directory, `operator-sdk up local --namespace=<yournamespace>`
+2. **Create custom resource**
+    `oc create -f deploy/crds/cf_v1alpha1_cf_cr.yaml`
```
Fryguy: Should be `manageiq_v1alpha_manageiq_cr.yaml`

manageiq-operator/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+## ManageIQ Custom Resource Structure (will change resource name from Cf to manageIQ)
```
Fryguy: I think think this cf comment can be removed.

manageiq-operator/README.md
```diff
@@ -0,0 +1,64 @@
...
+(Resource requests here were tailered to make it runnable on my local crc cluster)
+```apiVersion: cf.example.com/v1alpha1
```
Fryguy: A bunch of things in this example still say `Cf`

Fryguy: Also, I think you need a line break after the

    ```

otherwise the markdown is broken.

Fryguy:
Also please be sure to run `gofmt` on your files (there are ways to make your editor do it automatically on save).  The following files need to be `gofmt`ted

```
> gofmt -l manageiq-operator
manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
manageiq-operator/pkg/controller/controller.go
manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
manageiq-operator/pkg/helpers/miq-components/app-secret.go
manageiq-operator/pkg/helpers/miq-components/httpd.go
manageiq-operator/pkg/helpers/miq-components/memcached.go
manageiq-operator/pkg/helpers/miq-components/orchestrator.go
manageiq-operator/pkg/helpers/miq-components/postgresql-secrets.go
manageiq-operator/pkg/helpers/miq-components/postgresql.go
manageiq-operator/pkg/helpers/miq-components/rbac.go
manageiq-operator/pkg/helpers/randstring/randstring.go
manageiq-operator/pkg/helpers/tlstools/generate_crt.go
```

Fryguy:
Another follow up @carbonin , we should add `gofmt` to CI.

mali-chainzee:
> See requested changes above

@carbonin

manageiq-operator/Gopkg.toml
```diff
@@ -0,0 +1,74 @@
+# Force dep to vendor the code generators, which aren't imported just used at dev time.
+required = [
```
gyliu513: Why not use go mod?

manageiq-operator/README.md
```diff
@@ -0,0 +1,64 @@
+# ManageIQ Golang Operator
+
+## Deploying ManageIQ resources:
+1. **Start Operator**
+    In project root directory, `operator-sdk up local --namespace=<yournamespace>`
+2. **Create custom resource**
+    `oc create -f deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml`
+
+(Resource requests here were tailered to make it runnable on my local crc cluster)
+```
```
gyliu513: Add `yaml` to the format

```
```yaml
```

Fryguy:
@gyliu513 I don't mind merging with minor changes, but there are some things here that cannot be merged (for example, the PR presently has 2 million lines of code that need to be removed it they will be in the project history

mali-chainzee:
> See requested changes above



> See requested changes above

Just cleaned it up and squashed the commits

Fryguy:
Sorry accidentally hit the close button when trying to make a comment

Fryguy:
@mali-chainzee it seems you reintroduced computing the reps.  To avoid committing them you can use some .gitignore settings (is there a standard .gitignore for Go applications?)

manageiq-operator/Gopkg.toml
Fryguy: 👍 @gyliu513 is this something that can be done in a follow up or is this something that's hard to change later?

mali-chainzee:
> @mali-chainzee it seems you reintroduced computing the deps. To avoid committing them you can use some .gitignore settings (is there a standard .gitignore for Go applications?)

@carbonin mentioned using one commit for deps and one for controllers and scaffolding. I can just .gitignore on the vendor folder you don't want to see the deps

carbonin:
IIUC we would actually need `vendor` committed, right? I just want to be able to review the other changes so I suggested we put the non-dependency code in separate commits.

mali-chainzee:
> IIUC we would actually need `vendor` committed, right? I just want to be able to review the other changes so I suggested we put the non-dependency code in separate commits.

yeah we need it for build

manageiq-operator/Gopkg.toml
gyliu513: @Fryguy it is simple to change, we can merge it first and use go mod in follow up PRs :)

Fryguy:
Can the deps be pulled down at build time?

gyliu513:
@Fryguy if you are asking go mod, the answer is yes, here is an example for our project https://github.com/IBM/operand-deployment-lifecycle-manager/blob/cd53d395ad43ae0450be4775a72432511bbb2ab5/common/Makefile.common.mk#L63-L65

Fryguy:
Ok I most be thinking of go mod then, which is preferable anyway

carbonin:
If we can avoid adding the dependencies to the code-base shouldn't we make the change to a module before this is merged so we can avoid committing the vendor directory into the tree at any point as @Fryguy suggested initially?

mali-chainzee:
> Can the deps be pulled down at build time?
 
`operator-sdk build` doesn't do that for you. I'm using `dep` as a the dependency manager, which is recommended as the dependency graph need to be resolved first. One solution is to simply run `dep ensure` to grab packages before running `operator-sdk build`

manageiq-operator/Gopkg.toml
Fryguy: It seems that the previous way enforces committing all of the dependencies which affects the git history.  Given that it's a *lot* of code, I'd prefer we use go mod from the start.

Fryguy:
@mali-chainzee That would be preferable for now, and would allow us to switch to go mod afterwards.  Can you

- Remove the commit that contains the dependencies
- Change the readme and add `dep ensure` to the instructions.

manageiq-operator/build/bin/user_setup
```diff
@@ -0,0 +1,13 @@
...
+# ensure $HOME exists and is accessible by group 0 (we don't know what the runtime UID will be)
+mkdir -p ${HOME}
+chown ${USER_UID}:0 ${HOME}
+chmod ug+rwx ${HOME}
+
+# runtime user will need to be able to self-insert in /etc/passwd
+chmod g+rw /etc/passwd
+
+# no need for this script to remain in the image after running
+rm $0
```
carbonin: This is weird. Can you do this removal in the Dockerfile rather than in the script itself?

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: manageiq.example.com/v1alpha1
+kind: Manageiq
+metadata:
+  name: miq
+spec:
+  # Add fields here
+  appName:  "manageiq"
+  applicationAdminPassword: "smartvm" 
+  applicationDomain: "miqproject.apps-crc.testing"
```
carbonin: Is this an example file? I would expect that this would always need to be changed so does it make sense to provide a default?

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_crd.yaml
```diff
@@ -0,0 +1,38 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: manageiqs.manageiq.example.com
+spec:
+  group: manageiq.example.com
```
carbonin: @Fryguy do we want this to be `manageiq.manageiq.org` as we actually own that domain?

manageiq-operator/deploy/operator.yaml
```diff
@@ -0,0 +1,33 @@
...
+  template:
+    metadata:
+      labels:
+        name: manageiq-operator
+    spec:
+      serviceAccountName: manageiq-operator
+      containers:
+        - name: manageiq-operator
+          # Replace this with the built image name
+          image: docker.io/chainzeeli/manageiq-operator:v0.0.1
```
carbonin: I think it's safe to just change this to `docker.io/manageiq/manageiq-operator` now.

manageiq-operator/deploy/operator.yaml
```diff
@@ -0,0 +1,33 @@
...
+          command:
+          - manageiq-operator
+          imagePullPolicy: Always
+          env:
+            - name: WATCH_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
```
carbonin: If this is always going to be the current namespace, can we use the namespace file in the pod instead of passing additional env vars?

manageiq-operator/deploy/role.yaml
```diff
@@ -0,0 +1,61 @@
...
+  - events
+  - configmaps
+  - secrets
+  verbs:
+  - '*'
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  - daemonsets
```
carbonin: How strict should we be with this role? I don't think we manage `daemonsets` or `statefulsets` anywhere.

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -0,0 +1,116 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
carbonin: I think we can remove this comment.

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -0,0 +1,116 @@
...
+
+// ManageiqSpec defines the desired state of Manageiq
+// +k8s:openapi-gen=true
+type ManageiqSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
```
carbonin: Is this comment duplicated? We can remove one of the two.

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -0,0 +1,116 @@
...
+	// Application region number
+	DatabaseRegion string `json:"databaseRegion"`
+	// Containerized database volume size
+	DatabaseVolumeCapacity string `json:"databaseVolumeCapacity"`
+
+	HttpdCPUReq string `json:"httpdCPUReq"`
+	HttpdImageName string `json:"httpdImageName"`
+	HttpdImageTag string `json:"httpdImageTag"`
+	HttpdMemLimit string `json:"httpdMemLimit"`
+	HttpdMemReq string `json:"httpdMemReq"`
```
carbonin: @Fryguy do we want to take this opportunity to not abbreviate unnecessarily?

I'm thinking `...MemoryRequest` is not too much, right?

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -0,0 +1,116 @@
...
+	OrchestratorCPUReq string `json:"orchestratorCPUReq"`
+	OrchestratorImageName string `json:"orchestratorImageName"`
+	OrchestratorImageNamespace string `json:"orchestratorImageNamespace"`
+	OrchestratorImageTag string `json:"orchestratorImageTag"`
+	OrchestratorMemLimit string `json:"orchestratorMemLimit"`
+	OrchestratorMemReq string `json:"orchestratorMemReq"`
+
+	// database pod deployment information
+	PostgresqlCPUReq string `json:"postgresqlCPUReq"`
+	PostgresqlImgName string `json:"postgresqlImgName"`
```
carbonin: Same here. `Image` rather than `Img`

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -0,0 +1,116 @@
...
+// ManageiqStatus defines the observed state of Manageiq
+// +k8s:openapi-gen=true
+type ManageiqStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
+	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
+	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+	HttpdISSetup bool `json:"httpdIssetup"`
+	MemcachedISSetup bool `json:"memcachedIssetup"`
+	PostgresqlISSetup bool`json:"postgresqlIssetup"`
+	OrchestratorISSetup bool `json:"orchestratorIsrunning"`
```
carbonin: Why `running` here and `setup` everwhere else?
Also the capitalization here is in consistent.

I would think this would be the format:
```go
OrchestratorIsSetup bool `json:"orchestratorIsSetup"`
```

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_crd.yaml
```diff
@@ -0,0 +1,38 @@
...
+  names:
+    kind: Manageiq
```
Fryguy: Is it possible to name the `kind` as `ManageIQ`, and subsequently `listKind` to `ManageIQList`?

I'm wondering if this should be the name at all or something else like `manageiq-orchestrator` or something.

@carbonin Thoughts?

Fryguy: Also plural `manageiqs` reads funny to me, but I don't know what else it would be.

manageiq-operator/pkg/controller/controller.go
```diff
@@ -0,0 +1,18 @@
...
+
+// AddToManager adds all Controllers to the Manager
+func AddToManager(m manager.Manager) error {
+	for _, f := range AddToManagerFuncs {
+		if err := f(m); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
carbonin: Missing newline

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	miqtool "github.com/manageiq-operator/pkg/helpers/miq-components"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+var log = logf.Log.WithName("controller_manageiq")
+var currentAppName string
+/**
+* USER ACTION REQUIRED: This is a scaffold file intended for the user to modify with their own Controller
+* business logic.  Delete these comments after modifying this file.*
+ */
```
carbonin: I think we should "Delete these comments after modifying this file" :trollface:

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	}
+
+	return nil
+}
+
+// blank assignment to verify that ReconcileManageiq implements reconcile.Reconciler
+var _ reconcile.Reconciler = &ReconcileManageiq{}
+
+// ReconcileManageiq reconciles a Manageiq object
+type ReconcileManageiq struct {
```
carbonin: I think a better name would be `ManageiqReconciler`

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// blank assignment to verify that ReconcileManageiq implements reconcile.Reconciler
+var _ reconcile.Reconciler = &ReconcileManageiq{}
```
carbonin: Is this really the best way to do this? It just seems like something that should be covered by a keyword or something.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	// This client, initialized using mgr.Client() above, is a split client
+	// that reads objects from the cache and writes to the apiserver
+	client client.Client
+	scheme *runtime.Scheme
+}
+
+// Reconcile reads that state of the cluster for a Manageiq object and makes changes based on the state read
+// and what is in the Manageiq.Spec
+// TODO(user): Modify this Reconcile function to implement your Controller logic.  This example creates
+// a Pod as an example
```
carbonin: This TODO can be removed.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+func (r *ReconcileManageiq) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling Manageiq")
+
+	// Fetch the Manageiq instance
+	miqInstance := &miqv1alpha1.Manageiq{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, miqInstance)
+
+	if errors.IsNotFound(err) {
+		err = CleanUpOrchestratedDeployments(miqInstance, r)
```
carbonin: Is `err` unused in this context? Can we remove this assignment or should we be handling this error?

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	currentAppName = miqInstance.Spec.AppName
+
+	if miqInstance.Status.NetworkISSetup==false {
```
carbonin: Missing spaces around `==`.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		err = GenerateRbacResources(miqInstance, r)
+		if err != nil { return reconcile.Result{}, err }
```
carbonin: Would this be better as one line?
```go
if err := GenerateRbacResources(miqInstance, r); err != nil { return reconcile.Result{}, err }
```

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		err = GenerateSecrets(miqInstance, r)
+		if err != nil { return reconcile.Result{}, err }
+		miqInstance.Status.NetworkISSetup=true
```
carbonin: Spaces around `=`

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
carbonin: Also I think these would be better as `if !miqInstance.Status.NetworkISSetup` rather than comparing to a `bool`

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+
+	HttpdDeployment := miqtool.NewHttpdDeployment(cr)
+	err = Createk8sResIfNotExist(cr, HttpdDeployment, &appsv1.Deployment{}, r)
+	if err != nil { return err }
+
+	err = Createk8sResIfNotExist(cr, HttpdIngress, &extenv1beta1.Ingress{}, r)
+	if err != nil { return err }
+
+
+	return err
```
carbonin: Couldn't this be `return nil`?

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		return nil
+	} else if err != nil { return err }
+	return nil
+}	
+
+func CleanUpOrchestratedDeployments( cr *miqv1alpha1.Manageiq, r *ReconcileManageiq) error {
+	reqLogger := log.WithValues("task: ", "Clean up resources")
+	reqLogger.Info("Cleaning up orchestrated resources")
+	gracePeriod := int64(0);
+    deleteOpFunc := client.GracePeriodSeconds(gracePeriod)
```
carbonin: Spacing

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	reqLogger := log.WithValues("task: ", "create resource")
+	if err := controllerutil.SetControllerReference(cr, res, r.scheme); err != nil { return err }
+	resClient := res.(runtime.Object)
+	resTypeClient := restype.(runtime.Object)
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: res.GetName(), Namespace: res.GetNamespace()}, resTypeClient.(runtime.Object) )
+	if err != nil && errors.IsNotFound(err) {
+		reqLogger.Info("Creating ", "Resource.Namespace", res.GetNamespace(), "Resource.Name", res.GetName())
+		if err = r.client.Create(context.TODO(), resClient); err != nil { return err }
+		return nil
+	} else if err != nil { return err }
```
carbonin: Can this be a multi-line block? I think that would read better.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+		if err != nil { return reconcile.Result{}, err }
+		miqInstance.Status.OrchestratorISSetup = true
+	}
+
+	
+	return reconcile.Result{}, nil
+}
+
+func GenerateHttpdResources(cr *miqv1alpha1.Manageiq, r *ReconcileManageiq) error {
+	HttpdIngress := miqtool.NewIngress(cr)
```
carbonin: Why are these variables starting with an upper case?

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
carbonin: Same for all the functions in this file.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
carbonin: Same for all the functions in this file.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -0,0 +1,317 @@
...
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: res.GetName(), Namespace: res.GetNamespace()}, resTypeClient.(runtime.Object) )
+	if err != nil && errors.IsNotFound(err) {
+		reqLogger.Info("Creating ", "Resource.Namespace", res.GetNamespace(), "Resource.Name", res.GetName())
+		if err = r.client.Create(context.TODO(), resClient); err != nil { return err }
+		return nil
+	} else if err != nil { return err }
+	return nil
+}	
+
+func CleanUpOrchestratedDeployments( cr *miqv1alpha1.Manageiq, r *ReconcileManageiq) error {
```
carbonin: This function name seems strange. Maybe `RemoveManageiqResources` to contrast with the `Generate...Resources` functions?

manageiq-operator/pkg/helpers/httpd_conf/application.conf
carbonin: I think this is probably the one thing I'm okay having a copy of for a while. I really don't see this changing in the time it takes us to move from templates to the operator.

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -0,0 +1,443 @@
...
+		"app": cr.Spec.AppName,   
+	}
+	authconf :=  readContentFromFile("pkg/resources/httpd_conf/authentication.conf")
+	data := map[string]string{
+		"auth-type": "internal",
+		"auth-kerberos-realms": "undefined",
+		"auth-oidc-provider-metadata-url": "undefined",
+    	"auth-oidc-client-id": "undefined",
+    	"auth-oidc-client-secret": "undefined",
+    	"auth-configuration.conf": string(authconf),
```
carbonin: Spacing

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -0,0 +1,443 @@
...
+					
+				},
+			},
+		},
+	}
+}
+
+
+
+
```
carbonin: Woah, bonus whitespace.

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -0,0 +1,443 @@
...
+func TLSSecret(cr *miqv1alpha1.Manageiq) *corev1.Secret {
+	labels := map[string]string{
+		"app": cr.Spec.AppName,   
+	}
+
+	crt, key := tlstools.GenerateCrt("server")
```
carbonin: How does someone provide their own cert/key if they have them?

manageiq-operator/pkg/helpers/randstring/randstring.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+
+
+
+
+
+
+
+
```
carbonin: Lots of bonus whitespace.

Fryguy:
> Also, as I'm reading this, I realized that we don't really have a solution for configuring external auth with an operator.

Good catch @carbonin .  Let design for a followup PR.

manageiq-operator/pkg/helpers/miq-components/httpd.go
Fryguy: I think `gofmt` should catch a lot of this right?

manageiq-operator/build/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+
+ENV OPERATOR=/usr/local/bin/manageiq-operator \
+    USER_UID=1001 \
+    USER_NAME=manageiq-operator
+
+# install operator binary
+COPY build/_output/bin/manageiq-operator ${OPERATOR}
```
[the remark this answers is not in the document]
Fryguy: Agreed...let's remove the committed binaries and do the build inside of Docker.  Later we can make that a multi-stage build and do the build in an early stage to keep the container size small.

manageiq-operator/build/bin/user_setup
Fryguy: I thikn this might be boilerplate for operators...I think I've seen this elsewhere.

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
Fryguy: 👍 for not abbreviating...though we can probably do that in a followup

Fryguy:
@carbonin Doing an `operator-sdk new test-operator --repo github.com/example/test-operator` you can see the boilerplate for build and Dockerfile created.

However the code here seems to be from an old operator-sdk?  @mali-chainzee When I ran operator-sdk `v0.16.0`, I get some different output particularly in bin/entrypoint.  Which version did you run or how did you get that file?

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_crd.yaml
Fryguy: can it just be `manageiq.org` without the manageiq subdomain?

mali-chainzee:
> @mali-chainzee That would be preferable for now, and would allow us to switch to go mod afterwards. Can you
> 
> * Remove the commit that contains the dependencies
> * Change the readme and add `dep ensure` to the instructions.

I made to a new PR with things cleaned up

mali-chainzee:
> @carbonin Doing an `operator-sdk new test-operator --repo github.com/example/test-operator` you can see the boilerplate for build and Dockerfile created.
> 
> However the code here seems to be from an old operator-sdk? @mali-chainzee When I ran operator-sdk `v0.16.0`, I get some different output particularly in bin/entrypoint. Which version did you run or how did you get that file?

I use operator-sdk v0.8.0 because it fits my k8s cluster version.

manageiq-operator/pkg/helpers/miq-components/httpd.go
mali-chainzee: they'd have to put them somewhere inside this project dir and operator reads from it. this requires code change. now it's tls secrets are randomly generated.

manageiq-operator/deploy/operator.yaml
mali-chainzee: which pod are you referring to?

carbonin: Every pod. ref (https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#directly-accessing-the-rest-api-1)
> the default namespace to be used for namespaced API operations is placed in a file at /var/run/secrets/kubernetes.io/serviceaccount/namespace in each container

carbonin: We use this for kube API operations in manageiq https://github.com/ManageIQ/manageiq/blob/1283535e41099a9c34cebdb27aac83c151fbb70d/lib/container_orchestrator/object_definition.rb#L88-L91

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
mali-chainzee: how about this ` applicationDomain: # e.g. "miqproject.apps-crc.testing"`

manageiq-operator/deploy/operator.yaml
mali-chainzee: In this case should I just make a container startup script to set the ENVs?

mali-chainzee: > We use this for kube API operations in manageiq https://github.com/ManageIQ/manageiq/blob/1283535e41099a9c34cebdb27aac83c151fbb70d/lib/container_orchestrator/object_definition.rb#L88-L91

In this case should I just make a container startup script to set the ENVs?

carbonin: Oh is a `WATCH_NAMESPACE` env var required for the operator? I figured you would just go read the file rather than using an env var at all.

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
carbonin: Sure? I would prefer to use an actual documentation hostname though. `manageiq.example.com` maybe?

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_crd.yaml
carbonin: Yeah, it's just the group so I think that would be fine. The name would be `manageiqs.manageiq.org` then I guess?

manageiq-operator/pkg/helpers/miq-components/httpd.go
carbonin: It would also require an operator image rebuild, right? That's ... not great. We can tackle this in a followup though.

manageiq-operator/deploy/operator.yaml
mali-chainzee: > Oh is a `WATCH_NAMESPACE` env var required for the operator? I figured you would just go read the file rather than using an env var at all.

yeah all the yaml files in this directory are generated by the SDK, so the ENVs are parts of the architecture

carbonin: Okay, this is fine for now then.

manageiq-operator/Gopkg.toml
Fryguy: Given that we can commit without the dependencies, I'm inclined to merge as is, and do go mod as a follow up.

manageiq-operator/build/Dockerfile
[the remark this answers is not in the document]
mali-chainzee: should I write a script for building inside docker then?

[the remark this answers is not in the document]
carbonin: How many commands does it take? Can you put it right in a `RUN` command?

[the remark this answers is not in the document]
mali-chainzee: just one `go build` command. yeah that works

manageiq-operator/deploy/operator.yaml
mali-chainzee: wait actually we can't do everything inside `Dockerfile`.  when building the binary from go source code, we need some system environment variables, which we need to resolve outside of Dockerfile, so we either need a wrapper script for `docker build` or pass all the args in one `docker build` command.

manageiq-operator/build/Dockerfile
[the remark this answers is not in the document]
mali-chainzee: wait actually some system environment variables need to be resolved outside of dockerfile, so we prob still need a wrapper for `docker build`.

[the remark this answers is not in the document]
carbonin: Can you be more specific? We make liberal use of `--build-arg` for our existing images in `bin/build` I figure this wouldn't be any different (and we should probably add the operator image to that script as well).

Fryguy:
So, summarizing where we are at, I think there are a number of things that can be done after merge, and only a couple before merge.

@mali-chainzee This PR is awesome and I'm really looking forward to getting it over the finish line.  Thanks for your efforts!

---

Before merge:
- Remove the binaries from the PR (yeah it will make the Docker build not work immediately, but we can fix that in a followup PR, IMO)

After merge in follow up PRs:
- Build in the Dockerfile directly: https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397314395
- Set up CI to at least verify build and gofmt: https://github.com/ManageIQ/manageiq-pods/pull/389#issuecomment-602876631
- Switch to go mod: https://github.com/ManageIQ/manageiq-pods/pull/389#issuecomment-603300089
- Perhaps update to latest boilerplate for operator-sdk 0.16: https://github.com/ManageIQ/manageiq-pods/pull/389#issuecomment-603447748
- Determine if "kind" can be "ManageIQ" as opposed to "Manageiq": https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397324661
- Change the boilerplate roles to something more restrictive: https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397319716
- Fully spelled out keys: https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397322227
- Certs available outside of the build: https://github.com/ManageIQ/manageiq-pods/pull/389#discussion_r397344335
- Do we need productizable kind names based on APP_NAME ?  Not even sure that's possible.
- Some other minor formatting cleanup and variable naming

carbonin:
Okay, I'm fine with this as a first pass. @Fryguy want to get this in and start working on the issues you listed out?

mali-chainzee:
btw, where do you plan to release the operator at the end? because if we release via registry.redhat.io, all the deps for build will eventually have to be committed to dist-git repo

simaishi:
Jansa backport details:

```
$ git log -1
commit 0897f982e982964c52d137be726b1a964e788b68
Date:   Mon Mar 30 11:46:41 2020 -0400

    Merge pull request #389 from mali-chainzee/operator

    added operator for manageiq-pods

    (cherry picked from commit 242a5675d8034da854a71c37b59e0aa9e35ece2c)
```

PR #393, opened 2020-03-26
Fix the backup and restore jobs

### LINKED ISSUE #285: openshift jobs cannot be re-run
Usually users schedule database backup. However, paragraph regarding database backup/restore doesn't mention that openshift jobs can be run only once and old job should be removed before next run.
So, I believe there should be apropriate note regarding openshift jobs.

### LINKED ISSUE #287: it would be good to give an example how to restore db from some certain snapshot
It is possible to create many db backups. Restore job contains BACKUP_VERSION key which contains latest by default. But it isn't clear what has to be put there. I suspect folder name. 
So, it would be useful to give an example how to restore db from some certain snapshot

carbonin:
This PR alters the backup and restore job pod specs to use the new postgres secret format and moves the backup and restore jobs to templates which allows us to generate a unique name for the jobs and specify a specific backup to restore using template parameters.

Requires https://github.com/ManageIQ/container-postgresql/pull/17

Fixes #285
Fixes #287


simaishi:
Jansa backport details:

```
$ git log -1
commit 7af4bef0b0277991800add16012d5618244413fc
Date:   Mon Mar 30 13:52:03 2020 -0400

    Merge pull request #393 from carbonin/fix_backup_and_restore

    Fix the backup and restore jobs

    (cherry picked from commit 6643f9951217680d282eae9ffc599c2084daca8b)
```

PR #401, opened 2020-03-30
Update the readme with new sdk version

carbonin:


Fryguy:
I think the `dep ensure` line can be removed now (or does it need to be changed to something else?)

carbonin:
Closing in favor of https://github.com/ManageIQ/manageiq-pods/pull/402

PR #402, opened 2020-03-31
Updated readme for ManageIQ operator

gyliu513:
FYI @Fryguy

### DIFF
--- a/manageiq-operator/README.md
+++ b/manageiq-operator/README.md
@@ -1,28 +1,8 @@
-ManageIQ Operator
-------
+# ManageIQ Operator
+
 This operator manages the lifecycle of ManageIQ application on a OCP4 cluster.
 
-Prerequisites
-------
-  * [go version v1.12+](https://golang.org/doc/install#install)
-  * [dep version v0.5.0+](https://golang.github.io/dep/docs/installation.html)
-  * [operator SDK CLI v0.8.0+](https://github.com/operator-framework/operator-sdk/blob/v0.8.x/doc/user/install-operator-sdk.md) 
-
-Run Operator
-------
-### Setup Path for Go
-
-On your $GOPATH, create path `$GOPATH/src/github.com`, then place this project folder on it.
-
-### Generate k8s API and CRD
-
-```bash
-$ operator-sdk add api \
-    --api-version=manageiq.org/v1alpha1 \
-    --kind=Manageiq
-$ operator-sdk generate k8s
-$ oc create -f deploy/crds/manageiq_v1alpha1_manageiq_crd.yaml
-```
+## Run Operator
 
 ### Option A: Run Locally(outside of cluster)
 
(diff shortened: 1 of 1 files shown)


manageiq-operator/README.md
```diff
@@ -34,68 +27,70 @@ $ operator-sdk up local --namespace=<yournamespace>
 
 ### Option B: Run Inside Cluster
 
-First we need to install required Go packages. As `dep` is the dependency manager for this project, run
+Please refer to *"4. Build and run the Operator"* in [this guide](https://docs.openshift.com/container-platform/4.1/applications/operator_sdk/osdk-getting-started.html) for remaining steps
```
carbonin: This is kind of confusing. Can we just copy in the steps we need?

It should really just be building and pushing the image then creating the resources we need, right?

manageiq-operator/README.md
```diff
@@ -1,28 +1,21 @@
-ManageIQ Operator
-------
-This operator manages the lifecycle of ManageIQ application on a OCP4 cluster.
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
```
carbonin: This doesn't really seem necessary for a doc of this size.

manageiq-operator/README.md
gyliu513: How about handle this in future PRs? I want to introduce a `Makefile` for the operator to facilitate the build process.

manageiq-operator/README.md
gyliu513: As I was using `doctoc` and it was generated automatically, how about keep this? As we may need to update the toc later.

If we remove this line, and then when we run `doctoc`, this line will be added back automatically.

manageiq-operator/README.md
Fryguy: Is the build anything more than `operator-sdk build <image>`?

manageiq-operator/README.md
Fryguy: I think @carbonin is suggesting to just not run doctoc.  Otherwise, this becomes a dependency for developers writing the README, and for a simple README it seems unnecessary.

manageiq-operator/README.md
carbonin: Also the section is partially duplicated below. So I think we really should either put all the documentation here or deduplicate.

Really all this needs to be is:

- `operator-sdk build <image>`
- Update `deploy/operator.yaml` to use the new image tag
- Push the image
- Create stuff in deploy dir

carbonin: Even our "Option A" is already documented in the linked section.

gyliu513: @Fryguy it will be just `oeprator-sdk build`, but I think the makefile should also include some other options, like build, test, linter, fmt etc.

manageiq-operator/README.md
gyliu513: Ah, I see, let me remove this

chessbyte:
Can we change the title of this PR?  Not sure what the CF refers to in ManageIQ project

gyliu513:
@chessbyte Done, thanks!

simaishi:
Jansa backport details:

```
$ git log -1
commit ef52682dbad94b2f2705d24cd484d334d8360107
Date:   Wed Apr 1 10:17:08 2020 -0400

    Merge pull request #402 from gyliu513/readme

    Updated readme for ManageIQ operator

    (cherry picked from commit e083a60ef90752fffed1f107b008db6472ad54bf)
```

PR #419, opened 2020-04-02
Add .swp files to .gitignore

carbonin:


Fryguy:
I'm fine with this, but in the future these should probably go into your personal, global, .gitignore.

PR #421, opened 2020-04-02
Fix random generation

carbonin:
Previously we were using home-grown random string generators and uuid generators.
This PR moves us to a common uuid module for generating a GUID and uses more more secure crypto libraries to implement the same key generation algorithm as [manageiq-password](https://github.com/ManageIQ/manageiq-password/blob/f85e888fe60433390d0a1d5145889edc67437fdf/lib/manageiq/password.rb#L210-L213)

Additionally there is guidance to never use `math/rand` to generate keys: https://github.com/golang/go/wiki/CodeReviewComments#crypto-rand


simaishi:
Jansa backport details:

```
$ git log -1
commit 3f8091f9dabb31be2225c8ff2238be9f6d028da9
Date:   Thu Apr 2 17:27:53 2020 -0400

    Merge pull request #421 from carbonin/fix_random_generation

    Fix random generation

    (cherry picked from commit 1ebf278f5adb1113a3cd15d127b073a8d50198cd)
```

PR #424, opened 2020-04-06
Update pg secret handling

carbonin:
This PR removes the need for a user to specify the database secrets as a part of the CR by changing the CR to expect a secret name rather than the individual fields.

Now, the user can create a secret and specify that name in the CR in order to specify individual values. Alternatively, the operator will create a secret (the name defaults to "postgresql-secrets") with default/generated values.

Additionally the first commit does some refactoring to move the PG secret creation with the rest of the PG object handling.


manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -9,6 +12,46 @@ import (
...
+}
+
+func generateDatabasePassword() string {
+	buf := make([]byte, 8)
+	_, err := rand.Read(buf)
+	if err != nil {
+		panic(err) // out of randomness, should never happen
+	}
+	return hex.EncodeToString(buf)
+}
```
Fryguy, line 53: This feels like an almost copy/paste of the other secrets method [here](https://github.com/ManageIQ/manageiq-pods/blob/1ebf278f5adb1113a3cd15d127b073a8d50198cd/manageiq-operator/pkg/helpers/miq-components/app-secret.go#L32-L43).  I'm wondering if it's better to have a utility module where the baseline logic for generating passwords lives.  Perhaps taking a parameter on the size of the buffer?

carbonin: Yeah, I was debating extracting this, but I couldn't find a good place for it just yet.

I could make a separate file in the `miqtools` package for the utility functions ... was also debating turning the `tlstools` package into a more general secret tools thing ...

I ended up punting and just opening this as is.

carbonin:
Okay, extracted the random generator functions. @Fryguy

simaishi:
Jansa backport details:

```
$ git log -1
commit 7705a96e726db31a472faac4c04cf4efe78601a8
Date:   Tue Apr 7 12:30:40 2020 -0400

    Merge pull request #424 from carbonin/update_pg_secret_handling

    Update pg secret handling

    (cherry picked from commit d3a99678961f5ea9fcdc7e1bb559e9866ae0bc13)
```

PR #426, opened 2020-04-07
Use StdEncoding rather than URLEncoding for creating the encryption key

carbonin:
With the URLEncoding, we would get an error from our ruby code
complaining that the key was not 32 bytes. It seems the difference
in encoding was causing the issue.

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/generators.go
+++ b/manageiq-operator/pkg/helpers/miq-components/generators.go
@@ -17,12 +17,8 @@ func randomBytes(n int) []byte {
 }
 
 func generateEncryptionKey() string {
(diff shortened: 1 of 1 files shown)


simaishi:
Jansa backport details:

```
$ git log -1
commit f7cf270b8df24d7c0e006adf33d71c96127134e6
Date:   Tue Apr 7 16:36:05 2020 -0400

    Merge pull request #426 from carbonin/fix_encryption_key_generator

    Use StdEncoding rather than URLEncoding for creating the encryption key

    (cherry picked from commit 446a1102f9b4f0d8cf36db5b2b58f6dee05785e0)
```

PR #429, opened 2020-04-07
Watch on all resource types in case of accidental deletion of resources

mali-chainzee:
Updated controller to watch all resources created by manageiq instance to make sure they have the same life cycle as manageiq instance.

### DIFF
--- a/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
+++ b/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
@@ -51,15 +51,34 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		return err
 	}
 
+	ownerHandler := &handler.EnqueueRequestForOwner{IsController: true, OwnerType: &miqv1alpha1.Manageiq{}}
 	// Watch for changes to secondary resource Deployments and requeue the owner Manageiq
-	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, &handler.EnqueueRequestForOwner{
-		IsController: true,
-		OwnerType:    &miqv1alpha1.Manageiq{},
-	})
+	err = c.Watch(&source.Kind{Type: &appsv1.Deployment{}}, ownerHandler)
+
 	if err != nil {
 		return err
 	}
 
+	err = c.Watch(&source.Kind{Type: &corev1.Service{}}, ownerHandler)
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Kind{Type: &corev1.ConfigMap{}}, ownerHandler)
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Kind{Type: &corev1.Secret{}}, ownerHandler)
+	if err != nil {
+		return err
+	}
+
+	err = c.Watch(&source.Kind{Type: &corev1.PersistentVolumeClaim{}}, ownerHandler)
+	if err != nil {
+		return err
+	}
+	
 	return nil
 }
 
--- a/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
+++ b/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
@@ -78,7 +78,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	if err != nil {
 		return err
 	}
-	
+
 	return nil
 }
 


.gitignore
```diff
@@ -3,5 +3,4 @@ build.log
 manageiq-operator/build/_output/*
 parameters
 manageiq-operator/vendor/*
-
```
carbonin, line 6: Why change this in this PR?

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
```diff
@@ -1,39 +1,48 @@
-apiVersion: manageiq.org/v1alpha1
-kind: Manageiq
+apiVersion:  manageiq.org/v1alpha1
```
carbonin, line 1: Please remove extra spacing.

manageiq-operator/deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
```diff
@@ -46,3 +55,4 @@ spec:
   kafkaSecret: kafka-secrets
   kafkaVolumeCapacity: 1Gi
   zookeeperVolumeCapacity: 1Gi
+
```
carbonin, line 58: Bonus whitespace

manageiq-operator/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/manageiq/manageiq-pods/manageiq-operator
 go 1.14
 
 require (
+	github.com/NYTimes/gziphandler v1.0.1 // indirect
```
carbonin, line 6: This doesn't look used. Remove it.

manageiq-operator/go.mod
```diff
@@ -29,3 +30,5 @@ replace (
 )
 
 replace github.com/operator-framework/operator-sdk => github.com/operator-framework/operator-sdk v0.9.0
+
+
```
carbonin, line 34: Bonus whitespace.

manageiq-operator/go.sum
```diff
@@ -236,6 +236,7 @@ github.com/mailru/easyjson v0.0.0-20190403194419-1ea4449da983 h1:wL11wNW7dhKIcRC
 github.com/mailru/easyjson v0.0.0-20190403194419-1ea4449da983/go.mod h1:C1wdFJiN94OJF2b5HbByQZoLdCWB1Yqtg26g4irojpc=
 github.com/manageiq/manageiq-pods v0.0.0-20200330194538-eaa4f1ac930e h1:ox/otVzd8IhFYXzxj8O8Mh/5TxJJjrtw3k6p1dLD65g=
 github.com/manageiq/manageiq-pods v0.0.0-20200330194538-eaa4f1ac930e/go.mod h1:UlzmBz9L9IGReoZJGlvsWxf/g5Td9jZCfu9Y7Z7943I=
+github.com/manageiq/manageiq-pods v0.0.0-20200407163040-d3a99678961f h1:MAOt9rhzLkf8PpnwxM3KEULyIRahZGl2mV0XPR05xOM=
```
carbonin, line 239: Not sure why this was added.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -117,6 +138,12 @@ func (r *ReconcileManageiq) Reconcile(request reconcile.Request) (reconcile.Resu
 }
 
 func (r *ReconcileManageiq) generateHttpdResources(cr *miqv1alpha1.Manageiq) error {
+
+	httpdDeployment := miqtool.NewHttpdDeployment(cr)
+	if err := r.createk8sResIfNotExist(cr, httpdDeployment, &appsv1.Deployment{}); err != nil {
+		return err
+	}
```
carbonin, line 145: Why was this moved? It seems like the config maps should be created before the deployment.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -180,6 +202,12 @@ func (r *ReconcileManageiq) generateMemcachedResources(cr *miqv1alpha1.Manageiq)
 }
 
 func (r *ReconcileManageiq) generatePostgresqlResources(cr *miqv1alpha1.Manageiq) error {
+
+	postgresqlDeployment := miqtool.NewPostgresqlDeployment(cr)
+	if err := r.createk8sResIfNotExist(cr, postgresqlDeployment, &appsv1.Deployment{}); err != nil {
+		return err
+	}
+
```
carbonin, line 210: Same here, I think the PVC, secret, and config maps should be created first.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -200,10 +228,7 @@ func (r *ReconcileManageiq) generatePostgresqlResources(cr *miqv1alpha1.Manageiq
 		return err
 	}
 
-	postgresqlDeployment := miqtool.NewPostgresqlDeployment(cr)
-	if err := r.createk8sResIfNotExist(cr, postgresqlDeployment, &appsv1.Deployment{}); err != nil {
-		return err
-	}
+
```
carbonin, line 231: Bonus space not needed.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -299,9 +324,9 @@ func (r *ReconcileManageiq) generateRbacResources(cr *miqv1alpha1.Manageiq) erro
 	return nil
 }
 
-func (r *ReconcileManageiq) createk8sResIfNotExist(cr *miqv1alpha1.Manageiq, res, restype metav1.Object) error {
+func (r *ReconcileManageiq) createk8sResIfNotExist(owner, res, restype metav1.Object) error {
```
carbonin, line 327: Why change this? I still only see "Manageiq" structs being passed as this parameter?

carbonin:
@mali-chainzee I think most of these comments and the test failures will be fixed by running `go fmt`.

carbonin:
Also @mali-chainzee what behavior will this change exactly?

Right now all the objects we create are automatically removed when the `manageiq` CR is removed. What does this add to the "lifecycle"?

As far as I understand it this would allow the controller to re-create objects if they were removed, correct? I think some more work is needed to enforce state of an object that already exists, right? (Not that I'm asking you do make those changes in this PR).

mali-chainzee:
> Also @mali-chainzee what behavior will this change exactly?
> 
> Right now all the objects we create are automatically removed when the `manageiq` CR is removed. What does this add to the "lifecycle"?
> 
> As far as I understand it this would allow the controller to re-create objects if they were removed, correct? 
Exactly
>I think some more work is needed to enforce state of an object that already exists, right? (Not that I'm asking you do make those changes in this PR).
right now I'm just trying to make sure all resources are in place during run time so the app won't break. What else do you think is needed to enforce state of objects?

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -50,8 +50,49 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+	err = c.Watch(&source.Kind{Type: &corev1.ServiceAccount{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &miqv1alpha1.Manageiq{},
+	})
+	if err != nil {
+		return err
+	}
+	err = c.Watch(&source.Kind{Type: &rbacv1.RoleBinding{}}, &handler.EnqueueRequestForOwner{
+		IsController: true,
+		OwnerType:    &miqv1alpha1.Manageiq{},
```
carbonin, line 91: Can we deduplicate this handler reference? Maybe have one
```go
ownerHandler := &handler.EnqueueRequestForOwner{IsController: true, OwnerType:    &miqv1alpha1.Manageiq{}}
```

At the start and call `Watch` with the same one each time?

carbonin:
I think that if someone updates, for example, the memory limit for one of the deployments in the CR we should update that deployment.

Currently I think `createk8sResIfNotExist` only creates the object rather than updating it when it changes. I would like to replace `createk8sResIfNotExist` with [`controllerutil.CreateOrUpdate`](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.1.12/pkg/controller/controllerutil/controllerutil.go#L123).

I don't think it will be quite so simple as we'd have to alter the miqtools functions to be used as the mutate function rather than just creating a new instance of the requested object.

carbonin:
@mali-chainzee I'd prefer a separate PR for the httpd config fix. Additionally please reference an issue  with `Fixed #...` in the commit message to automatically close the issue when the commit is merged.

simaishi:
Jansa backport details:

```
$ git log -1
commit e17aa1acbcbfece63404bd2c1f065c59bbbf9bc0
Date:   Tue Apr 14 17:14:30 2020 -0400

    Merge pull request #429 from mali-chainzee/operator

    Watch on all resource types in case of accidental deletion of resources

    (cherry picked from commit 424a4d83423ebf00ed03e9d083dd2be7471bea77)
```

PR #441, opened 2020-04-14
[Operator] Update existing objects in the reconcile loop

### LINKED ISSUE #438: Update existing objects in the reconcile loop
Currently the only operation our reconcile method carries out is to create an object if it doesn't already exist. With https://github.com/ManageIQ/manageiq-pods/pull/429 we ensure that all the different types of objects that we require are present we don't verify (or fix) their contents.

This can manifest in many different ways, but some examples would be:
  - If a user updates the memory or cpu requirements in the CR the deployments would not be updated
  - If an object is edited, it won't be corrected

I believe we should be using [controllerutil.CreateOrUpdate](https://github.com/kubernetes-sigs/controller-runtime/blob/v0.1.12/pkg/controller/controllerutil/controllerutil.go#L123) for all of our objects to ensure they have the desired values.

mali-chainzee:

Fixes #438


carbonin:
@mali-chainzee why copy the content of `controllerutil.CreateOrUpdate` rather than using it directly?

carbonin:
Frankly you could just pass it a mutate function that just returns nil and the object we want created.
That should do the trick rather than copying it over.

mali-chainzee:
I used `controllerutil.CreateOrUpdate` first but updates didn't work, so I read the function code and found out that it requires you to change the resource object inside `type func mutateFn(existing runtime.object)`. (the object struct we pass in would be used to fetch object from api server, therefore overwritten with the old resource).

mali-chainzee:
`type func mutateFn(existing runtime.object)` lacks the flexibility as we need to pass in `manageiq` instance

carbonin:
@mali-chainzee I don't agree. Functions are closures so you can reference the cr in the callback when defining the function and it will work without parameter passing. See https://tour.golang.org/moretypes/25 for an interactive example.

carbonin:
Using that example as a reference you can turn the current functions that return a new object into ones that return mutate functions that have the cr bound inside the returned function closure like the `adder` function.

mali-chainzee:
okay, I'll do that.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -137,47 +135,92 @@ func (r *ReconcileManageIQ) Reconcile(request reconcile.Request) (reconcile.Resu
 
 func (r *ReconcileManageIQ) generateHttpdResources(cr *miqv1alpha1.ManageIQ) error {
 	httpdConfigMap := miqtool.NewHttpdConfigMap(cr)
-	if err := r.createk8sResIfNotExist(cr, httpdConfigMap, &corev1.ConfigMap{}); err != nil {
+
+	if _, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, httpdConfigMap, func() error {
```
carbonin, line 139: I was thinking the callback function could be returned by the `miqtool` functions, but if you would rather keep it here, can you move it out of the `CreateOrUpdate` call? I think it would make all of this easier to read.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -310,16 +445,25 @@ func (r *ReconcileManageIQ) generateSecrets(cr *miqv1alpha1.ManageIQ) error {
...
+		return nil
+	}); err != nil {
 		return err
 	}
 
 	return nil
+
 }
 
 func (r *ReconcileManageIQ) createk8sResIfNotExist(cr *miqv1alpha1.ManageIQ, res, restype metav1.Object) error {
```
carbonin, line 469: This should be removed now, right?

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
mali-chainzee: let me just put them in `miqtool` functions then

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -104,17 +112,23 @@ func NewHttpdConfigMap(cr *miqv1alpha1.ManageIQ) *corev1.ConfigMap {
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "httpd-configs",
 			Namespace: cr.ObjectMeta.Namespace,
 			Labels:    labels,
 		},
 		Data: data,
 	}
+
+	mutateF := func() error {
+		ConfigMap.Labels = labels
```
carbonin, line 125: I'm just going to comment here because it's a pretty simple case, but this applies generally.

I think that we should be setting everything that is important to us in this mutate function, so we should also be setting data.

Because of that we should deduplicate this logic and only set labels and data in the mutate function as it's called for both create and update. The only things we should need for the initial object are `Name` and `Namespace`.

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -104,17 +112,23 @@ func NewHttpdConfigMap(cr *miqv1alpha1.ManageIQ) *corev1.ConfigMap {
 		"external-auth-openid-connect-remote-user-conf": string(externalAuthOpenidConnectRemoteUserConf),
 	}
 
-	return &corev1.ConfigMap{
+	ConfigMap := &corev1.ConfigMap{
```
carbonin, line 115: Use lower case for non-exported names. So `configMap` or `cm` would work here.

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -66,7 +74,7 @@ func readContentFromFile(filename string) string {
 	return contents
 }
 
-func NewHttpdConfigMap(cr *miqv1alpha1.ManageIQ) *corev1.ConfigMap {
+func NewHttpdConfigMap(cr *miqv1alpha1.ManageIQ) (*corev1.ConfigMap, func() error) {
```
carbonin, line 77: I think `func() error` should probably be `controllerutil.MutateFn` as that type is defined and exported.

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -40,3 +44,107 @@ func addResourceReqs(memLimit string, memReq string, cpuReq string, c *corev1.Co
...
+		deployment.Spec.Template.ObjectMeta.Labels = podLabels
+		deployment.Spec.Template.Spec.Containers = append([]corev1.Container{}, container)
+		return nil
+	}
+
+	return mutateFn
+
+}
+
+func NilMutateFn() error {
```
carbonin, line 148: I would rather return this from the public service account functions than make this public. It makes them more similar to the rest of the functions that return the openshift objects.

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -40,3 +44,107 @@ func addResourceReqs(memLimit string, memReq string, cpuReq string, c *corev1.Co
...
+		roleBinding.RoleRef = roleRef
+		roleBinding.Subjects = subjects
+		return nil
+	}
+
+	return mutateFn
+
+}
+
+func deploymentMutateFn(deployment *appsv1.Deployment, labels map[string]string, repNum *int32, container corev1.Container, podTemplateSpec corev1.PodTemplateSpec, podLabels map[string]string) controllerutil.MutateFn {
```
carbonin, line 127: I feel like if we need 6 parameters here (and maybe another for the strategy type) there's probably too much variation here to make this a good generic function. We should probably just implement this for each of the deployments.

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -52,3 +56,96 @@ func addResourceReqs(memLimit, memReq, cpuLimit, cpuReq string, c *corev1.Contai
...
+		service.ObjectMeta.Labels = labels
+		service.Spec.Ports = ports
+		service.Spec.Selector = selector
+		return nil
+	}
+
+	return mutateFn
+}
+
+func secretMutateFn(secret *corev1.Secret, labels map[string]string, data map[string][]byte, stringData map[string]string) controllerutil.MutateFn {
```
carbonin, line 72: Same here, remove this if it isn't being used.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,81 +44,86 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
-				Requests: corev1.ResourceList{
-					"storage": storageReq,
-				},
-			},
 		},
 	}
 
-	if cr.Spec.StorageClassName != "" {
-		pvc.Spec.StorageClassName = &cr.Spec.StorageClassName
-	}
```
carbonin, line 84: This seems to be removing the storage class feature. Please handle this in the mutate function as well.

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -52,3 +56,96 @@ func addResourceReqs(memLimit, memReq, cpuLimit, cpuReq string, c *corev1.Contai
...
+	mutateFn := func() error {
+		role.Rules = rules
+		return nil
+	}
+
+	return mutateFn
+
+}
+
+func roleBindingMutateFn(roleBinding *rbacv1.RoleBinding, roleRef rbacv1.RoleRef, subjects []rbacv1.Subject) controllerutil.MutateFn {
```
carbonin, line 127: Same here, this isn't used in this patch so remove it for now.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -173,64 +182,68 @@ func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, erro
 			corev1.VolumeMount{Name: "miq-pg-configs", MountPath: "/opt/app-root/src/postgresql-cfg/"},
 		},
 	}
+}
 
-	err := addResourceReqs(cr.Spec.PostgresqlMemoryLimit, cr.Spec.PostgresqlMemoryRequest, cr.Spec.PostgresqlCpuLimit, cr.Spec.PostgresqlCpuRequest, &container)
-	if err != nil {
-		return nil, err
+func postgresqlPodSpec(cr *miqv1alpha1.ManageIQ) corev1.PodSpec {
```
carbonin, line 187: I don't see the `cr` used in this function, can we not pass it?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -173,64 +182,68 @@ func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, erro
...
+		},
 	}
 
-	deploymentLabels := map[string]string{
-		"app": cr.Spec.AppName,
+}
+
+func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, controllerutil.MutateFn, error) {
+
+	container := postgresqlContainer(cr)
```
carbonin, line 214: Why did you split this out from the pod spec function? If you keep this in the same function then I think this will be a lot simpler.

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -52,3 +56,96 @@ func addResourceReqs(memLimit, memReq, cpuLimit, cpuReq string, c *corev1.Contai
...
+		roleBinding.RoleRef = roleRef
+		roleBinding.Subjects = subjects
+		return nil
+	}
+
+	return mutateFn
+
+}
+
+func deploymentMutateFn(deployment *appsv1.Deployment, repNum *int32, container corev1.Container, podSpec corev1.PodSpec, strategy appsv1.DeploymentStrategyType) controllerutil.MutateFn {
```
carbonin, line 139: Generally I feel like this isn't worth splitting out into a separate function.  Passing 5 parameters to a function that only has 8 real lines doesn't feel like it's making anything better. I would rather see the mutate function defined for each deployment in their own methods.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -7,28 +7,32 @@ import (
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      postgresqlSecretName(cr),
 			Namespace: cr.ObjectMeta.Namespace,
 			Labels:    labels,
 		},
-		StringData: secret,
+		StringData: stringData,
 	}
+
+	return secret
```
carbonin, line 35: Why was any of this changed? It looks functionally identical to the original implementation?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,81 +44,86 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
 			Name:      "postgresql-configs",
 			Namespace: cr.ObjectMeta.Namespace,
-			Labels:    labels,
-		},
-		Data: map[string]string{
-			"01_miq_overrides.conf": postgresqlOverrideConf(),
 		},
 	}
+
+	mutateFn := configMapMutateFn(configMap, labels, data)
```
carbonin, line 63: Hm, I'm wondering if we don't actually want to enforce the contents here to allow for user modification... Probably okay for now, but worth considering.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -173,64 +182,77 @@ func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, erro
...
+	if err != nil {
+		return nil, nil, err
+	}
 	var repNum int32 = 1
 
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "postgresql",
 			Namespace: cr.ObjectMeta.Namespace,
-			Labels:    deploymentLabels,
```
carbonin, line 195: Why was this removed?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -173,64 +182,77 @@ func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, erro
...
-			Strategy: appsv1.DeploymentStrategy{
-				Type: "Recreate",
-			},
-			Replicas: &repNum,
 			Selector: &metav1.LabelSelector{
 				MatchLabels: podLabels,
 			},
 			Template: corev1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:   "postgresql",
```
carbonin, line 207: Why was this removed?

carbonin:
@mali-chainzee I'm not sure what happened here, but I am sure that you shouldn't have 114 commits in this branch. Can you rebase onto `upstream/master`?

PR #447, opened 2020-04-15
Do not use comments in the example CR.

gyliu513:

This is fixing the CR template in code, and make it works if I just follow the README guidance.

Fixes #

FYI @carbonin

### DIFF
--- a/manageiq-operator/README.md
+++ b/manageiq-operator/README.md
@@ -50,7 +50,7 @@ $ oc create -f deploy/operator.yaml
 ### Create the CR to deploy ManageIQ
 
 ```bash
-$ oc create -f deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
+$ oc create -f deploy/crds/manageiq.org_v1alpha1_manageiq_cr.yaml
 ```
 
 **ManageIQ Instance Example**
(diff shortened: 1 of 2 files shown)


carbonin:
I'm not really sure how useful this is as a user will almost always need to change something in this file.

Can we maybe just remove it and leave an example in the documentation? My concern is that people are going to deploy this then complain that it doesn't work in whatever environment they're in. Is there a compelling reason to keep it around?

gyliu513:
@carbonin if you take a look at https://github.com/ManageIQ/manageiq-pods/blob/master/manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq-operator.v0.0.1.clusterserviceversion.yaml#L5-L50 , you will see the CR was actually used in the CSV file and it has lot of `null` field, and this will be shown on OCP UI as the CR resource for the CSV file, so I was hoping we can add some real value so that customer will know what kind of values should set for each field.

carbonin:
Yeah, I see that now. But I also just opened up https://github.com/ManageIQ/manageiq-pods/pull/451 which will overwrite most of this anyway. Do you want to get this in first and I can rebase on top of these changes or do you think it makes more sense to just wait for #451 to land and close this for now?

gyliu513:
let me close this one and fix this in #451 , thanks @carbonin

PR #451, opened 2020-04-16
Use defaults for missing cr values

### LINKED ISSUE #443: Make more CR parameters optional
**Describe the solution you'd like**
More of the parameters should not have to be provided by the user. I would like things like images and tags to have reasonable defaults. Additionally we can make it easier for developers to deploy in environments with limited resources by only using resource checks if they were provided by the user.

The last bit for this issue would be to deal with the encryptionKey parameter which should never be provided in the CR, but would probably need to be dealt with similarly to the other secrets or just not exposed at all.

carbonin:

This PR allows some sensible defaults to be used for nearly all of the parameters in the manageiq CR. By doing this it also allows manageiq to be deployed more easily in resource constrained environments like Code Ready Containers. Specifically the defaults for the resource limits and requirements are to not set any limits or requirements. If any of the values are set they will be properly reflected in the deployment.

This also makes manageiq much easier to deploy. A valid CR is now just:
```yaml
apiVersion: manageiq.org/v1alpha1
kind: ManageIQ
metadata:
  name: miq
spec:
  applicationDomain: "miqproject.apps-crc.testing"
```

This prevents the user from needing to know about things like image location, and tags, while allowing them to continue to customize PV sizes and tuning parameters as needed.

I got very aggressive with this, but I would be okay with adding back some parameters like the PV sizes if we think those should be more visible.

Fixes #443

### DIFF
--- a/manageiq-operator/README.md
+++ b/manageiq-operator/README.md
@@ -50,7 +50,7 @@ $ oc create -f deploy/operator.yaml
 ### Create the CR to deploy ManageIQ
 
 ```bash
-$ oc create -f deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
+$ oc create -f deploy/crds/manageiq.org_v1alpha1_manageiq_cr.yaml
 ```
 
 **ManageIQ Instance Example**
--- a/manageiq-operator/README.md
+++ b/manageiq-operator/README.md
@@ -55,7 +55,9 @@ $ oc create -f deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
 
 **ManageIQ Instance Example**
 
-> Deployments' resource requests here are tailered to make them fit into a crc cluster, change them according to your cluster's resource capacity*
+> The domain here will work for a Code Ready Containers cluster. Change it to one that will work for your environment.
+
+> Additional parameters are available and documented in the Custom Resource Definition
 
 ```yaml
 apiVersion: manageiq.org/v1alpha1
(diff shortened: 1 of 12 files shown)


manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -7,62 +7,131 @@ import (
...
+	// Database region number (default: 0)
+	// +optional
 	DatabaseRegion string `json:"databaseRegion"`
-	// Containerized database volume size
+
+	// Database volume size (default: 15Gi)
+	// +optional
 	DatabaseVolumeCapacity string `json:"databaseVolumeCapacity"`
+
+	// Secret containing the database access information, content generated if not provided (default: postgresql-secrets)
```
gyliu513, line 30: My only concern is for the DatabaseSecret, KafakaSecret and TLSSecert, does it make sense to mention this in document that it is better to customize this in production environment?

Another question but not yours, I found that for those secrets, we only set a secret name but I did not see any logic for how the secret was generated, like the logic using username and password to generate etc, am I missing anything?

manageiq-operator/README.md
```diff
@@ -55,53 +55,15 @@ $ oc create -f deploy/crds/manageiq_v1alpha1_manageiq_cr.yaml
 
 **ManageIQ Instance Example**
 
-> Deployments' resource requests here are tailered to make them fit into a crc cluster, change them according to your cluster's resource capacity*
+> The domain here will work for a Code Ready Containers cluster. Change it to one that will work for your environment.
```
gyliu513, line 58: Please also change the CR name to `manageiq.org_v1alpha1_manageiq_cr.yaml` in L53

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
carbonin: > My only concern is for the DatabaseSecret, KafakaSecret and TLSSecert, does it make sense to mention this in document that it is better to customize this in production environment?

I think the only one that really *needs* to be customized in prod is the TLS one as the default would be self-signed. The others really only need to be changed if you're using a database or kafka outside of the project. Actually this reminded me that I need to open an issue to validate the secrets and to not deploy the database/kafka if an external hostname is provided in the secret.

> Another question but not yours, I found that for those secrets, we only set a secret name but I did not see any logic for how the secret was generated, like the logic using username and password to generate etc, am I missing anything?

The password and encryption key is generated [here](https://github.com/ManageIQ/manageiq-pods/blob/fe4c1445967f0abc47559baf920666aabedced74/manageiq-operator/pkg/helpers/miq-components/generators.go) and is called [here for PG](https://github.com/ManageIQ/manageiq-pods/blob/fe4c1445967f0abc47559baf920666aabedced74/manageiq-operator/pkg/helpers/miq-components/postgresql.go#L19). I think exposing the format of the secret would be good. Not sure where the best place to do that would be though. I don't think there's any reason to allow the user to specify a way to generate the password (if that's what you're asking).

gyliu513: Good to know, thanks @carbonin 

> I think the only one that really needs to be customized in prod is the TLS one as the default would be self-signed. 

Do you want to fix this in a follow up?

> I don't think there's any reason to allow the user to specify a way to generate the password (if that's what you're asking).

Yes, I did not mean specify a way to generate the password, but just mean if you're using a database or kafka outside of the project, I think you have answered my question.

carbonin: > Do you want to fix this in a follow up?

What kind of fix are you thinking about? Did you want to try to implement a way of generating a signed cert? I figured someone would bring their own cert and key files then just run `oc create secret tls tls-secret --cert=tls.crt --key=tls.key`

carbonin: Opened an issue to deal with documenting secret formats and validating user-provided secret content https://github.com/ManageIQ/manageiq-pods/issues/452

simaishi:
Jansa backport details:

```
$ git log -1
commit 40c4d59f4367198bfa736e487e47257092e60ac1
Author: Oleg Barenboim <chessbyte@gmail.com>
Date:   Thu Apr 16 10:05:13 2020 -0400

    Merge pull request #451 from carbonin/use_defaults_for_missing_cr_values

    Use defaults for missing cr values

    (cherry picked from commit ca7f33f82d77de006cd0bf6a02e660d1915da82c)
```

PR #454, opened 2020-04-16
Encode httpd and postgresql conf as static strings in functions

### LINKED ISSUE #453: Postgres and httpd config maps empty when deploying the operator as a container
**What steps did you take and what happened:**

After deploying the operator as a container and creating a ManageIQ CR all the deployments are created correctly but the httpd and postgres config maps are empty.

**What did you expect to happen:**

I expected them to be filled in correctly.

**Anything else you would like to add:**

This is essentially the same issue as https://github.com/ManageIQ/manageiq-pods/issues/427, but https://github.com/ManageIQ/manageiq-pods/pull/437 only addressed the issue when running the operator outside of the cluster.

After the operator binary is compiled it is copied into the container image at `/usr/local/bin` and the conf files are not copied into the image at all.

This means that the strategy of [reading the config from a file at runtime](https://github.com/ManageIQ/manageiq-pods/blob/ca7f33f82d77de006cd0bf6a02e660d1915da82c/manageiq-operator/pkg/helpers/miq-components/httpd.go#L76-L89) will not work.

I would suggest making the file content a constant string that is returned by a method so the content is compiled into the resulting binary.

carbonin:
Previously this only worked if you ran the operator-sdk locally
because the files were not copied into the image and the working
directory at runtime would have been different.
This will ensure that the string is available to us at runtime
in the container as well as running locally.

Fixes #453


simaishi:
Jansa backport details:

```
$ git log -1
commit 0aaf0b5ea07c48256faa900cefbfca1c7871b37e
Author: Oleg Barenboim <chessbyte@gmail.com>
Date:   Fri Apr 17 10:04:48 2020 -0400

    Merge pull request #454 from carbonin/fix_missing_config_map_contents

    Encode httpd and postgresql conf as static strings in functions

    (cherry picked from commit 5bd735e931754046e57e9ae776856eeead23c9b0)
```

PR #455, opened 2020-04-17
add app_name parameter to teardown script

bennett-white:
Setting an APP_NAME parameter in the teardown script that defaults to `manageiq`. Deployments using a non-default APP_NAME in the parameters file will be able to use teardown without editing it. 

Requesting that this change goes into jansa as well.

### DIFF
--- a/bin/teardown
+++ b/bin/teardown
@@ -1,15 +1,17 @@
 #!/bin/bash
 
-oc delete all -l app=manageiq
+APP_NAME=${APP_NAME:-manageiq}
+
+oc delete all -l app=$APP_NAME
 
 oc get pvc -o name |xargs oc delete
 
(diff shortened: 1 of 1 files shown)


carbonin:
I also think you need the role and role-binding too, right?

bennett-white:
Added this for role and role-bionding

simaishi:
Jansa backport details:

```
$ git log -1
commit b6891363d6b81912b8cab5e45f83375af4edd371
Date:   Fri Apr 17 09:54:21 2020 -0400

    Merge pull request #455 from bennett-white/teardown-parameter

    add app_name parameter to teardown script

    (cherry picked from commit b8b8d16e4af46ceb011dc955d71d2fa27dfb1d6f)
```

PR #456, opened 2020-04-17
Delay parameter

bennett-white:
If an environment is running slowly, the liveness probe will cause the orchestrator to restart before it's completely up and running. This parameter allows the initial delay on the orchestrator liveness probe to change without editing the template yaml. Typically, defaulting to 480 (8 minutes) should be plenty of time. 

Request to go into jansa as well


templates/app/orchestrator.yaml
```diff
@@ -30,7 +30,7 @@ objects:
               command:
               - pidof
               - MIQ Server
-            initialDelaySeconds: 480
+            initialDelaySeconds: ${{ORCHESTRATOR_INITIAL_DELAY}}
```
carbonin, line 33: The failing test is a format check so it's asking for quotes in this case, but does that make sense as this wasn't a string before? Also why the double curly braces?

bennett-white: The template parameters don't play nice with non-string types. On my system, the double brackets worked to interpret it as an int and not a string.

carbonin:
@bennett-white this should pass tests with the following patch:
```patch
diff --git a/templates/app/orchestrator.yaml b/templates/app/orchestrator.yaml
index 0459e98..17e6699 100644
--- a/templates/app/orchestrator.yaml
+++ b/templates/app/orchestrator.yaml
@@ -30,7 +30,7 @@ objects:
               command:
               - pidof
               - MIQ Server
-            initialDelaySeconds: ${{ORCHESTRATOR_INITIAL_DELAY}}
+            initialDelaySeconds: "${{ORCHESTRATOR_INITIAL_DELAY}}"
             timeoutSeconds: 3
           env:
           - name: ALLOW_INSECURE_SESSION
```

simaishi:
Jansa backport details:

```
$ git log -1
commit b823baae21fd3de14acdac41ab59ca30dd6227de
Date:   Wed Apr 22 15:02:36 2020 -0400

    Merge pull request #456 from bennett-white/delay-parameter

    Delay parameter

    (cherry picked from commit 11d357f5849db10228329c125da325e6ea948b9a)
```

Issue #442, opened 2020-04-14
Service account to scc mapping missing from operator


carbonin:
@miq-bot add_label bug

**What steps did you take and what happened:**
Deployed a manageiq installation to a new project using the operator.
The orchestrator and httpd pods failed with permission errors.

**What did you expect to happen:**
Everything should deploy cleanly.

**Anything else you would like to add:**
The problem here is that the steps to assign the anyuid scc to the service accounts created by the operator still need to be taken if they haven't been done before.

If this is possible to do in the operator itself, that would be the best solution. Otherwise we should doc this in the readme.

**Environment:**
- ManageIQ version: master
- Minikube/KIND/OCP version: 
```
$ crc version
crc version: 1.8.0+0a318dc
OpenShift version: 4.3.8 (embedded in binary)
```

carbonin:
The commands we're missing are:
```
oc adm policy add-scc-to-user anyuid system:serviceaccount:<your-namespace>:<app-name>-anyuid
oc adm policy add-scc-to-user anyuid system:serviceaccount:<your-namespace>:<app-name>-orchestrator
oc adm policy add-scc-to-user anyuid system:serviceaccount:<your-namespace>:<app-name>-httpd
```

carbonin:
So for this I think I'm going to just stop running as root for the app pods and configure an alternative httpd pod that doesn't implement the auth and so doesn't require privileged.

carbonin:
So for this I think I'm going to just stop running as root for the app pods and configure an alternative httpd pod that doesn't implement the auth and so doesn't require privileged.

gyliu513:
After adding liveness probe, the httpd pod keeps crashing.

```console
[root@gyliu-ocp-1-inf deploy]# oc describe pods httpd-55cb898d45-tldsl
Name:         httpd-55cb898d45-tldsl
Namespace:    default
Priority:     0
Node:         worker2.gyliu-ocp-1.os.fyre.ibm.com/10.16.8.109
Start Time:   Wed, 22 Apr 2020 04:17:59 -0700
Labels:       app=manageiq
              name=httpd
              pod-template-hash=55cb898d45
Annotations:  k8s.v1.cni.cncf.io/networks-status:
                [{
                    "name": "openshift-sdn",
                    "interface": "eth0",
                    "ips": [
                        "10.254.4.18"
                    ],
                    "dns": {},
                    "default-route": [
                        "10.254.4.1"
                    ]
                }]
Status:       Running
[...]

PR #466, opened 2020-04-23
Don't run the app as root

carbonin:
This makes the orchestrator and work pods run in the restricted scc rather than anyuid

A bunch of certifications require us not to run as root
in containers. Additionally the vast majority of the application
should run fine without root and the parts that need root are
typically things we can't do in containers anyway.

Specifically this changes the permissions of some directories to
allow the root group access which is what we'll be running as when
using the restricted scc in OpenShift

Related to https://github.com/ManageIQ/manageiq-pods/issues/442

Merge with https://github.com/ManageIQ/manageiq/pull/20095

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,25 +48,6 @@ _**Note:**_ This section assumes you have a basic user.
 $ oc new-project <project_name>
 ```
 
-### Add the anyuid and orchestrator service accounts to the anyuid security context
-
-_**Note:**_ The current MIQ images require the root user.
-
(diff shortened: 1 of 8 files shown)


simaishi:
Jansa backport details:

```
$ git log -1
commit fbc19317aac0941697e148a7817c46c825075924
Date:   Fri Apr 24 14:45:39 2020 -0400

    Merge pull request #466 from carbonin/dont_run_app_as_root

    Don't run the app as root

    (cherry picked from commit 420810947f374f936106a94638765dc6ead01d31)
```

Issue #467, opened 2020-04-23
Decrease privileges for httpd pod


carbonin:
The httpd pod should only require minimal privileges when using database-only auth or OIDC.
We can write logic in the operator to deploy a smaller and simpler image which doesn't require elevated privileges when either internal or OIDC auth is requested.

The specific tasks associated with this effort would be:
- [x] Separate out the httpd images based on required permissions
  - Includes building a minimal image for internal/OIDC auth
  - The lower privilege image should be running in the restricted scc with no init system
  - https://github.com/ManageIQ/manageiq-pods/pull/471
- [x] Implement operator logic to deploy the minimal necessary httpd image
  - Lower privilege image for internal auth and OIDC
  - init-based image for other forms of external auth
  - https://github.com/ManageIQ/manageiq-pods/pull/471
- [ ] Pull [OIDC config map generator logic](https://github.com/ManageIQ/httpd_configmap_generator/blob/master/lib/httpd_configmap_generator/oidc.rb) into the operator
  - OIDC provider information should be provided to the operator in a secret which can then by translated into a secret mounted into the httpd deployment
  - https://github.com/ManageIQ/manageiq-pods/pull/479

carbonin:
Additionally this kind of logic won't really be available for the template deploy so once this is completed I think it would make sense to move to the operator as the only way to deploy the app.

cc @sethrpeterson

carbonin:
Fixing this issue also makes https://github.com/ManageIQ/manageiq-pods/issues/361 a lower priority for us.

After this fix, we will be able to use database auth or oidc without any elevated privileges. Only the other auth types will require anyuid/privileged depending on the status of #361

PR #468, opened 2020-04-24
Use 8080 for httpd and 8081 for dbus api

carbonin:
This will make switching in an httpd pod without root privileges
easier as it won't be able to bind to port 80

Requires https://github.com/ManageIQ/container-httpd/pull/45

Related to https://github.com/ManageIQ/manageiq-pods/issues/467

### DIFF
--- a/templates/app/httpd.yaml
+++ b/templates/app/httpd.yaml
@@ -11,13 +11,14 @@ objects:
       app: "${APP_NAME}"
   data:
     application.conf: |
+      Listen 8080
       # Timeout: The number of seconds before receives and sends time out.
       Timeout 120
 
       RewriteEngine On
       Options SymLinksIfOwnerMatch
 
-      <VirtualHost *:80>
+      <VirtualHost *:8080>
         KeepAlive on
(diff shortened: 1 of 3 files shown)


templates/app/httpd.yaml
```diff
@@ -312,12 +313,14 @@ objects:
             name: httpd-auth-configs
         containers:
         - name: httpd
+          securityContext:
+            privileged: true
```
bdunne, line 317: Why do we need `privileged` if we're switching to non-privileged ports?

carbonin: oops, wasn't supposed to commit this. Will fix in a bit.

simaishi:
Jansa backport details:

```
$ git log -1
commit 9556c37dcc37ff9d69da7ea7e249565545cd25f0
Date:   Fri Apr 24 17:12:05 2020 -0400

    Merge pull request #468 from carbonin/httpd_port_8080

    Use 8080 for httpd and 8081 for dbus api

    (cherry picked from commit 882e5235f90aee335bffbf1ba3ec5cfa81e2c79e)
```

PR #471, opened 2020-04-28
Only use httpd with systemd when necessary

carbonin:
This PR changes the operator to deploy an httpd image which requires fewer privileges and does not run `systemd` when we don't need it. It also adds the image to this repo (which we may or may not want to move forward with).

When the `HttpdAuthenticationType` parameter is either `internal` or `openid-connect`, we will deploy the image without systemd and will not require the httpd sa to have anyuid or privileged scc access.

This handles the first two bullets of https://github.com/ManageIQ/manageiq-pods/issues/467
I'll open a separate patch to configure oidc using the operator.

My only open issues with this PR are the name of the new image and its location. I would prefer this image be the un-qualified `manageiq/httpd` image as it's the less complicated of the two and change the name of the original image to `manageiq/httpd-init` or something. Additionally I'm not sure if we want it to live in this repo, the container-httpd repo, or somewhere else.

### DIFF
--- a/manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq.org_manageiqs_crd.yaml
+++ b/manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq.org_manageiqs_crd.yaml
@@ -51,11 +51,19 @@ spec:
             databaseVolumeCapacity:
               description: 'Database volume size (default: 15Gi)'
               type: string
+            httpdAuthenticationType:
+              description: 'Type of httpd authentication (default: internal) Options:
+                internal, external, active-directory, saml, openid-connect Note: external,
+                active-directory, and saml require an httpd container with elevated
+                privileges'
+              type: string
             httpdCpuRequest:
               description: 'Httpd deployment CPU request (default: no request)'
               type: string
-            httpdImageName:
-              description: 'Image used for the httpd deployment (default: manageiq/httpd)'
+            httpdImageNamespace:
+              description: 'Image namespace used for the httpd deployment (default:
+                manageiq) Note: the exact image will be determined by the authentication
+                method selected'
               type: string
             httpdImageTag:
               description: 'Image tag used for the httpd deployment (default: latest)'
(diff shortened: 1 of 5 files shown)


Fryguy:
I agree with you...I like httpd and httpd-init.  Is that difficult to change?

With respect to where the Dockerfiles live, should they be separate repos container-httpd and containter-httpd-init to match the current pattern?

manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq.org_manageiqs_crd.yaml
```diff
@@ -51,11 +51,19 @@ spec:
...
+                internal, external, active-directory, saml, openid-connect Note: external,
+                active-directory, and saml require an httpd container with elevated
+                privileges'
+              type: string
             httpdCpuRequest:
               description: 'Httpd deployment CPU request (default: no request)'
               type: string
             httpdImageName:
-              description: 'Image used for the httpd deployment (default: manageiq/httpd)'
+              description: 'Image namespace used for the httpd deployment (default:
```
Fryguy, line 64: Since this is the namespace should the key be httpdImageNamespace?

carbonin: Yeah, missed a rebuild

carbonin: Fixed

simaishi:
Jansa backport details:

```
$ git log -1
commit be74b8bbd00380a16b15b0f65f82bda9a64bf744
Date:   Wed Apr 29 14:50:40 2020 -0400

    Merge pull request #471 from carbonin/unprivileged_httpd

    Only use httpd with systemd when necessary

    (cherry picked from commit 4f0437ade43fccf3c940238a497cc6a13048503b)
```

PR #474, opened 2020-04-29
Pin the base image to ubi 8.1 and don't install net-snmp from ubi repos

carbonin:
The ubi repos are not versioned by minor version so we were pulling
in a later version from there which was incompatible with the earlier
version that we had available for the other net-snmp-* packages
from the CentOS repos.

This lead to errors such as:
```
Error:
 Problem: package net-snmp-utils-1:5.8-12.el8_1.1.x86_64 requires net-snmp-libs(x86-64) = 1:5.8-12.el8_1.1, but none of the providers can be installed
  - cannot install both net-snmp-libs-1:5.8-14.el8.x86_64 and net-snmp-libs-1:5.8-12.el8_1.1.x86_64
  - cannot install the best candidate for the job
(try to add '--allowerasing' to command line to replace conflicting packages or '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

### DIFF
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi8/ubi
+FROM registry.access.redhat.com/ubi8/ubi:8.1
 
 ARG MIQ_REF=master
 ARG SUI_REF=master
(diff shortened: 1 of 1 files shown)


simaishi:
Jansa backport details:

```
$ git log -1
commit 6a74bd07f824cf51a2f02dc8f393ba053fb7bbef
Date:   Wed Apr 29 13:11:45 2020 -0400

    Merge pull request #474 from carbonin/net_snmp_ubi_81

    Pin the base image to ubi 8.1 and don't install net-snmp from ubi repos

    (cherry picked from commit 31b4ae03940762349cb09f37263fc5487ab37b98)
```

PR #478, opened 2020-04-30
Remove external auth from the template-based deployment

carbonin:
We're now using a separate image to implement external auth. In the operator we can switch between these two images based on the parameters provided in the CR, but the templates don't give us that kind of flexibility.

I think it's a better user-experience to be able to deploy the templates out-of-the-box without needing admin privileges rather than having the full suite of external auth available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,52 +48,6 @@ _**Note:**_ This section assumes you have a basic user.
 $ oc new-project <project_name>
 ```
 
-### Set up the httpd service account
-
-#### If running without OCI systemd hooks (Minishift)
-
-_**Note:**_ Minishift clusters are not equipped with OCI systemd hooks which are used to assist with containerized systemd deployments.
-
-An extra SCC is used on Minishift to compensate for the lack of oci-systemd-hooks. **Please SKIP this step if your cluster is equipped with oci-systemd-hooks.**
-
-__*As admin*__
-
-Create the miq-sysadmin SCC:
-
-```bash
-$ oc create -f templates/miq-scc-sysadmin.yaml
-```
-
-The miq-httpd service account must be added to the miq-sysadmin SCC before the front-end httpd pod can run.
-
-```bash
-$ oc adm policy add-scc-to-user miq-sysadmin system:serviceaccount:<your-namespace>:<app-name>-httpd
-```
-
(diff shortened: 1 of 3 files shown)


simaishi:
Jansa backport details:

```
$ git log -1
commit 78354135dbc5f30671737df6a2f320cc5decea54
Date:   Thu Apr 30 15:51:42 2020 -0400

    Merge pull request #478 from carbonin/no_httpd_auth_templates

    Remove external auth from the template-based deployment

    (cherry picked from commit 03a263b16264064a13b16b210ff0f0131b3fa944)
```

carbonin:
So @Fryguy I think switching out the name of the httpd image means that we broke all the other releases that don't have this PR. So we either need to open a PR to < ivanchuk to change the image to httpd-init or backport this.

Issue opened as https://github.com/ManageIQ/container-httpd/issues/50

Fryguy:
> So @Fryguy I think switching out the name of the httpd image means that we broke all the other releases that don't have this PR. So we either need to open a PR to < ivanchuk to change the image to httpd-init or backport this.
> 
> Issue opened as [ManageIQ/container-httpd#50](https://github.com/ManageIQ/container-httpd/issues/50)

Doh...I hadn't considered that.  Yeah, that's an actual issue.

Fryguy:
> So we either need to open a PR to < ivanchuk to change the image to httpd-init or backport this.

Yeah I'm leaning towards this approach.  It means that ivanchuk-4 and below would stay broken, but I think that's ok as people should upgrade to ivanchuk-5.  We may need to release an emergency ivanchuk-5 sooner than later though.  cc @simaishi @dmetzger57 @chessbyte

carbonin:
> this approach

Which approach? There were two options in my comment ... Do you want to backport this or make separate PRs to keep the old versions the same as the were (using `httpd-init`)

Fryguy:
Sorry...I quoted only one approach and meant that one  (actually I quoted both then edited one away, so maybe you saw it before that)

carbonin:
I don't know. I guess it depends on if we're still releasing bug-fixes for those versions. The image name  has always been configurable so there is a pretty simple workaround.

carbonin:
Opened https://github.com/ManageIQ/manageiq-pods/pull/510 @Fryguy @simaishi

PR #479, opened 2020-05-01
Allow more flexible external auth configuration

### LINKED ISSUE #467: Decrease privileges for httpd pod

carbonin:
This PR changes the way the operator deploys the external auth configuration for the httpd pod.
Now, we only deploy exactly what is required for the authentication type specified in the CR.
Additionally there are now two options for deploying OIDC authentication.

1. What we had before minus the [httpd_configmap_generator](https://github.com/ManageIQ/httpd_configmap_generator)
    - Previously a user would have to utilize the config map generator to translate their oidc url, client secret, and client id into a config map then overwrite the existing auth-configs created by the operator after the deployment finishes
    - Now the url can be supplied as `oidcProviderURL` in the CR and a secret should be created with the keys `CLIENT_ID` and `CLIENT_SECRET` and the name supplied as `oidcClientSecret` in the CR
2. Provide all the config files in a secret directly
    - In this case the operator takes the secret name from `httpdAuthConfig` and mounts all the files in that secret into `/etc/httpd/user-conf.d`
    - Any file named `*.conf` in that directory is included into the httpd config

Fixes #467 

cc @gyliu513 @jvlcek @abellotti 

This PR was mostly focused on OIDC, but in the future we should allow the other types to utilize the `httpdAuthConfig` parameter to provide the config files from the config map generator rather than creating a stub config map that then needs to be overwritten.

Longer term we could enhance the config map generator to actually create the configmaps in the cluster and have the operator run it as a job.

Another idea I had, but didn't pursue very far was that we could probably create a separate controller an CR just for auth, I'm not sure this would be worth the effort, but it feels like we'll end up with quite a lot of auth-related parameters in the manageiq CR if we try to move all the inputs to the generator into this CR.

### DIFF
--- a/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
+++ b/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
@@ -1,7 +1,10 @@
 package v1alpha1
 
 import (
+	"errors"
+	"fmt"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strings"
 )
 
 // ManageIQSpec defines the desired state of ManageIQ
@@ -56,6 +59,10 @@ type ManageIQSpec struct {
 	// Only used with the openid-connect authentication type
 	// +optional
 	OIDCClientSecret string `json:"oidcClientSecret"`
+	// Secret containing the httpd configuration files
+	// Mutually exclusive with the OIDCCliencSecret and OIDCProviderURL if using openid-connect
+	// +optional
+	HttpdAuthConfig string `json:"httpdAuthConfig"`
 
 	// Httpd deployment CPU request (default: no request)
 	// +optional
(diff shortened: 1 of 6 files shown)


manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -48,6 +51,18 @@ type ManageIQSpec struct {
...
+	// URL for the OIDC provider
+	// Only used with the openid-connect authentication type
+	// +optional
+	OIDCProviderURL string `json:"oidcProviderURL"`
+	// Secret name containing the OIDC client id and secret
+	// Only used with the openid-connect authentication type
+	// +optional
+	OIDCClientSecret string `json:"oidcClientSecret"`
+	// Secret containing the httpd configuration files
+	// Mutually exclusive with the OIDCCliencSecret and OIDCProviderURL if using openid-connect
```
Fryguy, line 63:
```suggestion
	// Mutually exclusive with the OIDCClientSecret and OIDCProviderURL if using openid-connect
```

simaishi:
Jansa backport details:

```
$ git log -1
commit ba15ecafc55321dd54f46394bf3c9cbfc8a5efff
Date:   Fri May 1 17:11:25 2020 -0400

    Merge pull request #479 from carbonin/oidc

    Allow more flexible external auth configuration

    (cherry picked from commit 1b5a80c50e6af361b82b9b9932844a3239163537)
```

PR #484, opened 2020-05-05
Configure application auth settings when passed to the operator

carbonin:
If the CR is configured to set up authentication for saml or oidc
we should also configure the application settings to match the
environment.

This will happen when the orchestrator starts up, but before any
workers are started.


images/manageiq-orchestrator/container-assets/entrypoint
```diff
@@ -16,6 +16,33 @@ function check_svc_status() {
...
+      'authentication/saml_enabled'         => provider_type == 'saml',
+      'authentication/oidc_enabled'         => provider_type == 'oidc',
+      'authentication/provider_type'        => provider_type,
+      'authentication/local_login_disabled' => ENV['LOCAL_LOGIN_DISABLED'] == "true"
+    }
+    MiqRegion.my_region.add_settings_for_resource(settings)
+EOS
+)
+
+  bin/rails runner "$script"
```
Fryguy, line 43: Not for this PR, but I'm wondering if this should be a core rake task.  Then here we can simplify this to something like `bundle exec rake evm:auth:setup`

manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
```diff
@@ -51,6 +51,13 @@ spec:
             databaseVolumeCapacity:
               description: 'Database volume size (default: 15Gi)'
               type: string
+            disableApplicationLocalLogin:
```
Fryguy, line 54: I'm finding the negatively named option with a default of false confusing.  Is it possible to make it `allowApplicationLocalLogin` with a default of `true` ?

manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
```diff
@@ -51,6 +51,13 @@ spec:
...
+              description: 'Flag to disable local login to the application (default:
+                false)'
+              type: boolean
+            enableSSO:
+              description: 'Flag to enabile sso in the application (default: false)'
```
Fryguy, line 59:
```suggestion
              description: 'Flag to enable SSO in the application (default: false)'
```

images/manageiq-orchestrator/container-assets/entrypoint
carbonin: Yeah, there was the one for setting up settings, but creating the command line for that was going to be just as difficult as this. Plus it only worked for server settings and we needed region here.

manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
carbonin: I was just matching what the setting in our database was, but I can definitely change it here if you think the translation won't be more confusing than the option name.

Fryguy: I guess I don't like our decision at the database either :/

I don't know how hard it is to change CRDs later, which is why I'm thinking go positive now and we can change our application later, but I'm not sure...it's up to you.

carbonin: I think the CRD should be semver versioned (at least for OLM) so I think we can go to positive now.

carbonin: Hm, so when trying to change this the "default value" semantics are a bit difficult for booleans.

In Go the "zero value" for a bool is false so the case when someone doesn't set a value and when a user explicitly sets `false` look identical.

So if we want to keep this a bool I think the defalut value needs to remain `false`. Otherwise we could change it to a string, but that might be weird.

What do you think @Fryguy ?

Fryguy: Well you know what I think of go LOLOLOLOL 🤣 

In go, is it possible to have the field be like a `Bool | Nil` union thing (i.e. Maybe or Option type from other langs)?  If so, then you have the possibility to differentiate "null" vs "false"

carbonin: So that might be an option, but I don't know how that would work for the CRD as it has to map to a type that the sdk knows about I think. Generally I'd rather not use a non-standard type in the CR if we can avoid it.

I've seen internal structs to this with pointers but I don't know how that would work for something that needs to be provided in a yaml file. For example: https://github.com/kubernetes/api/blob/a4f2d87006da76427d0d3d613f2ffec13d8e49d3/apps/v1/types.go#L274. I'll give this a shot.

Fryguy: Ah ok, yeah that package I linked I think works in effectively the same way (i.e. use a pointer, and nil means not specified)

carbonin: The pointer conversion seems to have done the job.

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -223,6 +229,15 @@ func (m *ManageIQ) Initialize() {
...
 	}
 
+	if spec.EnableSSO == nil {
+		spec.EnableSSO = new(bool)
+	}
+
+	if spec.EnableApplicationLocalLogin == nil {
+		t := true
+		spec.EnableApplicationLocalLogin = &t
+	}
```
Fryguy, line 239: ❤️

simaishi:
Jansa backport details:

```
$ git log -1
commit df1dabde0caf7f736652d73fcff8b2b97237ab06
Date:   Fri May 8 15:38:30 2020 -0400

    Merge pull request #484 from carbonin/oidc_app_config

    Configure application auth settings when passed to the operator

    (cherry picked from commit a03e6745aa57ce07d7b10b83476cf2aae5b228ac)
```

PR #485, opened 2020-05-05
RPM based pods build

simaishi:
~~TODO: replace my copr yum repo with manageiq.~~

Depends on: https://github.com/ManageIQ/manageiq-appliance-build/pull/415
Also includes https://github.com/ManageIQ/manageiq-appliance-build/pull/420


images/manageiq-base/Dockerfile
```diff
@@ -19,17 +12,6 @@ ARG HOOKS_SCRIPT_URL
...
-RUN mkdir sui && \
-    if [[ -n "$GIT_AUTH" ]]; then GIT_HOST=${GIT_AUTH}@${GIT_HOST}; fi && curl -L https://${GIT_HOST}/${SUI_ORG}/${CORE_REPO_NAME}-ui-service/tarball/${SUI_REF} | tar vxz -C sui --strip 1 && \
-    echo "`git ls-remote https://${GIT_HOST}/${SUI_ORG}/${CORE_REPO_NAME}-ui-service.git ${SUI_REF} | cut -f 1`" > sui/BUILD
-
-RUN mkdir app && \
-    if [[ -n "$GIT_AUTH" ]]; then GIT_HOST=${GIT_AUTH}@${GIT_HOST}; fi && curl -L https://${GIT_HOST}/${MIQ_ORG}/${CORE_REPO_NAME}/tarball/${MIQ_REF} | tar vxz -C app --strip 1 && \
-    echo "`date +'%Y%m%d%H%M%S'`_`git ls-remote https://${GIT_HOST}/${MIQ_ORG}/${CORE_REPO_NAME}.git ${MIQ_REF} | cut -c 1-7`" > app/BUILD
-
 RUN mkdir build && \
     if [[ -n "$GIT_AUTH" ]]; then GIT_HOST=${GIT_AUTH}@${GIT_HOST}; fi && curl -L https://${GIT_HOST}/${BUILD_ORG}/${CORE_REPO_NAME}-appliance-build/tarball/${BUILD_REF} | tar vxz -C build --strip 1
```
Fryguy, line 16: We may be able to drop this entire hook now, because all of the customization should happen in the RPM layer.  cc @carbonin

simaishi: 2 things are still coming outside of RPMs. I'll remove the hook when that goes away.

https://github.com/ManageIQ/manageiq-pods/pull/485/files#diff-4f1c0cb40242780de19fd44d5c011b0eL121-L124

Fryguy:
@simaishi Do we also need any changes in the [core Dockerfile](https://github.com/ManageIQ/manageiq/blob/master/Dockerfile)?

images/manageiq-base/Dockerfile
```diff
@@ -57,108 +37,29 @@ LABEL name="manageiq-base" \
...
-      postgresql-devel            \
-      procps-ng                   \
       python3-devel               \
-      qpid-proton-c-devel         \
-      redhat-rpm-config           \
-      ruby-devel                  \
-      socat                       \
-      sqlite-devel                \
-      v2v-conversion-host-ansible \
       wget                        \
```
Fryguy, line 53: @simaishi I wonder if we should create a manageiq-pods-tools rpm analogous to the manageiq-appliance-tools rpm that houses stuff like wget and patch here.  What do you think? (not for this PR, we can iterate on that later)

simaishi:
> @simaishi Do we also need any changes in the [core Dockerfile](https://github.com/ManageIQ/manageiq/blob/master/Dockerfile)?

So far, I only have 1 change there: https://github.com/ManageIQ/manageiq/pull/20127

images/manageiq-base/Dockerfile
simaishi: I'm not sure where file, lvm2, patch and wget are used, so I left them here for now. I'm hoping all those will go to one of the manageiq-* RPM "requires". If they are just tools and not required to run our app, we can certainly create manageiq-pods-tools.

Fryguy: @carbonin do you know?  I assumed patch and wget were just convenience tools.

carbonin:
Hm, this will make things a lot more difficult to run from source in containers.
Do we have a script/tool to build the rpms from local source?

carbonin:
Also given the differences between the architectures, is this going to break ppc64le build again?

simaishi:
> Do we have a script/tool to build the rpms from local source?

https://github.com/ManageIQ/manageiq-rpm_build contains info and scripts needed to create RPMs.

> Also given the differences between the architectures, is this going to break ppc64le build again?

I think the idea here is for whoever need to build ppc64le to provide pcc64le RPMs. So we don't need to worry about problem building gems or some RPMs missing in this repo. That will be a part of rpm build and RPMs with needed changes/workarounds can be shared between pods and appliances.

We just have to point to an arch specific yum repo (a follow up PR).

Fryguy:
> Also given the differences between the architectures, is this going to break ppc64le build again?

If the RPMs are built as ppc64, I'd expect that when installed into a ppc64 container it would pull the right thing.

Fryguy:
> We just have to point to an arch specific yum repo (a follow up PR).

Do the repos have to be arch specific or can multiple arch's exist in the same repo and yum will pull the right one?

simaishi:
We can use `$basearch` so it will go to the right one depending on the environment.

Fryguy:
> Hm, this will make things a lot more difficult to run from source in containers.

This is a good point, and is technically the same problem we discussed with appliances.  On appliances, we decided that the majority use case are users, not developers, and we can document away turning an appliance into a dev appliance.  Because containers are so much easier to build I could see them being more of a dev convenience, but even then, I'd argue that the majority use case is users.  Unfortunately, @carbonin, you in particular are one of the few developers, so admittedly this affects you the most.

Let's brainstorm on something...I have a feeling that we can do RPMs for normal build, and then find a way to drive source builds via ENV var or a particular volume mount or something

> Do we have a script/tool to build the rpms from local source?

Building RPMs is a simple docker command [1], but the feedback cycle time on it can be very long (like 30 minutes to do a build because it has to do all bundler/node/webpack), so that feels pretty nasty.

[1] `docker run -it manageiq/rpm_build build`

carbonin:
> Because containers are so much easier to build I could see them being more of a dev convenience, but even then, I'd argue that the majority use case is users.

Oh I agree, I only brought it up because there are much more differences between running in a container and running on a laptop. So there are cases where things will *need* to be tested in pods before merge.

> Building RPMs is a simple docker command [1], but the feedback cycle time on it can be very long (like 30 minutes to do a build because it has to do all bundler/node/webpack), so that feels pretty nasty.

Building the entire chain of containers took way longer so that's why I asked about building the RPMs. If we script something that would build the RPMs from a branch then use local RPMs for the container builds as an override, that would work for me.

simaishi:
> Building the entire chain of containers took way longer so that's why I asked about building the RPMs. If we script something that would build the RPMs from a branch then use local RPMs for the container builds as an override, that would work for me.

OK, so I think we're missing 2 things right now.
1. support for "local" RPMs
1. exposing repo/branch options per repo

I'll take a look.

images/manageiq-base/Dockerfile
carbonin: `wget` is used here https://github.com/ManageIQ/manageiq-pods/pull/485/files#diff-4f1c0cb40242780de19fd44d5c011b0eL167 but I suppose we could use `curl`.

No idea where `lvm2` or `patch` are used (if at all)

Fryguy: Ah yeah, if we can use curl there and avoid adding a package I'm :+1: for that

Fryguy: Can we just try dropping lvm2 and patch and see what happens?

Fryguy: Also noticed `file`

simaishi:
> 1. support for "local" RPMs

https://github.com/ManageIQ/manageiq-pods/pull/491

> 2. exposing repo/branch options per repo

https://github.com/ManageIQ/manageiq-rpm_build/pull/41

Fryguy:
@simaishi Since we have a direction in #491, can this be unWIP'd now?

simaishi:
> @simaishi Since we have a direction in #491, can this be unWIP'd now?

Not until we actually have RPMs in Copr.

simaishi:
Regarding ppc64le, 3 changes made in this repo specifically for pcc64le will need to be made in manageiq-rpm_build.

- Do not install wmi in manageiq-base image
- Install python2 on build container
- special handling of sassc and unf_ext gems

Opened https://github.com/ManageIQ/manageiq-rpm_build/issues/43

images/manageiq-base/Dockerfile
carbonin: > Can we just try dropping lvm2 and patch and see what happens?

I'm okay with that.

simaishi: Removed file, lvm2, patch and wget (But file is installed as a dependency anyway).

images/manageiq-ui-worker/Dockerfile
```diff
@@ -8,9 +8,7 @@ LABEL name="manageiq-ui-worker" \
       summary="ManageIQ user interface worker image"
 
 # Install httpd and remove the existing httpd config from manageiq-appliance
-RUN dnf -y --disableplugin=subscription-manager --setopt=tsflags=nodocs install \
```
Fryguy: @simaishi The comment now is kind of misleading?  Maybe something like: ?

"manageiq-ui will install httpd, so remove the existing httpd config from manageiq-appliance"

carbonin:
I built the images from this branch and when running the orchestrator I get this in the logs:

```
sh-4.4$ bin/rake environment
rake aborted!
Errno::EACCES: Permission denied @ dir_s_mkdir - /var/lib/data
/var/www/miq/vmdb/lib/vmdb_helper.rb:28:in `<top (required)>'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `require'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `block in require'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:257:in `load_dependency'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `require'
/var/www/miq/vmdb/config/application.rb:145:in `block in <class:Application>'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/lazy_load_hooks.rb:69:in `block in execute_hook'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/lazy_load_hooks.rb:62:in `with_execution_control'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/lazy_load_hooks.rb:67:in `execute_hook'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/lazy_load_hooks.rb:52:in `block in run_load_hooks'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/lazy_load_hooks.rb:51:in `each'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/lazy_load_hooks.rb:51:in `run_load_hooks'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/application/bootstrap.rb:81:in `block in <module:Bootstrap>'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/initializable.rb:32:in `instance_exec'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/initializable.rb:32:in `run'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/initializable.rb:61:in `block in run_initializers'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/initializable.rb:60:in `run_initializers'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/application.rb:361:in `initialize!'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/railtie.rb:190:in `public_send'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/railtie.rb:190:in `method_missing'
/var/www/miq/vmdb/config/environment.rb:5:in `<top (required)>'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `require'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `block in require'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:257:in `load_dependency'
/opt/manageiq/manageiq-gemset/gems/activesupport-5.2.4.3/lib/active_support/dependencies.rb:291:in `require'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/application.rb:337:in `require_environment!'
/opt/manageiq/manageiq-gemset/gems/railties-5.2.4.3/lib/rails/application.rb:520:in `block in run_tasks_blocks'
Tasks: TOP => environment
(See full trace by running task with --trace)
```

@jrafanie any idea why `rake environment` is trying to create a directory in `/var/lib`?

carbonin:
Well that wasn't hard: https://github.com/ManageIQ/manageiq/blob/5996d45ed01c2d2f4cd662fd26ecc7f9e3218fc8/lib/vmdb_helper.rb#L26-L28

But why is it failing now and not before?

images/manageiq-base/Dockerfile
```diff
@@ -56,115 +36,36 @@ LABEL name="manageiq-base" \
...
-
-# build and install gems
-WORKDIR ${APP_ROOT}
-RUN source /etc/default/evm && \
-    export RAILS_USE_MEMORY_STORE="true" && \
-    gem install bundler -v ">=1.16.2" && \
-    if [ ${ARCH} = "ppc64le" ] ; then gem install sassc  -- --disable-march-tune-native && \
-        gem install unf_ext -v '0.0.7.2' -- --with-cxxflags="-fsigned-char"; fi && \
-    bundle install --with qpid_proton && \
-    rake tmp:clear evm:compile_sti_loader log:clear && \
```
carbonin, line 145: This line was actually creating the `/var/lib/data` directory at build time which is why we were not having this problem before.

Not sure if the right answer is to just dump a `rake environment` in here at build time or try to make the data directory logic conditional on if we're running in a container or not... Thoughts @Fryguy @jrafanie ?

carbonin: I guess it also depends on what we're using that directory for at runtime. If it's not in the source tree we might not have write permissions to it.

Fryguy: What is this used for?  I was assuming the miq_tmp stuff was for the RHV fleecing code, but according to https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/main/disk_layout.ks.erb#L7  it should be at /var/www/miq_tmp .  @simaishi @jrafanie ?

Even so, I don't think it actually matters where the data directory ultimately ends up, particularly in podified.  I'm not even sure the RHV fleecing code will work in podified, regardless.

Either way I feel like this directory should be precreated at build time with write permissions for the application.  So, I'm fine with a `rake environment`, or even a straight `RUN mkdir -p /var/lib/data` in the Dockerfile.

carbonin: @Fryguy I'm actually thinking it might make more sense to just add to [this conditional](https://github.com/ManageIQ/manageiq/blob/5996d45ed01c2d2f4cd662fd26ecc7f9e3218fc8/lib/vmdb_helper.rb#L26) so that it stays within the source tree. Then we don't have to fix the permissions.

carbonin: Neither of the locations are on a volume in pods, so it makes more sense for it to live in rails tmp directory.

carbonin: https://github.com/ManageIQ/manageiq/pull/20224 Fixes it

Fryguy:
Merged https://github.com/ManageIQ/manageiq/pull/20224

carbonin:
Okay, so do we want to wait for updated rpm to test this end to end? @simaishi is that still a manual process or is that a part of the nightly build now?

simaishi:
It's a part of nightly build now, but I just kicked off another rpm build. Will ping you when done.

simaishi:
... and it's done.

carbonin:
Now the workers are coming up but they are all failing the liveness probe:
```
  Normal   Scheduled  8m                default-scheduler                                         Successfully assigned manageiq/1-reporting-68cb689fdb-cm5v5 to dell-r430-19.cloudforms.lab.eng.rdu2.redhat.com
  Normal   Created    2m (x3 over 8m)   kubelet, dell-r430-19.cloudforms.lab.eng.rdu2.redhat.com  Created container
  Normal   Started    2m (x3 over 8m)   kubelet, dell-r430-19.cloudforms.lab.eng.rdu2.redhat.com  Started container
  Warning  Unhealthy  29s (x9 over 5m)  kubelet, dell-r430-19.cloudforms.lab.eng.rdu2.redhat.com  Liveness probe failed: /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- active_support/core_ext/numeric/time (LoadError)
           from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:54:in `require'
           from /var/www/miq/vmdb/lib/workers/miq_defaults.rb:1:in `<top (required)>'
           from /var/www/miq/vmdb/lib/workers/heartbeat.rb:1:in `require_relative'
           from /var/www/miq/vmdb/lib/workers/heartbeat.rb:1:in `<top (required)>'
           from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:54:in `require'
           from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:54:in `require'
           from /var/www/miq/vmdb/lib/workers/bin/heartbeat_check.rb:27:in `<main>'
  Normal   Pulling  26s (x4 over 8m)  kubelet, dell-r430-19.cloudforms.lab.eng.rdu2.redhat.com  pulling image "docker.io/carbonin/manageiq-base-worker:latest"
  Normal   Killing  26s (x3 over 5m)  kubelet, dell-r430-19.cloudforms.lab.eng.rdu2.redhat.com  Killing container with id docker://1-reporting:Container failed liveness probe.. Container will be killed and recreated.
  Normal   Pulled   24s (x4 over 8m)  kubelet, dell-r430-19.cloudforms.lab.eng.rdu2.redhat.com  Successfully pulled image "docker.io/carbonin/manageiq-base-worker:latest"
```

Will look into this tomorrow.

Fryguy:
@simaishi Does anything special need to be done in order for arbitrary ruby scripts to work? https://github.com/ManageIQ/manageiq-pods/blob/d63f974bcd51a83fa09df4e5027bba670c732967/images/manageiq-base-worker/container-assets/manageiq_liveness_check#L3

simaishi:
Not that I know of.

carbonin:
Adding this patch seems to resolve the liveness probe issue:
```patch
diff --git a/images/manageiq-base-worker/container-assets/manageiq_liveness_check b/images/manageiq-base-worker/container-assets/manageiq_liveness_check
index 34c9ce5..258b5c5 100755
--- a/images/manageiq-base-worker/container-assets/manageiq_liveness_check
+++ b/images/manageiq-base-worker/container-assets/manageiq_liveness_check
@@ -1,3 +1,4 @@
 #!/bin/bash
 
+source ${APP_ROOT}/container_env
 ruby ${APP_ROOT}/lib/workers/bin/heartbeat_check.rb --heartbeat-file=${APP_ROOT}/tmp/worker.hb
```

I'm not sure exactly why, but it makes sense to me as `container_env` sets up both our ruby environment (by sourcing `/etc/default/evm`) and the database access by configuring `DATABASE_URL` correctly. I imagine if the liveness check were any more complicated than checking a file, we would need this anyway.

simaishi:
@carbonin Thank you for testing and fixing issues found!!

Updated the PR to include sourcing container_env.

simaishi:
Jansa backport details (replaced yum repo with 10-jansa):

```
$ git log -1
commit 347ff304282d63bc7053eae63da7bf91202c81fc
Date:   Fri May 29 11:49:16 2020 -0400

    Merge pull request #485 from simaishi/rpm_build

    RPM based pods build

    (cherry picked from commit ff7aa27dfbbaa262aae59a025ee3b3a8e13f9cda)
```

PR #491, opened 2020-05-07
Support for local RPMs

simaishi:
Built on top of #485.

Adding 2 options to bin/build:
- `-b` builds RPMs from scratch
- `-l` uses RPMs that exist locally (the expected directory structure is what `-b` produces: manageiq-base/rpms/\<arch\>/*rpm

Also, 2 env can be set for rpm build:
- `RPM_BUILD_OPTIONS` overrides [default rpm build options](https://github.com/ManageIQ/manageiq-rpm_build/blob/master/config/options.yml) (repo, branches, etc.) - this is the path to the directory where override options.yml is.
- `RPM_BUILD_IMAGE` sets rpm build container image name (default: manageiq/rpm_build)

Notes:
- Running with `-b` will remove previously built RPMs.
- Considering this is a "developer" mode, I skipped cleanup (removing RPMs) in manageiq-base and only added cleanup to manageiq-ui-worker, to avoid having duplicate code in manageiq-ui-worker build. If that's a problem, I will change.

Tested with https://github.com/ManageIQ/manageiq-rpm_build/pull/40

~~TODO: replace my copr yum repo with manageiq.~~.

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -15,14 +15,19 @@ HOOKS_SCRIPT_URL=${HOOKS_SCRIPT_URL:-""}
 
 OPERATOR_DIR=${OPERATOR_DIR:-manageiq-operator}
 
-while getopts "t:d:r:nhp" opt; do
+RPM_BUILD_OPTIONS=${RPM_BUILD_OPTIONS:-""}
+RPM_BUILD_IMAGE=${RPM_BUILD_IMAGE:-"manageiq/rpm_build"}
+
+while getopts "t:d:r:bnhlp" opt; do
   case $opt in
+    b) REBUILD_RPM="true" ;;
     d) IMAGE_DIR=$OPTARG ;;
+    l) LOCAL_RPM="true" ;;
     n) NO_CACHE="true" ;;
     p) PUSH="true" ;;
     r) REPO=$OPTARG ;;
     t) TAG=$OPTARG ;;
-    h) echo "Usage: $0 -d IMAGE_DIR -r IMAGE_REPOSITORY [-hnp] [ -t IMAGE_TAG ]"; exit 1
+    h) echo "Usage: $0 -d IMAGE_DIR -r IMAGE_REPOSITORY [-hblnp] [ -t IMAGE_TAG ]"; exit 1
   esac
 done
 
@@ -36,7 +41,25 @@ if [ -z "$REPO" ]; then
   exit 1
 fi
 
+if [ -n "$LOCAL_RPM" ] && [ -n "$REBUILD_RPM" ]; then
+  echo "Local rpm (-l) and rebuild rpm (-b) options can't be used together"
+  exit 1
+fi
+
 set -e
+
+if [ -n "$REBUILD_RPM" ]; then
+  rpm_dir=$IMAGE_DIR/manageiq-base/rpms
+  rm -rf $rpm_dir/*
+  options="-v $PWD/$rpm_dir:/root/BUILD/rpms"
+  if [ -n "$RPM_BUILD_OPTIONS" ]; then
+    options+=" -v $PWD/$RPM_BUILD_OPTIONS:/root/OPTIONS"
+  fi
+  docker pull $RPM_BUILD_IMAGE
+  cmd="docker run -it --rm $options $RPM_BUILD_IMAGE build"
+  echo "Building RPMs locally: $cmd"
+  $cmd
+fi
 
 pushd $IMAGE_DIR
   cmd="docker build --tag $REPO/manageiq-base:$TAG \
@@ -51,6 +74,10 @@ pushd $IMAGE_DIR
 
   if [ -n "$NO_CACHE" ]; then
     cmd+=" --no-cache"
+  fi
+
+  if [ -n "$LOCAL_RPM" ] || [ -n "$REBUILD_RPM" ]; then
+    cmd+=" --build-arg LOCAL_RPM=true"
   fi
 
   echo "Building manageiq-base: $cmd"
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -22,6 +22,7 @@ FROM registry.access.redhat.com/ubi8/ubi:8.1
 MAINTAINER ManageIQ https://manageiq.org
 
 ARG ARCH=x86_64
+ARG LOCAL_RPM
 
 ENV TERM=xterm \
     CONTAINER=true \
@@ -36,6 +37,9 @@ LABEL name="manageiq-base" \
       io.k8s.description="ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures." \
       io.openshift.tags="ManageIQ,miq,manageiq"
 
+COPY rpms/* /tmp/rpms/
+COPY container-assets/create_local_yum_repo.sh /
+
 RUN curl -L https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo > /etc/yum.repos.d/ansible-runner.repo
 
 RUN dnf -y --disableplugin=subscription-manager install \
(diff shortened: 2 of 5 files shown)


images/manageiq-base/container-assets/create_local_yum_repo.sh
```diff
@@ -0,0 +1,15 @@
...
+
+cat > /etc/yum.repos.d/local_rpm.repo << EOF
+[local-rpm]
+baseurl=file:///tmp/rpms/$basearch
+name=Local yum repo
+enabled=1
+gpgcheck=0
+EOF
+
+dnf config-manager --setopt=copr:copr.fedorainfracloud.org:simaishi:test.exclude=manageiq-* --save
```
Fryguy, line 15: What will this look like when it's not your personal repo?

images/manageiq-ui-worker/Dockerfile
```diff
@@ -18,3 +18,5 @@ RUN chgrp root /var/run/httpd && chmod g+rwx /var/run/httpd && \
 RUN sed -i '/^Listen 80/d' /etc/httpd/conf/httpd.conf
 
 COPY container-assets/manageiq-http.conf /etc/httpd/conf.d
+
+RUN rm -rf /tmp/rpms /create_local_yum_repo.sh /etc/yum.repos.d/local_rpm.repo
```
Fryguy, line 22: Why do we clean up here, but not in the base Docker file?

images/manageiq-base/container-assets/create_local_yum_repo.sh
simaishi: `--setopt=copr:copr.fedorainfracloud.org:manageiq:ManageIQ-Master.exclude=manageiq-*`

But we will have to add support for customization, for both the repo name and the RPM name.

images/manageiq-ui-worker/Dockerfile
simaishi: Explained in 2nd 'Note' in the PR description.  I can change for both manageiq-base and manageiq-ui-worker to do the same local rpm steps (copy rpm, install createrepo, disable default ones and cleanup).

bin/build
```diff
@@ -45,18 +41,30 @@ if [ -z "$REPO" ]; then
   exit 1
 fi
 
+if [ -n "$LOCAL_RPM" ] && [ -n "$REBUILD_RPM" ]; then
+  echo "Local rpm (-l) and rebuild rpm (-b) options can't be used together"
```
Fryguy, line 45: Why not?  I think of rebuild_rpm as a subset of local_rpm.  That is rebuild_rpm builds the rpms and writes them into the local-rpm directory, and local-rpm says to use what's in the local-rpm directory.  Or am I minunderstanding?

bin/build
```diff
@@ -45,18 +41,30 @@ if [ -z "$REPO" ]; then
...
+fi
+
 set -e
 
+if [ -n "$REBUILD_RPM" ]; then
+  rpm_dir=$IMAGE_DIR/manageiq-base/rpms
+  rm -rf $rpm_dir/*
+  options="-v $PWD/$rpm_dir:/root/BUILD/rpms"
+  if [ -n "$RPM_BUILD_OPTIONS" ]; then
+    options+=" -v $PWD/$RPM_BUILD_OPTIONS:/root/OPTIONS"
```
Fryguy, line 56: Why prefix with $PWD?  I would expect the caller to specify the proper path

bin/build
```diff
@@ -45,18 +41,30 @@ if [ -z "$REPO" ]; then
...
+  echo "Local rpm (-l) and rebuild rpm (-b) options can't be used together"
+  exit 1
+fi
+
 set -e
 
+if [ -n "$REBUILD_RPM" ]; then
+  rpm_dir=$IMAGE_DIR/manageiq-base/rpms
+  rm -rf $rpm_dir/*
+  options="-v $PWD/$rpm_dir:/root/BUILD/rpms"
```
Fryguy, line 54: Same here.

images/manageiq-base/container-assets/create_local_yum_repo.sh
Fryguy: Probably can extract those from the options.yml for the build.

bin/build
simaishi: Since rebuild_rpm will delete RPMs on the machine if exists, I thought it will be confusing to allow both options to be used together.

rpm_build implies local_rpm, but the difference is "local RPMs already exist on the machine" vs "local RPMs that will be built".

bin/build
simaishi: `-v` doesn't take relative path and it will fail to mount.

bin/build
Fryguy: Oh I see...you are concerned about killing the local RPMs.  I can understand that.

bin/build
Fryguy: If `realpath` is available, maybe that's preferable, but I'm good with this for now.

simaishi: I wasn't sure what command works on different platforms (linux, osx, etc), so I went for $PWD...

Fryguy: yeah realpath isn't in mac by default, but it's a simple brew command away.

Fryguy:
LGTM.  Will approve when the WIP is remove since it has your personal copr in there still.

images/manageiq-base/Dockerfile
```diff
@@ -36,6 +37,9 @@ LABEL name="manageiq-base" \
       io.k8s.description="ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures." \
       io.openshift.tags="ManageIQ,miq,manageiq"
 
+COPY rpms/* /tmp/rpms/
```
Fryguy, line 40: I'm ok with merging as is for now, but can we prefer a volume mount instead?  That way, we don't get the RPMs in one image layer only to remove them in a later image layer.

images/manageiq-base/container-assets/create_local_yum_repo.sh
```diff
@@ -0,0 +1,15 @@
...
+
+cat > /etc/yum.repos.d/local_rpm.repo << EOF
+[local-rpm]
+baseurl=file:///tmp/rpms/$basearch
+name=Local yum repo
+enabled=1
+gpgcheck=0
+EOF
+
+dnf config-manager --setopt=manageiq-11-kasparov*.exclude=manageiq-* --save
```
Fryguy, line 15: Should manageiq be hardcoded here?

simaishi: Right now, everything is hard-coded to use manageiq prefix. I'll open an issue to make it customizable.

carbonin: :confused: I mean ... it's in all the rpm names. If it needs to be configurable then we're going to need to change a lot more than this line.

images/manageiq-base/Dockerfile
carbonin: Can we do this all the RUN command then? I think that would solve the problem, right?

carbonin: Oh, we need them in the other images .... Nevermind.

images/manageiq-base/container-assets/create_local_yum_repo.sh
simaishi: https://github.com/ManageIQ/manageiq-pods/issues/521

images/manageiq-base/Dockerfile
simaishi: I definitely want to get rid of adding RPMs in base and removing in ui. If you can think of a better way than duplicating that part of the codes, let me know!

simaishi: I think volume mount isn't supported during build?

images/manageiq-base/container-assets/create_local_yum_repo.sh
simaishi: Actually....

That will need to stay hardcoded because we're trying to avoid RPMs coming from manageiq yum repo. When we add an option to use a different yum repo (#522), that new repo needs to be added here, so it will also be excluded.

Fryguy:
Merging...we can do the other 2 things in followups if we need to.

simaishi:
Jansa backport details (replaced yum repo with 10-jansa):

```
$ git log -1
commit 45d8ea7691c1e89ba9023e121edfc8c7a6a28643
Date:   Fri May 29 14:19:51 2020 -0400

    Merge pull request #491 from simaishi/local_rpm

    Support for local RPMs

    (cherry picked from commit 904772e6a5929f965414ec301d39e4b733103264)
```

PR #492, opened 2020-05-08
Add OIDC config to allow for API access

carbonin:
This is the operator version of https://github.com/ManageIQ/manageiq-appliance/pull/282

cc @jvlcek @abellotti 

Going to leave this as WIP until all the details are worked out in https://github.com/ManageIQ/manageiq-appliance/pull/282

This should get us full UI and API support for OIDC auth.

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
+++ b/manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
@@ -223,22 +223,34 @@ func httpdAuthLoginFormConf() string {
 
 func httpdAuthApplicationAPIConf() string {
 	return `
-<LocationMatch ^/api>
-    SetEnvIf Authorization     '^Basic +YWRtaW46' let_admin_in
-    SetEnvIf X-Auth-Token      '^.+$'             let_api_token_in
-    SetEnvIf X-MIQ-Token       '^.+$'             let_sys_token_in
+<LocationMatch ^/api(?!\/(v[\d\.]+\/)?product_info$)>
+  SetEnvIf Authorization '^Basic +YWRtaW46' let_admin_in
+  SetEnvIf X-Auth-Token  '^.+$'             let_api_token_in
+  SetEnvIf X-MIQ-Token   '^.+$'             let_sys_token_in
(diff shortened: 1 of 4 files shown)


manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
```diff
@@ -162,30 +202,37 @@ LoadModule auth_mellon_module modules/mod_auth_mellon.so
...
 OIDCClientSecret             ${HTTPD_AUTH_OIDC_CLIENT_SECRET}
-
 OIDCRedirectURI              "https://%s/oidc_login/redirect_uri"
+OIDCCryptoPassphrase         sp-secret
 OIDCOAuthRemoteUserClaim     username
 
-OIDCCryptoPassphrase         sp-secret
+OIDCOAuthClientID                  ${HTTPD_AUTH_OIDC_CLIENT_ID}
+OIDCOAuthClientSecret              ${HTTPD_AUTH_OIDC_CLIENT_SECRET}
+OIDCOAuthIntrospectionEndpoint     %s
```
Fryguy, line 226: Typically the introspection endpoint is derived from the provider url, so there would be no reason to specify it normally.  That is, I would expect that the value would be left blank, and then only if specified would we have this line.

carbonin: I think that's essentially what @abellotti is saying in https://github.com/ManageIQ/manageiq-appliance/pull/282#issuecomment-625236558 so I was going to wait to see what direction they go.

If we want to make this optional in the non-entire-conf case I can do what he's suggesting I think.

Fryguy: Yeah, I think I'd like a sensible default if not specified based on what @abellotti wrote there, as opposed to requiring it to be specified.

simaishi:
Jansa backport details:

```
$ git log -1
commit d468fc78ff55cd617fffadeee4b45ddd16a995f7
Date:   Tue May 12 16:13:15 2020 -0400

    Merge pull request #492 from carbonin/new_oidc_config

    Add OIDC config to allow for API access

    (cherry picked from commit 4b9f64f2108621c230e8f18a2686374b22a1498d)
```

PR #493, opened 2020-05-12
Remove extra sha256 sum when generating the encryption key

carbonin:
This was present in the [original implementation]
(https://github.com/ManageIQ/manageiq-password/blob/f85e888fe60433390d0a1d5145889edc67437fdf/lib/manageiq/password.rb#L210-L213)
but was only needed because that code allowed for user-provided
passwords to be used as the encryption key. In our implementation
we are only using random bytes as the input so an extra sha sum
doesn't do anything to improve security.

Additionally this was triggering an issue in some static security
analysis tools.

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/generators.go
+++ b/manageiq-operator/pkg/helpers/miq-components/generators.go
@@ -2,7 +2,6 @@ package miqtools
 
 import (
 	"crypto/rand"
-	"crypto/sha256"
 	"encoding/base64"
 	"encoding/hex"
 )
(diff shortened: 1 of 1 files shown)


simaishi:
Jansa backport details:

```
$ git log -1
commit 1dad635d9be8d59fa020ff1814150ba7090b5578
Date:   Tue May 12 12:59:44 2020 -0400

    Merge pull request #493 from carbonin/remove_bonus_sha_sum

    Remove extra sha256 sum when generating the encryption key

    (cherry picked from commit 1b8b37b45c3b7ac85e70752c7aeb206e74977508)
```

PR #500, opened 2020-05-13
Remove ability to set the admin user password

### LINKED ISSUE #499: Do not expose password in CR
@miq-bot add_label bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

> According to Kubernetes official doc: https://kubernetes.io/docs/concepts/configuration/secret/
> 
> Kubernetes Secrets let you store and manage sensitive information, such as passwords, OAuth tokens, and ssh keys. Storing confidential information in a Secret is safer and more flexible than putting it verbatim in a Pod definition or in a container image.
> 
> Secret Design Doc: https://git.k8s.io/community/contributors/design-proposals/auth/secrets.md
> 
> According to: https://blog.aquasec.com/managing-kubernetes-secrets
> 
> Placing sensitive info into a secret object does not automatically make it secure. By default, data in Kubernetes secrets is stored in Base64 encoding, which is practically the same as plaintext.
> However, secrets give you more control over access and usage of passwords, keys, etc. Kubernetes can either mount secrets separately from the pods that use them, or save them as environment variables.

We are now putting password in CR as https://github.com/ManageIQ/manageiq-pods/blob/master/manageiq-operator/deploy/crds/manageiq.org_v1alpha1_manageiq_cr.yaml#L6 , this is not safe and it is better also use a secret as well.

FYI @carbonin @Fryguy @chessbyte

carbonin:
The application has logic to configure the initial user and
also allows for changing that user's password. The ability to set
the password from multiple places is error-prone and confusing

Fixes #499

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ Examine output and point your web browser to the reported URL/HOST.
 
 ### Logging In
 
-Per the ManageIQ project [basic configuration](http://manageiq.org/docs/get-started/basic-configuration) documentation, you can now login to the MIQ web interface using the default username (`admin`) and either the default password (`smartvm`) of the password configured using the `APPLICATION_ADMIN_PASSWORD` parameter.
+Per the ManageIQ project [basic configuration](http://manageiq.org/docs/get-started/basic-configuration) documentation, you can now login to the MIQ web interface using the default username and password (`admin`/`smartvm`).
 
 ## Backup and restore of the MIQ database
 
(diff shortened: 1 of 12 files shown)


simaishi:
Jansa backport details:

```
$ git log -1
commit 16b120adf2d6cd8a08319dacfcf1af52e4be181b
Date:   Wed May 13 16:11:25 2020 -0400

    Merge pull request #500 from carbonin/remove_admin_password

    Remove ability to set the admin user password

    (cherry picked from commit df47af8842092a252dba40a4cdf03d9d6e3e8603)
```

PR #505, opened 2020-05-14
Disable admin user API access when enableApplicationLocalLogin is false

### LINKED ISSUE #501: enableApplicationLocalLogin should prevent admin user via API
When enableApplicationLocalLogin is set to false, the expectation is that regular db users (including admin) cannot log into the application via UI or API.  At present, in appliances, this only applies to the UI, since it drives showing the user/pass textboxes.

Since the operator is responsible for the OIDC configuration, I think it should remove these 2 lines [[1]] [[2]], when enableApplicationLocalLogin=false.  Additionally, I think there needs to be a validation condition if someone chooses httpdMode=internal and enableApplicationLocalLogin=false, because otherwise they can't login at all.

[1]: https://github.com/ManageIQ/manageiq-pods/blob/df47af8842092a252dba40a4cdf03d9d6e3e8603/manageiq-operator/pkg/helpers/miq-components/httpd_conf.go#L281
[2]: https://github.com/ManageIQ/manageiq-pods/blob/df47af8842092a252dba40a4cdf03d9d6e3e8603/manageiq-operator/pkg/helpers/miq-components/httpd_conf.go#L290

carbonin:
After this change, any request to the API with basic auth (including
the admin user name) will be rejected by httpd with a 401 return code.

Fixes #501

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
+++ b/manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
@@ -63,11 +63,11 @@ Options SymLinksIfOwnerMatch
 func httpdAuthenticationConf(spec *miqv1alpha1.ManageIQSpec) string {
 	switch spec.HttpdAuthenticationType {
 	case "openid-connect":
-		return httpdOIDCAuthConf(spec.OIDCProviderURL, spec.OIDCOAuthIntrospectionURL, spec.ApplicationDomain)
+		return httpdOIDCAuthConf(spec.OIDCProviderURL, spec.OIDCOAuthIntrospectionURL, spec.ApplicationDomain, *spec.EnableApplicationLocalLogin)
 	case "external":
-		return httpdExternalAuthConf()
+		return httpdExternalAuthConf(*spec.EnableApplicationLocalLogin)
 	case "active-directory":
-		return httpdADAuthConf()
+		return httpdADAuthConf(*spec.EnableApplicationLocalLogin)
 	case "saml":
 		return httpdSAMLAuthConf()
 	default:
@@ -75,7 +75,7 @@ func httpdAuthenticationConf(spec *miqv1alpha1.ManageIQSpec) string {
 	}
 }
 
-func httpdExternalAuthConf() string {
+func httpdExternalAuthConf(enableLocalLogin bool) string {
 	s := `
 %s
 
@@ -111,13 +111,13 @@ func httpdExternalAuthConf() string {
 		s,
 		httpdAuthLoadModulesConf(),
 		httpdAuthLoginFormConf(),
-		httpdAuthApplicationAPIConf("Basic", "\"External Authentication (httpd) for API\"", apiExtraConfig),
+		httpdAuthApplicationAPIConf("Basic", "\"External Authentication (httpd) for API\"", apiExtraConfig, enableLocalLogin),
 		httpdAuthLookupUserDetailsConf(),
 		httpdAuthRemoteUserConf(),
 	)
 }
 
-func httpdADAuthConf() string {
+func httpdADAuthConf(enableLocalLogin bool) string {
 	s := `
 %s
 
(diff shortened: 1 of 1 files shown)


manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
```diff
@@ -63,19 +63,19 @@ Options SymLinksIfOwnerMatch
 func httpdAuthenticationConf(spec *miqv1alpha1.ManageIQSpec) string {
 	switch spec.HttpdAuthenticationType {
 	case "openid-connect":
-		return httpdOIDCAuthConf(spec.OIDCProviderURL, spec.OIDCOAuthIntrospectionURL, spec.ApplicationDomain)
+		return httpdOIDCAuthConf(spec.OIDCProviderURL, spec.OIDCOAuthIntrospectionURL, spec.ApplicationDomain, *spec.EnableApplicationLocalLogin)
```
Fryguy, line 66: Not for this PR, but would it make more sense to just pass the spec object in it's entirety and then the methods pull out what they need?

manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
```diff
@@ -287,17 +286,23 @@ func httpdAuthApplicationAPIConf(authType, authName, extraConfig string) string
...
+  %s
 
   %s
 </LocationMatch>
 `
+	letAdminIn := ""
+	if enableLocalLogin {
+		letAdminIn = `
+  SetEnvIf Authorization '^Basic +YWRtaW46' let_admin_in
+  Allow from env=let_admin_in`
```
Fryguy, line 302: Does order matter with respect to these settings (I know it does for some other settings in the conf).  This PR changes the order, and I wonder if that's a problem or not.

carbonin: Pretty sure it doesn't. @jvlcek can confirm though

manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
carbonin: Yeah, maybe when we need one more parameter. Originally this method only needed one or two so it made sense to pass them individually. I feel like 5 is my tipping point

manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
jvlcek: > Does order matter with respect to these settings (I know it does for some other settings in the conf). This PR changes the order, and I wonder if that's a problem or not.

Order does not seem to matter. I just successfully tested with the following segment and it worked fine

```
<LocationMatch ^/api(?!\/(v[\d\.]+\/)?product_info$)>
  SetEnvIf X-Auth-Token  '^.+$'                 let_api_token_in
  SetEnvIf X-MIQ-Token   '^.+$'                 let_sys_token_in
  SetEnvIf X-CSRF-Token  '^.+$'                 let_csrf_token_in

  AuthType     oauth20
  AuthName     "External Authentication (oidc) for API"

  Require   valid-user
  Order          Allow,Deny
  Allow from env=let_admin_in
  Allow from env=let_api_token_in
  SetEnvIf Authorization '^Basic +YWRtaW46'     let_admin_in
  Allow from env=let_sys_token_in
  Allow from env=let_csrf_token_in
  Satisfy Any
</LocationMatch>
```

simaishi:
Jansa backport details:

```
$ git log -1
commit 7cbfa931b0b5fe3c523f9e236e787a1965cd4329
Date:   Fri May 15 12:24:46 2020 -0400

    Merge pull request #505 from carbonin/enable_local_login_api

    Disable admin user API access when enableApplicationLocalLogin is false

    (cherry picked from commit 3e0699ed3f4d530504f235c2593286fa2bacdb11)
```

PR #510, opened 2020-05-18
[IVANCHUK] Change the httpd image to httpd-init

carbonin:
After the image rename, the deploy will fail with the non-init httpd
image.


Fryguy:
@simaishi Please review.

simaishi:
Are we safe using `latest` tag in this image?

carbonin:
Actually I think I need some more changes to make this work. We also changed the port httpd exposes. Marking this as WIP until I can test the whole thing out.

simaishi:
All other port "80" still used in this repo are ok to stay as is? I'm not familiar with the port change, so I just grep'ed for it...

carbonin:
Yeah, I tested the new templates and everything deployed fine.

Issue #488, opened 2020-05-06
Enable customer can set storageclass in CR


gyliu513:
@miq-bot add_label bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

If I do not have a default storageclass, then some pods like kafka, postgres etc will keeps pending due to pvc cannot be created.

```console
oc describe pvc kafka-data
Name:          kafka-data
Namespace:     bluecf
StorageClass:
Status:        Pending
Volume:
Labels:        app=inframgmtinstall
Annotations:   <none>
Finalizers:    [kubernetes.io/pvc-protection]
Capacity:
Access Modes:
VolumeMode:    Filesystem
Mounted By:    kafka-7655bb6865-697qt
Events:
  Type    Reason         Age                 From                         Message
  ----    ------         ----                ----                         -------
  Normal  FailedBinding  8s (x13 over 3m2s)  persistentvolume-controller  no persistent volumes available for this claim and no storage class is set
```

**What did you expect to happen:**

it is better to introduce some new fields for those pvcs so that customer can set storageclass manually if there is no default storageclass.

FYI @carbonin 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- ManageIQ version:
- Minikube/KIND/OCP version:
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

gyliu513:
@Fryguy @carbonin seems the `VolumeSource` in Kubernetes does not support `StorageClass` based on API here https://github.com/kubernetes/api/blob/master/core/v1/types.go#L51-L159 , any comments for this?

carbonin:
I don't understand why this would matter. The storage class is a property of the PVC. In this case we would set the storage class on the claim as described in https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1. I'm not sure how a different `VolumeSource` comes into the discussion.

gyliu513:
@carbonin if you check https://github.com/ManageIQ/manageiq-pods/blob/master/manageiq-operator/pkg/helpers/miq-components/kafka.go#L219-L226 , you will see we are using `volumes` in a `pod` spec, and we cannot set `storageclass` in a `volume` spec.

carbonin:
[...]

PR #511, opened 2020-05-18
enable setting storageclass

gyliu513:

Fixes https://github.com/ManageIQ/manageiq-pods/issues/488

### DIFF
--- a/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
+++ b/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
@@ -201,6 +201,9 @@ spec:
             postgresqlSharedBuffers:
               description: 'PostgreSQL shared buffers setting (default: 1GB)'
               type: string
+            storageClassName:
+              description: StorageClass name that will be used by manageiq data stores
+              type: string
             tlsSecret:
               description: 'Secret containing the tls cert and key for the ingress,
                 content generated if not provided (default: tls-secret)'
--- a/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
+++ b/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
@@ -3,8 +3,9 @@ package v1alpha1
 import (
 	"errors"
 	"fmt"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 // ManageIQSpec defines the desired state of ManageIQ
@@ -37,6 +38,10 @@ type ManageIQSpec struct {
 	// Secret containing the image registry authentication information needed for the manageiq images
 	// +optional
 	ImagePullSecret string `json:"imagePullSecret"`
+
+	// StorageClass name that will be used by manageiq data stores
+	// +optional
+	StorageClassName string `json:"storageClassName"`
 
 	// Image namespace used for the httpd deployment (default: manageiq)
 	// Note: the exact image will be determined by the authentication method selected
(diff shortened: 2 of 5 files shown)


manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -38,6 +39,10 @@ type ManageIQSpec struct {
 	// +optional
 	ImagePullSecret string `json:"imagePullSecret"`
 
+	// StorageClass name that will be used by manageiq data stores
+	// +optional
+	StorageClassName string `json:"storageClassName"`
```
carbonin, line 44: Do you think we will need separate storage classes for the separate PVCs? I'm not sure what the difference between classes would be.

manageiq-operator/pkg/helpers/miq-components/kafka.go
```diff
@@ -42,13 +42,34 @@ func KafkaPVC(cr *miqv1alpha1.ManageIQ) *corev1.PersistentVolumeClaim {
...
+	}
+
 	return &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "kafka-data",
 			Namespace: cr.ObjectMeta.Namespace,
 			Labels:    labels,
 		},
 		Spec: corev1.PersistentVolumeClaimSpec{
+			StorageClassName: cr.Spec.StorageClassName,
```
carbonin, line 72: Can we deduplicate these definitions? Maybe define the pvc then add the storage class name to the spec if it's defined in the CR?

We do something similar for the [httpd service account](https://github.com/ManageIQ/manageiq-pods/blob/3e0699ed3f4d530504f235c2593286fa2bacdb11/manageiq-operator/pkg/helpers/miq-components/httpd.go#L310) to avoid defining an entire extra deployment.

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
gyliu513: Like `ImagePullSecret`, I want one cluster use one `StorageClassName` for the ManageIQ Operator.

manageiq-operator/pkg/helpers/miq-components/kafka.go
gyliu513: Yes, make sense, let me update.

carbonin:
@Fryguy @gyliu513 Does this need to go back to `jansa`?

manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
```diff
@@ -201,6 +201,9 @@ spec:
             postgresqlSharedBuffers:
               description: 'PostgreSQL shared buffers setting (default: 1GB)'
               type: string
+            storageClassName:
```
Fryguy, line 204: Is this meant for all storage things, or just postgres?  If the latter we should probably prefix.

carbonin: All PVCs. I had a similar comment in https://github.com/ManageIQ/manageiq-pods/pull/511#discussion_r426686836.

We just only happen to have PVCs for PG and kafka/zookeeper.

carbonin:
This should have went back to Jansa.

simaishi:
Jansa backport details:

```
$ git log -1
commit d7cb191de6e99b14013d40cf9a55399b4d2e9280
Date:   Mon May 18 11:56:54 2020 -0400

    Merge pull request #511 from gyliu513/sc

    enable setting storageclass

    (cherry picked from commit c8cc5b2c79b02ac5326c54c170871488341be3cd)
```

PR #513, opened 2020-05-18
Update issue templates

Fryguy:
I find the current issue templates wordy, and also put too much "stuff" into the actual issue.  These templates instead use comments to show the user what to do, but won't present themselves in the final issue, making it cleaner for readers (i.e. committers/mergers)

@carbonin Please review.


chessbyte:
Looks good to me!

PR #515, opened 2020-05-20
Add a CR parameter for an initial group name

### LINKED ISSUE #508: Create an initial group when using external auth
If a user configures OIDC, and also disables local login they won't be able to access the application unless they happen to have one of our pre-seeded groups configured in their auth server.

It would probably be better if we let the user provide us with an initial group name that we create at first boot. The user can then initially log in as a user in that group to configure the rest of the app.

Not sure of what needs to be done here for RBAC so that the initial group can create other groups and whatever else we want them to be able to do on that first login.

carbonin:
When provided, we will create a group with the given name with
the super administrator role. The main use case would be when
a user wants to deploy with basic auth totally disabled as soon
as the application comes up. This means that they need at least one
group in the application that matches a group in their identity
provider.

We also now run a full seed after migrating the database on initial
deploy so that we can create the initial group correctly.

Fixes #508

### DIFF
--- a/images/manageiq-orchestrator/container-assets/entrypoint
+++ b/images/manageiq-orchestrator/container-assets/entrypoint
@@ -43,6 +43,13 @@ EOS
   bin/rails runner "$script"
 }
 
+function create_initial_group() {
+  if [ "$ADMIN_GROUP" != "" ]; then
+    echo "== Creating initial group "$ADMIN_GROUP" =="
+    bin/rails runner "MiqGroup.create!(:miq_user_role => MiqUserRole.find_by(:name => 'EvmRole-super_administrator'), :description => ENV['ADMIN_GROUP'], :tenant => Tenant.root_tenant)"
+  fi
+}
+
 check_svc_status ${MEMCACHED_SERVICE_HOST} ${MEMCACHED_SERVICE_PORT}
 check_svc_status ${DATABASE_HOSTNAME} ${DATABASE_PORT}
 
@@ -66,7 +73,8 @@ case $? in
     pushd ${APP_ROOT}
       set -e
       REGION=${DATABASE_REGION} bin/rake db:migrate
-      REGION=${DATABASE_REGION} bin/rails runner "MiqDatabase.seed; MiqRegion.seed"
+      REGION=${DATABASE_REGION} bin/rake db:seed
+      create_initial_group
     popd
   ;;
   4) # new_replica
@@ -88,5 +96,7 @@ case $? in
     exit 1
 esac
 
-update_auth_settings
+pushd ${APP_ROOT}
+  update_auth_settings
+popd
 exec ruby /var/www/miq/vmdb/lib/workers/bin/evm_server.rb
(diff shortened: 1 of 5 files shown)


images/manageiq-orchestrator/container-assets/entrypoint
```diff
@@ -43,6 +43,13 @@ EOS
   bin/rails runner "$script"
 }
 
+function create_initial_group() {
+  if [ "$ADMIN_GROUP" != "" ]; then
+    echo "== Creating initial group "$ADMIN_GROUP" =="
+    bin/rails runner "MiqGroup.create!(:miq_user_role => MiqUserRole.find_by(:name => 'EvmRole-super_administrator'), :description => ENV['ADMIN_GROUP'], :tenant => Tenant.root_tenant)"
+  fi
+}
```
abellotti, line 51: is DB always empty/just seeded when this happens ?

carbonin: Oh, right this isn't idempotent. Let me fix.

Fryguy: LOL you just wanted to say "idempotent", didn't you?

images/manageiq-orchestrator/container-assets/entrypoint
```diff
@@ -66,7 +73,8 @@ case $? in
     pushd ${APP_ROOT}
       set -e
       REGION=${DATABASE_REGION} bin/rake db:migrate
-      REGION=${DATABASE_REGION} bin/rails runner "MiqDatabase.seed; MiqRegion.seed"
+      REGION=${DATABASE_REGION} bin/rake evm:db:seed
```
Fryguy, line 76: What is the side effect of switching to this full seed from the near-primordial seed?

Fryguy: Also should we prefer `rake db:seed` here?

images/manageiq-orchestrator/container-assets/entrypoint
carbonin: Maybe .... :grin:

images/manageiq-orchestrator/container-assets/entrypoint
carbonin: > What is the side effect of switching to this full seed from the near-primordial seed?

I wasn't sure what we needed to add a group correctly, so I just went for everything. I figured that we would need `MiqGroup`, `MiqUserRole`, `Tenant` at least, but couldn't be sure of what else and was running into weird issues (but a successful add) when I didn't have the right things seeded.

> Also should we prefer rake db:seed here?

Oh, they're the same .. Sure? Why do we have two?

Fryguy: > Why do we have two?

I honestly can't remember 😅

carbonin:
Nope, @Fryguy I fixed that by moving the group creation to only during the "initial deploy" condition. That way it will only happen once.

Fryguy:
Ohhh....I see...it's the in "Starting New Deployment" section... I was concerned about redeployment, but didn't notice from the diff that it wasn't in there.

simaishi:
Jansa backport details:

```
$ git log -1
commit f961fb1696a20508839bdd17a71c8c45d4383c86
Date:   Mon Jun 1 08:17:54 2020 -0400

    Merge pull request #515 from carbonin/add_initial_admin_group

    Add a CR parameter for an initial group name

    (cherry picked from commit 5c6729313ca8e945f98dd4348f4b6d6067e48207)
```

PR #523, opened 2020-05-29
Allow rpm prefix to be changed

### LINKED ISSUE #521: Allow customized rpm name to be used
Currently, rpm names are hardcoded to use `manageiq` prefix. Need to allow customization.

@miq-bot add-label enhancement

simaishi:
Fixes #521

@sethrpeterson

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -17,6 +17,7 @@ OPERATOR_DIR=${OPERATOR_DIR:-manageiq-operator}
 
 RPM_BUILD_OPTIONS=${RPM_BUILD_OPTIONS:-""}
 RPM_BUILD_IMAGE=${RPM_BUILD_IMAGE:-"manageiq/rpm_build"}
+RPM_PREFIX=${RPM_PREFIX:-"manageiq"}
 
 while getopts "t:d:r:bnhlp" opt; do
   case $opt in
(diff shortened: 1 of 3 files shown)


bin/build
```diff
@@ -83,12 +85,20 @@ pushd $IMAGE_DIR
...
     $cmd
   done
+
+  cmd="docker build --tag $REPO/manageiq-ui-worker:$TAG \
+                    --build-arg FROM_REPO=$REPO \
+                    --build-arg FROM_TAG=$TAG \
+                    --build-arg RPM_PREFIX=$RPM_PREFIX \
+                    manageiq-ui-worker"
+  echo "Building manageiq-ui-worker: $cmd"
+  $cmd
```
Fryguy, line 101: would it be simpler to just add the arg to every image build and keep it in the `for` loop?

simaishi: I wasn't sure if passing an argument when not needed is a good idea (that generates a warning), so I went for splitting. That's definitely easier to read though.

simaishi:
Jansa backport details:

```
$ git log -1
commit fb079a90575c1e510412cd09c2434c4580a9017a
Date:   Fri May 29 16:46:15 2020 -0400

    Merge pull request #523 from simaishi/customize_rpm_name

    Allow rpm prefix to be changed

    (cherry picked from commit 60de2294e22cb500c5e61e4aabbffbe6bd471762)
```

PR #509, opened 2020-05-18
Add Power support on travis-ci

Siddhesh-Ghadi:
**What this PR does / why we need it:** Add travis jobs for ppc64le architecture.

**Special notes for your reviewer:**
- Build passes: https://travis-ci.org/github/Siddhesh-Ghadi/manageiq-pods/builds/688274723

Issue #489, opened 2020-05-06
Backup and Restore in the Operator


carbonin:
We should implement application backup and restore as a part of the operator.
I'm not sure of the best-practices around this kind of thing, but I was thinking a separate controller (with separate CRDs) that would watch for ManageIQBackup/ManageIQRestore objects.
Those would represent either a schedule to backup which we would turn into a CronJob object or a one-off request to restore from a particular backup.

I was also thinking we should go farther than just database backup for this, but that can come as a second pass. For example it would be convenient to store the encryption key with the backup so that a new environment could be created quickly, but that might also be a security concern.

Additionally I would like to be able to utilize this for users who may want to move from the appliance-based deployment to the container-based one so whatever format we choose should be easily created by the appliance (say, a tarball that includes the database backup as well as the encryption key and maybe some other metadata).

I think this would be the last gap between the operator and the templates.

For Backup/Restore using Velero:
- [X] https://github.com/ManageIQ/manageiq-pods/pull/626 Annotate volumes that should be backed up
- [x] https://github.com/ManageIQ/manageiq-pods/pull/619 Allow using a selector to backup a smaller set of only essential resources.

Fryguy:
> a tarball that includes the database backup as well as the encryption key and maybe some other metadata

I believe this is what the current backup from the appliance console does, so it would be really cool if we could enable backup via appliance console, then restore via this new controller

carbonin:
> it would be really cool if we could enable backup via appliance console, then restore via this new controller

Yeah, that's what I was thinking.

carbonin:
Based on the [etcd operator](https://github.com/operator-framework/community-operators/blob/master/community-operators/etcd/0.9.2/etcdoperator.v0.9.2.clusterserviceversion.yaml#L79-L111) it looks like the way to go for this would be to represent a single backup request in as a CR then report on the status of that backup in the CR status field.

Then everything needed to complete the backup should be provided in the CR.

carbonin:
Also of note: when looking at the etcd operator it also looks like they [deploy multiple containers in the operator pod](https://github.com/operator-framework/community-operators/blob/master/community-operators/etcd/0.9.4/etcdoperator.v0.9.4.clusterserviceversion.yaml#L210-L223) to handle different functions like backup and restore. I'm not sure what the utility of that is vs separate controllers in the same operator process, but that's probably worth investigating.

bdunne:
For Backup and restore using Velero:
[...]

PR #539, opened 2020-06-05
Test things in rspec

### LINKED ISSUE #448: Verify generated files as a part of CI tests
We should run `operator-sdk generate csv --csv-version 0.0.1 --update-crds` and `operator-sdk generate k8s` as a part of CI tests and ensure nothing changes.

This will prevent us from merging source that should have changed the CRD or only changing one CRD when both should have changed.

bdunne:
- Convert the commands in the "go" Travis jobs into rspec tests reducing the number of jobs that need to be run.
- Install the operator-sdk in the test suite
- Test that regenerating the CRDs doesn't change any tracked files
- Regerate the CRDs to fix the spec failure (adding missing `type: object`)

Fixes #448


spec/manageiq-operator/crd_spec.rb
```diff
@@ -0,0 +1,10 @@
+require 'awesome_spawn'
+describe "Generate CRDs" do
+  it "doesn't change any tracked files" do
+    AwesomeSpawn.run!("operator-sdk generate k8s", :chdir => ROOT.join("manageiq-operator"))
+    AwesomeSpawn.run!("operator-sdk generate crds", :chdir => ROOT.join("manageiq-operator"))
+    AwesomeSpawn.run!("operator-sdk generate csv --update-crds --csv-version=0.0.1", :chdir => ROOT.join("manageiq-operator"))
```
carbonin, line 6: This version string might be problematic once we start actually releasing versions, but I don't think we're there yet. Just something to consider.

bdunne: Is `manageiq-operator/version/version.go:4: Version = "0.0.1"` a safe place to get the version from?

Fryguy:
I'm not sure on this one.  Ultimately, I'd expect the templates to go away, and we'd be left with the operator, which is in Go.  The RSpec tests are just there because we needed something to validate the YAML, but it can honestly be in any language.  I think the better direction would be to write all the tests in Go, and removing RSpec.

We don't currently have any tests in Go aside from ensuring the build works, but we probably should add them, particularly for some of the more interesting logic in the operator.

Fryguy:
If we do go this direction, then you can completely change the .travis.yml to collapse the multiple sections.  We only had multiple sections because you can't include the language in the matrix and we had rvm and go.

I think something like this will work:

```yaml
language: ruby
rvm:
- 2.5.7
- 2.6.5
arch:
- amd64
- ppc64le
cache:
  bundler: true
before_script:
- eval "$(gimme 1.14)"
- curl -L https://github.com/operator-framework/operator-sdk/releases/download/v0.15.1/operator-sdk-v0.15.1-x86_64-linux-gnu -o /home/travis/bin/operator-sdk && chmod +x /home/travis/bin/operator-sdk
script:
- bundle exec rspec
```

.travis.yml
```diff
@@ -4,48 +4,39 @@ jobs:
...
+  - eval "$(gimme 1.14)"
+  - curl -L https://github.com/operator-framework/operator-sdk/releases/download/v0.15.1/operator-sdk-v0.15.1-x86_64-linux-gnu -o /home/travis/bin/operator-sdk && chmod +x /home/travis/bin/operator-sdk
   script:
   - bundle exec rspec
-- language: go
-  go: stable
-  cache:
-    directories:
-    - $HOME/.cache/go-build
-    - $HOME/gopath/pkg/mod
```
Fryguy, line 20: Did the caching get lost?

carbonin:
> We don't currently have any tests in Go aside from ensuring the build works, but we probably should add them, particularly for some of the more interesting logic in the operator.

Yeah, this is a big hole right now. We really should be looking into go test frameworks (or just use the built-in [testing](https://golang.org/pkg/testing/) package) and write some unit tests. So the question is then: do we need to get this in for now to test the file generation, or can it wait until we have time to investigate moving to go tests entirely.

A related question is also: should we remove the templates in master now?
I think now is a good time. We could keep the backup job templates in a utils directory or something until we solve https://github.com/ManageIQ/manageiq-pods/issues/489

carbonin:
Also, to be clear, I didn't intend the fix for this issue to be implementing a bunch of go tests, but maybe that just means we shouldn't collapse the existing tests into rspec.

I would say it's fine to test the file generation stuff in rspec for now though.

spec/manageiq-operator/crd_spec.rb
carbonin: I'm not actually sure. The operator might be versioned separately from the csv, but I haven't learned enough about either yet to say for sure. My guess would be that they would be the same version, but I also don't love the idea of parsing the version out of source to run this test.

I would say that this is fine for now.

carbonin: > The CSV version is the same as the Operator’s, and a new CSV is generated when upgrading Operator versions.

ref: https://docs.openshift.com/container-platform/4.4/operators/operator_sdk/osdk-generating-csvs.html

bdunne: Done.

.travis.yml
bdunne: Fixed

bdunne:
> I'm not sure on this one. Ultimately, I'd expect the templates to go away, and we'd be left with the operator, which is in Go. The RSpec tests are just there because we needed something to validate the YAML, but it can honestly be in any language. I think the better direction would be to write all the tests in Go, and removing RSpec.
> 
> We don't currently have any tests in Go aside from ensuring the build works, but we probably should add them, particularly for some of the more interesting logic in the operator.

Yeah, I spent a bunch of time debating this myself, but I think it's better to have it in rspec than a bunch of shell commands.  Once we write some go tests and pick a test framework we can switch over, but this didn't seem like the time to do that since we still don't have any go tests.

spec/support/go_mod.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def parse
+    return if @parsed
+    parse_block = nil
+
+    path.read.each_line do |line|
+      if line.start_with?("go ")
+        parse_go_version(line)
+        next
+      elsif line.start_with?("require (")
```
carbonin, line 18: Is the space required here? Should that be an optional space?

spec/support/go_mod.rb
```diff
@@ -0,0 +1,42 @@
...
+        parse_block = :require
+      elsif line.start_with?(")")
```
carbonin, line 20: Same here, could we have spaces before the paren which would break this?

spec/support/go_mod.rb
bdunne: Removed as discussed

spec/support/go_mod.rb
bdunne: Removed as discussed

simaishi:
@bdunne @Fryguy backport conflicts as jansa branch travis doesn't run for pcc64le (#509). Is this something we should add to jansa? If not, I'll just exclude ppc64le when backporting.

bdunne:
I think it should include PPC.  Do you want me to make a jansa PR?

simaishi:
@bdunne Please create a PR if it's not just about backporting https://github.com/ManageIQ/manageiq-pods/pull/509 and this PR.

simaishi:
Jansa backport details:

```
$ git log -1
commit 97661ead7ad3fe837628f247178e30e0b81b1a61
Date:   Wed Jun 10 08:27:51 2020 -0400

    Merge pull request #539 from bdunne/test_things

    Test things in rspec

    (cherry picked from commit 4b5eb15af45867db073524d60c3f1f943b148bb3)
```

PR #543, opened 2020-06-10
Add a flag to gate deploying kafka and zookeeper

### LINKED ISSUE #541: Should we remove kafka and zookeeper from Jansa?
Kafka and zookeeper add a significant overhead to the deployed application. Not only do they both require a PV, but they are also resource hungry (an idle Kafka uses 500Mi of memory) and the dependencies that we are least familiar with maintaining.

Additionally, I believe they are only used for writing events currently (correct me if I'm wrong @agrare @Fryguy). If no other part of the application is reading those events, and we don't expect a user to read from our Kafka instance, it feels like we should remove them for this release.

carbonin:
The application doesn't actually rely on these services so we don't
need to deploy them by default.

If a user wants events published to kafka they can still enable the
flag on deploy and consume those events from a separate application.

Requires https://github.com/ManageIQ/manageiq/pull/20268
Fixes #541

### DIFF
--- a/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
+++ b/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
@@ -48,6 +48,10 @@ spec:
             databaseVolumeCapacity:
               description: 'Database volume size (default: 15Gi)'
               type: string
+            deployMessagingService:
+              description: 'Flag to indicate if Kafka and Zookeeper should be deployed
+                (default: false)'
+              type: boolean
             enableApplicationLocalLogin:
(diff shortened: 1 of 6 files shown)


simaishi:
Jansa backport details:

```
$ git log -1
commit c44a92c3836c3af7c3745ac6953f34460b2b546b
Date:   Wed Jun 10 14:39:41 2020 -0400

    Merge pull request #543 from carbonin/gate_kafka_install

    Add a flag to gate deploying kafka and zookeeper

    (cherry picked from commit 1e9c2d46f80a06d7cd2d609006249776eca2fae4)
```

PR #545, opened 2020-06-16
Explicitly set the image pull policy to IfNotPresent

carbonin:
Similar to https://github.com/ManageIQ/manageiq/pull/20279

### DIFF
--- a/manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq-operator.v0.0.1.clusterserviceversion.yaml
+++ b/manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq-operator.v0.0.1.clusterserviceversion.yaml
@@ -58,7 +58,7 @@ spec:
                 - name: OPERATOR_NAME
                   value: manageiq-operator
                 image: docker.io/manageiq/manageiq-operator:v0.0.1
-                imagePullPolicy: Always
+                imagePullPolicy: IfNotPresent
                 name: manageiq-operator
                 resources:
                   limits:
--- a/manageiq-operator/deploy/operator.yaml
+++ b/manageiq-operator/deploy/operator.yaml
@@ -18,7 +18,7 @@ spec:
           image: docker.io/manageiq/manageiq-operator:v0.0.1
           command:
           - manageiq-operator
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
           env:
             - name: WATCH_NAMESPACE
               valueFrom:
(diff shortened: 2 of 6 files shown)


manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq-operator.v0.0.1.clusterserviceversion.yaml
```diff
@@ -58,7 +58,7 @@ spec:
                 - name: OPERATOR_NAME
                   value: manageiq-operator
                 image: docker.io/manageiq/manageiq-operator:v0.0.1
-                imagePullPolicy: Always
+                imagePullPolicy: IfNotPresent
```
bdunne, line 61: Why are we switching from `Always` to `IfNotPresent`?

carbonin: I think whenever you're using a specific tag it should probably not be `Always`.

Also `Always` here wasn't a conscious decision, I believe it was just the default from the `operator-sdk new`

simaishi:
Backported without changes to `manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq-operator.v0.0.1.clusterserviceversion.yaml` as https://github.com/ManageIQ/manageiq-pods/pull/549 changed the values for jansa branch.

Jansa backport details:

```
$ git log -1
commit e17540ffa3afc30ef2f1ceaf69a6b6d572ab90ef
Date:   Tue Jun 16 12:04:22 2020 -0400

    Merge pull request #545 from carbonin/explicitly_set_pull_policy

    Explicitly set the image pull policy to IfNotPresent

    (cherry picked from commit 938a7cde01483cf4a11fde1fd28dc73df2f20358)
```

carbonin:
So for this I feel like we're in a weird state. This PR used IfNotPresent because at this time the file was pinned to a released version (granted, one that didn't exist).

But when we moved to `jansa-latest` I feel like it might make more sense to use `Always` as that's an intentionally moving tag.

That all said, should we be pinning the operator deploy to a particular release? For example, when `jansa-1` is tagged, should we use that? Then we would go back to `IfNotPresent`.

I think we need to make a decision on what tag should be in this file for each of our branches before we can really call this resolved. @Fryguy @bdunne ?

bdunne:
Totally agree, we should either pull "always" if using the "latest" tag or pull "ifNotPresent" with a release tag.  I don't think anything else makes sense.

I think it's fine for a release branch to point to a release tag and the tip of the branch can be ahead of the release, README.md already has instructions for changing it for development purposes.

simaishi:
So the jansa branch is good as is for now, right?  "latest-jansa" with "Always" is what it currently has.

I did discuss with @Fryguy about possibly changing to an actual release tag only for tags, which is what we do with Gemfile.

bdunne:
> So the jansa branch is good as is for now, right? "latest-jansa" with "Always" is what it currently has.

I think it's wrong now, but I think we need to make a decision about the tags first.

Also, it's not consistent within the branch.  I marked https://github.com/ManageIQ/manageiq-pods/pull/539 as `jansa/yes?` to ensure tests verify consistency.

simaishi:
https://github.com/ManageIQ/manageiq-pods/pull/549 is a direct PR to jansa branch and that was merged later than this PR. So when I backported this PR, I made sure I don't override #549.

I'm confused and not sure what the right values are at this point, so please create a PR to jansa branch as needed.

PR #546, opened 2020-06-17
Switch to latest since v0.0.1 tag doesn't exist yet

bdunne:
Pull always since we're using the latest tag which is not static

### DIFF
--- a/manageiq-operator/README.md
+++ b/manageiq-operator/README.md
@@ -29,7 +29,7 @@ $ operator-sdk build docker.io/example/manageiq-operator:latest
 2. Update the operator deployment with the new image:
 
 ```bash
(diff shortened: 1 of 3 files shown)


carbonin:
I think we probably need a jansa-specific PR to keep this "IfNotPresent" and `latest-jansa` for that branch.

carbonin:
Or maybe wait until we have an actually released jansa branch?

I'm not really sure what to do for this.

bdunne:
I think the pull policy needs to reflect the tag.  If it's a moving tag (latest), the pull policy should be `Always`.  Once we have a static release tag, the pull policy should be `IfNotPresent`

carbonin:
Okay, this looks good to me then. I'll set jansa/no just to make sure we do something different there.

Fryguy:
Marking this as backported since techically the change came over, just in a different form

PR #547, opened 2020-06-17
Change Postgresql component to support CreateOrUpdate

### LINKED ISSUE #438: Update existing objects in the reconcile loop

mali-chainzee:
handles updating postpresql resources 

Applies to #438

### DIFF
--- a/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
+++ b/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
@@ -231,32 +231,41 @@ func (r *ReconcileManageIQ) generateMemcachedResources(cr *miqv1alpha1.ManageIQ)
 }
 
 func (r *ReconcileManageIQ) generatePostgresqlResources(cr *miqv1alpha1.ManageIQ) error {
-	postgresqlSecret := miqtool.DefaultPostgresqlSecret(cr)
-	if err := r.createk8sResIfNotExist(cr, postgresqlSecret, &corev1.Secret{}); err != nil {
+	secret := miqtool.DefaultPostgresqlSecret(cr)
+	if err := r.createk8sResIfNotExist(cr, secret, &corev1.Secret{}); err != nil {
 		return err
 	}
 
-	postgresqlConfigsConfigMap := miqtool.NewPostgresqlConfigsConfigMap(cr)
-	if err := r.createk8sResIfNotExist(cr, postgresqlConfigsConfigMap, &corev1.ConfigMap{}); err != nil {
+	configMap, mutateFunc := miqtool.NewPostgresqlConfigsConfigMap(cr)
+	if result, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, configMap, mutateFunc); err != nil {
 		return err
+	} else {
+		logger.Info("ConfigMap has been reconciled", "component", "postgresql", "result", result)
 	}
 
-	postgresqlPVC := miqtool.NewPostgresqlPVC(cr)
-	if err := r.createk8sResIfNotExist(cr, postgresqlPVC, &corev1.PersistentVolumeClaim{}); err != nil {
+	pvc, mutateFunc := miqtool.NewPostgresqlPVC(cr)
+	if result, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, pvc, mutateFunc); err != nil {
 		return err
+	} else {
+		logger.Info("Service has been reconciled", "component", "postgresql", "result", result)
 	}
 
-	postgresqlService := miqtool.NewPostgresqlService(cr)
-	if err := r.createk8sResIfNotExist(cr, postgresqlService, &corev1.Service{}); err != nil {
+	service, mutateFunc := miqtool.NewPostgresqlService(cr)
+	if result, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, service, mutateFunc); err != nil {
 		return err
+	} else {
+		logger.Info("Service has been reconciled", "component", "postgresql", "result", result)
 	}
 
-	postgresqlDeployment, err := miqtool.NewPostgresqlDeployment(cr)
+	deployment, mutateFunc, err := miqtool.NewPostgresqlDeployment(cr, r.scheme)
 	if err != nil {
 		return err
 	}
-	if err := r.createk8sResIfNotExist(cr, postgresqlDeployment, &appsv1.Deployment{}); err != nil {
+
+	if result, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, deployment, mutateFunc); err != nil {
 		return err
+	} else {
+		logger.Info("Deployment has been reconciled", "component", "postgresql", "result", result)
 	}
 
 	return nil
--- a/manageiq-operator/pkg/helpers/miq-components/postgresql.go
+++ b/manageiq-operator/pkg/helpers/miq-components/postgresql.go
@@ -6,7 +6,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	resource "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	intstr "k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 func DefaultPostgresqlSecret(cr *miqv1alpha1.ManageIQ) *corev1.Secret {
@@ -40,42 +42,44 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
 	return secretName
 }
 
-func NewPostgresqlConfigsConfigMap(cr *miqv1alpha1.ManageIQ) *corev1.ConfigMap {
+func NewPostgresqlConfigsConfigMap(cr *miqv1alpha1.ManageIQ) (*corev1.ConfigMap, controllerutil.MutateFn) {
 	labels := map[string]string{
 		"app": cr.Spec.AppName,
 	}
-	return &corev1.ConfigMap{
+	data := map[string]string{
+		"01_miq_overrides.conf": postgresqlOverrideConf(),
+	}
+	configMap := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "postgresql-configs",
 			Namespace: cr.ObjectMeta.Namespace,
-			Labels:    labels,
-		},
-		Data: map[string]string{
-			"01_miq_overrides.conf": postgresqlOverrideConf(),
 		},
 	}
+
+	return configMap, configMapMutateFn(configMap, labels, data)
 }
 
-func NewPostgresqlPVC(cr *miqv1alpha1.ManageIQ) *corev1.PersistentVolumeClaim {
+func NewPostgresqlPVC(cr *miqv1alpha1.ManageIQ) (*corev1.PersistentVolumeClaim, controllerutil.MutateFn) {
 	labels := map[string]string{
 		"app": cr.Spec.AppName,
 	}
+
 	storageReq, _ := resource.ParseQuantity(cr.Spec.DatabaseVolumeCapacity)
+
+	resources := corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			"storage": storageReq,
+		},
+	}
+
+	accessModes := []corev1.PersistentVolumeAccessMode{
+		"ReadWriteOnce",
+	}
+
 	pvc := &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "postgresql",
 			Namespace: cr.ObjectMeta.Namespace,
-			Labels:    labels,
-		},
-		Spec: corev1.PersistentVolumeClaimSpec{
-			AccessModes: []corev1.PersistentVolumeAccessMode{
-				"ReadWriteOnce",
-			},
-			Resources: corev1.ResourceRequirements{
-				Requests: corev1.ResourceList{
-					"storage": storageReq,
-				},
-			},
 		},
 	}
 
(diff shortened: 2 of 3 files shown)


carbonin:
@mali-chainzee This doesn't fix 438 so can you remove that from the commit and PR description so that the issue doesn't close when we merge this?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,84 +44,93 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
 		Spec: corev1.PersistentVolumeClaimSpec{
-			AccessModes: []corev1.PersistentVolumeAccessMode{
-				"ReadWriteOnce",
-			},
-			Resources: corev1.ResourceRequirements{
-				Requests: corev1.ResourceList{
-					"storage": storageReq,
-				},
-			},
+			StorageClassName: &cr.Spec.StorageClassName,
```
carbonin, line 91: Why change the logic for this? I would keep the conditional as it was previously.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -7,28 +7,32 @@ import (
...
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      postgresqlSecretName(cr),
 			Namespace: cr.ObjectMeta.Namespace,
 			Labels:    labels,
 		},
-		StringData: secret,
+		StringData: stringData,
 	}
+
+	return secret
```
carbonin, line 35: The logic here doesn't seem to have changed. Please only change what is necessary for the fix.

Also this was originally mentioned in https://github.com/ManageIQ/manageiq-pods/pull/441#discussion_r435219686

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -174,64 +187,78 @@ func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, erro
...
+		return nil, nil, err
+	}
 	var repNum int32 = 1
 
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "postgresql",
 			Namespace: cr.ObjectMeta.Namespace,
-			Labels:    deploymentLabels,
+			Labels:    podLabels,
```
carbonin, line 242: Why change this?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -174,64 +187,78 @@ func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, erro
...
 		},
 		Spec: appsv1.DeploymentSpec{
-			Strategy: appsv1.DeploymentStrategy{
-				Type: "Recreate",
-			},
-			Replicas: &repNum,
 			Selector: &metav1.LabelSelector{
 				MatchLabels: podLabels,
```
carbonin, line 246: I think we should be ensuring that this is correct in the mutate function. Can you model this after [the way we did it for memcached](https://github.com/ManageIQ/manageiq-pods/blob/0e3e06e2ac81923b8fb27064b1eef40b599d1d54/manageiq-operator/pkg/helpers/miq-components/memcached.go#L81)?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
carbonin: Additionally it should be in the mutate function, not in the initial object.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,84 +44,93 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
 
-func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, error) {
+func postgresqlContainer(cr *miqv1alpha1.ManageIQ) corev1.Container {
 	var initialDelaySecs int32 = 60
-	container := corev1.Container{
+	return corev1.Container{
 		Name:            "postgresql",
 		Image:           cr.Spec.PostgresqlImageName + ":" + cr.Spec.PostgresqlImageTag,
 		ImagePullPolicy: corev1.PullIfNotPresent,
+
```
carbonin, line 133: Bonus space?

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -52,3 +53,42 @@ func addResourceReqs(memLimit, memReq, cpuLimit, cpuReq string, c *corev1.Contai
...
+	mutateFn := func() error {
+		configMap.ObjectMeta.Labels = labels
+		if len(data) > 0 {
+			configMap.Data = data
+		}
+		return nil
+	}
+
+	return mutateFn
+
```
carbonin, line 80: Bonus space not needed.

Even if gofmt says it's okay I would prefer this type of style be consistent. At least it should be the same in each source file.

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -52,3 +53,42 @@ func addResourceReqs(memLimit, memReq, cpuLimit, cpuReq string, c *corev1.Contai
...
+	mutateFn := func() error {
+		pvc.ObjectMeta.Labels = labels
+		pvc.Spec.AccessModes = accessModes
+		pvc.Spec.Resources = resources
+		return nil
+	}
+
+	return mutateFn
+
```
carbonin, line 93: Same here, no space after a return.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -231,31 +231,31 @@ func (r *ReconcileManageIQ) generateMemcachedResources(cr *miqv1alpha1.ManageIQ)
...
-	if err := r.createk8sResIfNotExist(cr, postgresqlSecret, &corev1.Secret{}); err != nil {
+	secret := miqtool.DefaultPostgresqlSecret(cr)
+	if err := r.createk8sResIfNotExist(cr, secret, &corev1.Secret{}); err != nil {
 		return err
 	}
 
-	postgresqlConfigsConfigMap := miqtool.NewPostgresqlConfigsConfigMap(cr)
-	if err := r.createk8sResIfNotExist(cr, postgresqlConfigsConfigMap, &corev1.ConfigMap{}); err != nil {
+	configMap, configMapMutateF := miqtool.NewPostgresqlConfigsConfigMap(cr)
+	if _, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, configMap, configMapMutateF); err != nil {
```
carbonin, line 240: Can you log the result of the call [as we do for memcached](https://github.com/ManageIQ/manageiq-pods/blob/0e3e06e2ac81923b8fb27064b1eef40b599d1d54/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go#L217-L228)?

(Applies to all of the `CreateOrUpdate` calls in this function)

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -6,13 +6,16 @@ import (
...
+	"k8s.io/apimachinery/pkg/runtime"
 	intstr "k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 func DefaultPostgresqlSecret(cr *miqv1alpha1.ManageIQ) *corev1.Secret {
 	labels := map[string]string{
 		"app": cr.Spec.AppName,
 	}
+
```
carbonin, line 18: Bonus whitespace not needed.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -29,6 +32,7 @@ func DefaultPostgresqlSecret(cr *miqv1alpha1.ManageIQ) *corev1.Secret {
 		},
 		StringData: secret,
 	}
+
```
carbonin, line 35: Same here.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,80 +44,94 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
-	return &corev1.ConfigMap{
+
+	data := map[string]string{
+		"01_miq_overrides.conf": postgresqlOverrideConf(),
+	}
+
+	configMap := &corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "postgresql-configs",
+			Name:      "miq-pg-configs",
```
carbonin, line 58: Why change the name?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,80 +44,94 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
+	}
+
+	accessModes := []corev1.PersistentVolumeAccessMode{
+		"ReadWriteOnce",
+	}
+
 	pvc := &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "postgresql",
+			Name:      "miq-pgdb-volume",
```
carbonin, line 87: Same, why change the name?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -175,63 +193,63 @@ func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, erro
 		},
 	}
 
-	err := addResourceReqs(cr.Spec.PostgresqlMemoryLimit, cr.Spec.PostgresqlMemoryRequest, cr.Spec.PostgresqlCpuLimit, cr.Spec.PostgresqlCpuRequest, &container)
+	err := addResourceReqs(cr.Spec.PostgresqlMemoryLimit, cr.Spec.PostgresqlMemoryRequest, cr.Spec.PostgresqlCpuRequest, cr.Spec.PostgresqlCpuLimit, &container)
```
carbonin, line 196: This shouldn't have been changed I think.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -175,63 +193,63 @@ func NewPostgresqlDeployment(cr *miqv1alpha1.ManageIQ) (*appsv1.Deployment, erro
...
-	return deployment, nil
+	f := func() error {
+		if err := controllerutil.SetControllerReference(cr, deployment, scheme); err != nil {
+			return err
+		}
+		if deployment.ObjectMeta.Labels == nil {
+			deployment.ObjectMeta.Labels = make(map[string]string)
+		}
+		deployment.ObjectMeta.Labels["app"] = cr.Spec.AppName
+		deployment.Spec = spec
```
carbonin, line 251: I think this will also run into https://github.com/ManageIQ/manageiq-pods/issues/550 as it is currently, but we can fix that separately.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,80 +44,94 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
-				corev1.ServicePort{
-					Name: "postgresql",
-					Port: port,
-				},
-			},
-			Selector: selector,
 		},
 	}
+
+	mutateFn := serviceMutateFn(service, labels, ports, selector)
```
carbonin, line 124: I think you can just inline this into the return value if the function doesn't return an error, no need for an extra variable.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,80 +44,94 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
-			},
 		},
 	}
 
 	if cr.Spec.StorageClassName != "" {
 		pvc.Spec.StorageClassName = &cr.Spec.StorageClassName
 	}
 
-	return pvc
+	mutateFn := pvcMutateFn(pvc, labels, accessModes, resources)
```
carbonin, line 96: I think you can just inline this into the return value if the function doesn't return an error, no need for an extra variable.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -40,80 +44,94 @@ func postgresqlSecretName(cr *miqv1alpha1.ManageIQ) string {
...
+			Name:      "miq-pg-configs",
 			Namespace: cr.ObjectMeta.Namespace,
-			Labels:    labels,
-		},
-		Data: map[string]string{
-			"01_miq_overrides.conf": postgresqlOverrideConf(),
 		},
 	}
+
+	mutateFn := configMapMutateFn(configMap, labels, data)
```
carbonin, line 63: I think you can just inline this into the return value if the function doesn't return an error, no need for an extra variable.

manageiq-operator/pkg/helpers/miq-components/postgresql.go
Fryguy: Also -1 for using miq hardcoded because then it prevents alternate naming.

PR #549, opened 2020-06-17
[JANSA] Switch to latest-jansa since v0.0.1 tag doesn't exist yet

bdunne:
Pull always since we're using the latest tag which is not static

Based on https://github.com/ManageIQ/manageiq-pods/pull/546#issuecomment-645419573


Fryguy:
cc @simaishi

PR #558, opened 2020-06-19
Add an override for the worker images deployed by the orchestrator

### LINKED ISSUE #557: Allow the user to specify image tags/digests for the worker images
Right now we can configure the image for all the "operands" with parameters to the CR, but the workers deployed by the orchestrator are always bound to the orchestrator's tag.

This is a problem if the orchestrator is tagged differently or pulled using a sha so we should allow users to specify image strings for the `manageiq-base-worker`, `manageiq-webserver-worker` and `manageiq-ui-worker` images through the operator.

This will also require some changes on the manageiq side to pull the env vars and use them when deploying the workers.

cc @Fryguy

carbonin:
Previously a user was not able to change the actual image name for
the worker images nor were they able to deploy with anything other
than the namespace and tag deployed for the orchestrator.

This gives a bit more flexability in that one image can be replaced
at a time and also allows a user to specify an image ref with a
digest rather than a tag

Should be backported only if https://github.com/ManageIQ/manageiq/pull/20297 is backported.

Fixes #557 
cc @Fryguy

### DIFF
--- a/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
+++ b/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
@@ -38,6 +38,10 @@ spec:
               description: Domain name for the external route. Used for external authentication
                 configuration
               type: string
+            baseWorkerImage:
+              description: Image string used for the base worker deployments By default
+                this is determined by the orchestrator pod
+              type: string
             databaseRegion:
               description: 'Database region number (default: 0)'
               type: string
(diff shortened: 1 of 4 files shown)


manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
```diff
@@ -227,6 +231,14 @@ spec:
               description: 'Secret containing the tls cert and key for the ingress,
                 content generated if not provided (default: tls-secret)'
               type: string
+            uiWorkerImage:
+              description: Image string used for the UI worker deployments By default
+                this is determined by the orchestrator pod
+              type: string
+            webserverWorkerImage:
```
bdunne, line 238: What do you think of changing this to `apiWorkerImage`?  In the core repo it is currently `webservice`.  It is API not httpd, correct?  I find the name confusing because the UI and httpd are also webservers and I suspect others will too.

carbonin: This is used for both API and remote console workers. I just named them after the actual image names as it's image specific, not necessarily worker specific.

simaishi:
Jansa backport details:

```
$ git log -1
commit cc607f971d86f1483593bf940e6ac06f4f224899
Date:   Tue Jun 23 11:30:17 2020 -0400

    Merge pull request #558 from carbonin/add_worker_image_overrides

    Add an override for the worker images deployed by the orchestrator

    (cherry picked from commit 2dd6c413f43f8808d50f01afddf9f63ecc3b1ba9)
```

PR #562, opened 2020-06-23
Document customizing secrets for Postgres and messaging

### LINKED ISSUE #452: Document and validate secret format
When a user wants to provide their own secrets, they need to know the format of the secret as they only specify the secret name in the CR. This applies to the tls, postgres, and kafka secrets.

I'm not sure where we should do this (README, CRD, elsewhere?), but we already have [secret templates](https://github.com/ManageIQ/manageiq-pods/blob/ca7f33f82d77de006cd0bf6a02e660d1915da82c/templates/app/postgresql-secrets.yaml) for pg and kafka, and tls secrets can be created using `oc create secret tls tls-secret --cert=tls.crt --key=tls.key`.

Additionally, we should validate the secret fields if a secret name was provided and the secret was not created by the operator.

Originally raised in https://github.com/ManageIQ/manageiq-pods/pull/451#discussion_r409553809

bdunne:
Fixes #452

### DIFF
--- a/manageiq-operator/README.md
+++ b/manageiq-operator/README.md
@@ -2,6 +2,12 @@
 
 This operator manages the lifecycle of ManageIQ application on a OCP4 cluster.
 
+## Customizing the installation
+
+ManageIQ can be run with an external Postgres or messaging server.  To do so, please create the required OpenShift secret(s) with the correct parameters using the template(s) for [Postgres](templates/app/postgresql-secrets.yaml) or [messaging](/templates/app/kafka-secrets.yaml) and provide those secret names as `databaseSecret` and/or `kafkaSecret` in `manageiq.org_v1alpha1_manageiq_cr.yaml`.
+
+If you want to use a custom TLS certificate, it can be created with `oc create secret tls tls-secret --cert=tls.crt --key=tls.key` and setting the secret name as `tlsSecret` in `manageiq.org_v1alpha1_manageiq_cr.yaml`.
+
 ## Run Operator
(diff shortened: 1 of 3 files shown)


manageiq-operator/README.md
```diff
@@ -2,6 +2,10 @@
 
 This operator manages the lifecycle of ManageIQ application on a OCP4 cluster.
 
+## Customizing the installation
+
+ManageIQ can be run with an external Postgres or messaging server.  To do so, please create the required OpenShift secret(s) with the correct parameters using the template(s) for [Postgres](templates/app/postgresql-secrets.yaml) or [messaging](/templates/app/kafka-secrets.yaml).  Then provide those secret names
```
carbonin, line 7: Can you reference the parameter names that should be set in the CR here?

carbonin:
Also, can we call out how to create the tls secret here? I think that's probably the most important part as users can only get a signed cert if they create it themselves.

templates/app/kafka-secrets.yaml
```diff
@@ -15,11 +15,15 @@ objects:
     username: "${MESSAGING_USERNAME}"
 parameters:
 - name: APP_NAME
+  description: The name of the application that ManageIQ will be deployed as in OpenShift.  Used to group OpenShift objects.
```
carbonin, line 18: Can you remove the `ManageIQ` name from these comments as we try to keep that configurable.
(Hence APP_NAME :trollface: )

PR #566, opened 2020-06-30
Add reconcile functions to Kafka and Zookeeper

bdunne:
Allow the Kafka and Zookeeper objects to be updated

Addresses part of https://github.com/ManageIQ/manageiq-pods/issues/438


carbonin:
Except for the spec failures :laughing: 

I think you're missing a controllerutil import in `kafka.go`

Issue #512, opened 2020-05-18
Fetch the OIDC introspection endpoint on deploy


carbonin:
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
Implement https://github.com/ManageIQ/manageiq-appliance_console/pull/121 in the operator.

abellotti:
Jansa specific PR: https://github.com/ManageIQ/manageiq-pods/pull/582

PR #571, opened 2020-07-02
Enhancement to optionally fetch the OIDCAuthIntrospectionURL

abellotti:
- Enhancement to optionally fetch the OIDCAuthIntrospectionURL from the Provider Metadata URL if not provided to us.

Solves: https://github.com/ManageIQ/manageiq-pods/issues/512

### DIFF
--- a/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
+++ b/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
@@ -74,7 +74,10 @@ type ManageIQSpec struct {
 	// +optional
 	OIDCCACertSecret string `json:"oidcCaCertSecret"`
 	// URL for OIDC authentication introspection
-	// Only used with the openid-connect authentication type
+	// Only used with the openid-connect authentication type.
+	// If not specified, the operator will attempt to fetch its value from the
+	// "token_introspection_endpoint" field in the Provider metadata at the
+	// OIDCProviderURL provided.
 	// +optional
 	OIDCOAuthIntrospectionURL string `json:"oidcAuthIntrospectionURL"`
 	// Secret name containing the OIDC client id and secret
--- a/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
+++ b/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
@@ -1,8 +1,11 @@
 package v1alpha1
 
 import (
+	"crypto/tls"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
 	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -287,6 +290,28 @@ func init() {
 	SchemeBuilder.Register(&ManageIQ{}, &ManageIQList{})
 }
 
+func fetchIntrospectionUrl(providerUrl string) (introspectionUrl string, errMsg string) {
+	introspectionUrl = ""
+	errMsg = ""
+
+	if providerUrl != "" {
+		customTransport := http.DefaultTransport.(*http.Transport).Clone()
+		customTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+		client := &http.Client{Transport: customTransport}
+
+		resp, err := client.Get(providerUrl)
+		if err != nil {
+			errMsg = fmt.Sprintf("Could not query the OIDCProviderURL %s - error: %s", providerUrl, err)
+		} else {
+			var result map[string]interface{}
+			json.NewDecoder(resp.Body).Decode(&result)
+			introspectionUrl = result["token_introspection_endpoint"].(string)
+			resp.Body.Close()
+		}
+	}
+	return
+}
+
 func (m *ManageIQ) Initialize() {
 	spec := &m.Spec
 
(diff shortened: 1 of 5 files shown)


manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -408,9 +433,25 @@ func (m *ManageIQ) Validate() error {
...
 			// Invalid if config and any other info is also provided
 			errs = append(errs, "OIDCProviderURL, OIDCOAuthIntrospectionURL, and OIDCClientSecret are invalid when HttpdAuthConfig is specified")
-		} else if spec.HttpdAuthConfig == "" && (spec.OIDCProviderURL == "" || spec.OIDCOAuthIntrospectionURL == "" || spec.OIDCClientSecret == "") {
-			// Need to provide either the entire config or a secret and provider url
-			errs = append(errs, "HttpdAuthConfig or all of OIDCProviderURL, OIDCOAuthIntrospectionURL, and OIDCClientSecret must be provided for openid-connect authentication")
+		} else if spec.HttpdAuthConfig == "" {
+			if spec.OIDCProviderURL == "" || spec.OIDCClientSecret == "" {
+				errs = append(errs, "HttpdAuthConfig or both OIDCProviderURL and OIDCClientSecret must be provided for openid-connect authentication")
+			} else {
+				// If the OAuth Introspection URL was not specified, let's try to fetch it from the Provider URL
```
carbonin, line 440: I don't really think this belongs in `Validate`.

My intention here was to make this act kind of like an AR validation where we wouldn't change the object we're validating and wouldn't take any further action.

I think it would be better to make the introspection URL optional and fetch it if it was not provided. Then if we fail to fetch it at that point we can error out of reconcile.

manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go
```diff
@@ -283,6 +286,28 @@ func init() {
 	SchemeBuilder.Register(&ManageIQ{}, &ManageIQList{})
 }
 
+func fetchIntrospectionUrl(providerUrl string) (introspectionUrl string, errMsg string) {
```
carbonin, line 289: This doesn't belong in the `v1alpha1` package. I think this all should go at least into the controller, possibly into the httpd component.

chessbyte:
@abellotti @carbonin let's get this one over the finish line, so we can call https://github.com/ManageIQ/manageiq/issues/19867 complete for Jansa (I removed the lower priority nice-to-haves from the list on that issue).

carbonin:
Alberto and I chatted about the direction this should go last week. That new direction would have conflicted with https://github.com/ManageIQ/manageiq-pods/pull/572, but since that was merged today we should be able to make more progress here.

Fryguy:
I'm assuming that #572 is jansa/no, though, right?  So will it still have issues on backport?

carbonin:
Ah, yeah. Will probably need a separate jansa-only PR as I did with #579 and #580

abellotti:
The Jansa PR: https://github.com/ManageIQ/manageiq-pods/pull/582

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -81,7 +84,16 @@ func Ingress(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*extensionsv1bet
...
 
-func HttpdConfigMap(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*corev1.ConfigMap, controllerutil.MutateFn) {
+func HttpdConfigMap(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*corev1.ConfigMap, controllerutil.MutateFn, error) {
+
+	if cr.Spec.HttpdAuthenticationType == "openid-connect" && cr.Spec.OIDCProviderURL != "" && cr.Spec.OIDCOAuthIntrospectionURL == "" {
+		introspectionURL, err := fetchIntrospectionUrl(cr.Spec.OIDCProviderURL)
+		if err != nil {
+			return nil, nil, err
+		}
+		(&cr.Spec).OIDCOAuthIntrospectionURL = introspectionURL
```
carbonin, line 94: You don't need the & and parens here. `cr.Spec.OIDCOAuthIntrospectionURL = introspectionURL` will do what you want as `cr` is already a pointer.

I assume you lifted this from [`Initialize`](https://github.com/ManageIQ/manageiq-pods/blob/26539c7eed15b4d07ff215e1416ee21be34cff1f/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go#L290-L295)?

If that's the case that's different because we're using an intermediate local to store the spec so we don't need `m.Spec` everywhere.

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -534,3 +546,32 @@ func tlsSecretName(cr *miqv1alpha1.ManageIQ) string {
...
+	}
+
+	var result map[string]interface{}
+	err = json.NewDecoder(resp.Body).Decode(&result)
+	if err != nil {
+		return "", fmt.Errorf("%s - %s", errMsg, err)
+	}
+
+	if result["token_introspection_endpoint"] == nil {
+		return "", fmt.Errorf("%s - token_introspection_endpoint is missing", errMsg)
```
carbonin, line 573: Maybe "token_introspection_endpoint is missing **from response**"?
I don't think I would understand what this was saying as is.

abellotti:
ok @carbonin I `think` I got 'em.

simaishi:
Backported to jansa via https://github.com/ManageIQ/manageiq-pods/pull/582

PR #572, opened 2020-07-02
Allow httpd Services, Deployment and ConfigMap to be reconciled

bdunne:
Applies to #438

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/httpd.go
+++ b/manageiq-operator/pkg/helpers/miq-components/httpd.go
@@ -8,97 +8,121 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	intstr "k8s.io/apimachinery/pkg/util/intstr"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
-func HttpdServiceAccount(cr *miqv1alpha1.ManageIQ) *corev1.ServiceAccount {
-	return &corev1.ServiceAccount{
+func HttpdServiceAccount(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*corev1.ServiceAccount, controllerutil.MutateFn) {
+	serviceAccount := &corev1.ServiceAccount{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      cr.Spec.AppName + "-httpd",
 			Namespace: cr.ObjectMeta.Namespace,
 		},
 	}
-}
 
-func NewIngress(cr *miqv1alpha1.ManageIQ) *extensionsv1beta1.Ingress {
-
-	labels := map[string]string{
-		"app": cr.Spec.AppName,
+	f := func() error {
+		if err := controllerutil.SetControllerReference(cr, serviceAccount, scheme); err != nil {
+			return err
+		}
+		return nil
 	}
 
-	return &extensionsv1beta1.Ingress{
+	return serviceAccount, f
+}
+
+func Ingress(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*extensionsv1beta1.Ingress, controllerutil.MutateFn) {
+	ingress := &extensionsv1beta1.Ingress{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "httpd",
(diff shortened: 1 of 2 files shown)


manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -278,197 +303,195 @@ func initializeHttpdContainer(spec *miqv1alpha1.ManageIQSpec, privileged bool, c
...
-							Name: "httpd-config",
-							VolumeSource: corev1.VolumeSource{
-								ConfigMap: &corev1.ConfigMapVolumeSource{
-									LocalObjectReference: corev1.LocalObjectReference{Name: "httpd-configs"},
-								},
-							},
-						},
-					},
 				},
+				Spec: corev1.PodSpec{},
```
carbonin, line 337: Same with this, why add it here?

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -278,197 +303,195 @@ func initializeHttpdContainer(spec *miqv1alpha1.ManageIQSpec, privileged bool, c
...
-		"app": cr.Spec.AppName,
-	}
-	selector := map[string]string{
-		"service": "web-service",
+	f := func() error {
+		if err := controllerutil.SetControllerReference(cr, service, scheme); err != nil {
+			return err
+		}
+		addAppLabel(cr.Spec.AppName, &service.ObjectMeta)
+		service.Spec.Ports[0].Name = "ui-service-3000"
```
carbonin, line 388: Same as https://github.com/ManageIQ/manageiq-pods/pull/577/files#r450876638, I think we need to check for nil as that's the array zero value.

I'm not sure if the `Get` in `CreateOrUpdate` will overwrite the Ports array, but it's probably safer to remove it from the spec above and `make` a new one here if it's nil

carbonin: This really applies to all of the other services as well. I'll only write it here though.

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -278,197 +303,195 @@ func initializeHttpdContainer(spec *miqv1alpha1.ManageIQSpec, privileged bool, c
...
+				Spec: corev1.PodSpec{},
 			},
 		},
 	}
 
-	configureHttpdAuth(&cr.Spec, &deployment.Spec.Template.Spec)
-
 	// Only assign the service account if we need additional privileges
 	if privileged {
 		deployment.Spec.Template.Spec.ServiceAccountName = cr.Spec.AppName + "-httpd"
```
carbonin, line 344: This should be in the mutate function, right?

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -8,97 +8,122 @@ import (
...
-					},
-					SecretName: tlsSecretName(cr),
+		Spec: extensionsv1beta1.IngressSpec{},
+	}
+
+	f := func() error {
+		if err := controllerutil.SetControllerReference(cr, ingress, scheme); err != nil {
+			return err
+		}
+		ingress.Spec.TLS = []extensionsv1beta1.IngressTLS{
```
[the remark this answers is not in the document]
bdunne: Done

manageiq-operator/pkg/helpers/miq-components/httpd.go
bdunne: Done

manageiq-operator/pkg/helpers/miq-components/httpd.go
bdunne: Done

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -278,197 +302,195 @@ func initializeHttpdContainer(spec *miqv1alpha1.ManageIQSpec, privileged bool, c
...
+	f := func() error {
+		if err := controllerutil.SetControllerReference(cr, service, scheme); err != nil {
+			return err
+		}
+		addAppLabel(cr.Spec.AppName, &service.ObjectMeta)
+		if len(service.Spec.Ports) == 0 {
+			service.Spec.Ports = append(service.Spec.Ports, corev1.ServicePort{})
+		}
+		service.Spec.Ports[0].Name = "remote-console-3000"
+		service.Spec.Ports[0].Port = 3000
```
carbonin, line 438: :+1:

bdunne:
> This actually fixes #438, right?

Yes it does

PR #575, opened 2020-07-02
Use the CR's UID rather than generating a new one for the orchestrator

### LINKED ISSUE #573: Constant updates to the orchestrator prevent the app from deploying
When the operator is running it constantly updates the orchestrator deployment causing it to redeploy every second or so.
This prevents the app from ever starting up.

I can see near-constant logging from the operator indicating updates to the orchestrator deployment:
```
{"level":"info","ts":1593721407.695829,"logger":"controller_manageiq","msg":"Deployment has been reconciled","component":"orchestrator","result":"updated"}
{"level":"info","ts":1593721408.1546383,"logger":"controller_manageiq","msg":"Deployment has been reconciled","component":"orchestrator","result":"updated"}
{"level":"info","ts":1593721408.8235636,"logger":"controller_manageiq","msg":"Deployment has been reconciled","component":"orchestrator","result":"updated"}
{"level":"info","ts":1593721409.396162,"logger":"controller_manageiq","msg":"Deployment has been reconciled","component":"orchestrator","result":"updated"}
{"level":"info","ts":1593721410.0562649,"logger":"controller_manageiq","msg":"Deployment has been reconciled","component":"orchestrator","result":"updated"}
{"level":"info","ts":1593721410.6785092,"logger":"controller_manageiq","msg":"Deployment has been reconciled","component":"orchestrator","result":"updated"}
{"level":"info","ts":1593721411.3213987,"logger":"controller_manageiq","msg":"Deployment has been reconciled","component":"orchestrator","result":"updated"}
```

Deploy the app with the operator

bdunne:
Since we only have one server, it should be static.
Also, prevent the reconcile from generating a new GUID on every run.

Fixes #573

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/orchestrator.go
+++ b/manageiq-operator/pkg/helpers/miq-components/orchestrator.go
@@ -10,8 +10,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"strconv"
 	"strings"
-
-	"github.com/google/uuid"
 )
 
 func OrchestratorServiceAccount(cr *miqv1alpha1.ManageIQ) *corev1.ServiceAccount {
@@ -192,7 +190,7 @@ func OrchestratorDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*
 			},
 			corev1.EnvVar{
 				Name:  "GUID",
-				Value: uuid.New().String(),
+				Value: string(cr.GetUID()),
 			},
 			corev1.EnvVar{
 				Name:  "DATABASE_REGION",


manageiq-operator/pkg/helpers/miq-components/orchestrator.go
```diff
@@ -192,7 +190,7 @@ func OrchestratorDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*
 			},
 			corev1.EnvVar{
 				Name:  "GUID",
-				Value: uuid.New().String(),
+				Value: string(cr.GetUID()),
```
carbonin, line 193: What is this value? Is this something openshift assigns?

Fryguy: All objects in kubernetes get a UID assigned by kube (in the metadata), which is a UUID, and is not modifyable by end-users.

PR #577, opened 2020-07-06
Prevent the reconcile loop from updating the services on every run

bdunne:


manageiq-operator/pkg/helpers/miq-components/memcached.go
```diff
@@ -101,19 +101,18 @@ func NewMemcachedService(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*cor
...
 			return err
 		}
 		addAppLabel(cr.Spec.AppName, &service.ObjectMeta)
-		service.Spec.Ports = []corev1.ServicePort{
-			corev1.ServicePort{
-				Name: "memcached",
-				Port: 11211,
-			},
-		}
+		service.Spec.Ports[0].Name = "memcached"
```
carbonin, line 114: I'm worried that his `Ports` array will come back as `nil` if there isn't anything set. Can you check for nil here?

carbonin: Also it seems like it wouldn't be worth setting it in the service spec above, right?

bdunne: Nope, turns out I don't need that

bdunne: Fixed the nil issue and tested that it gets recreated is the service is deleted

carbonin:
Looks good, but what was the problem @bdunne? What were we touching that we shouldn't have been?

bdunne:
Previously:
`{"level":"info","ts":1594070650.2048993,"logger":"controller_manageiq","msg":"Service has been reconciled","component":"memcached","result":"updated"}`
Now:
`{"level":"info","ts":1594146078.004267,"logger":"controller_manageiq","msg":"Service has been reconciled","component":"memcached","result":"unchanged"}`

Something on the ServicePort was being overwritten causing the service to be updated on every loop.

carbonin:
Oh, sure. I was just curious what exactly the server what changing between reconcile calls.

bdunne:
I'm not sure what adds the `TargetPort`, maybe it's removing that and the server adds it back in?

PR #579, opened 2020-07-08
Use a service account for pull secrets

carbonin:
This PR creates a default service account which will be used by any deployment not already using one we define (the orchestrator and, in some cases, httpd, already use a service account to escalate various privileges). Having done that, we can now associate the image pull secret to all the service accounts we create to ensure it is available to all deployments that are created by either the operator or orchestrator.

I think we may want this in Jansa, so I'll probably have to make a separate PR there as master has diverged rather significantly with the `CreateOrUpdate` changes.

### DIFF
--- a/manageiq-operator/go.mod
+++ b/manageiq-operator/go.mod
@@ -3,7 +3,6 @@ module github.com/ManageIQ/manageiq-pods/manageiq-operator
 go 1.14
 
 require (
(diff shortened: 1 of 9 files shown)


simaishi:
Backported to jansa via https://github.com/ManageIQ/manageiq-pods/pull/580

PR #581, opened 2020-07-09
Fix missing label for selector

bdunne:
The httpd pod is missing the label "service": "httpd" that the service is using as a selector.  So traffic is not getting through to the httpd pods.

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/httpd.go
+++ b/manageiq-operator/pkg/helpers/miq-components/httpd.go
@@ -468,7 +468,7 @@ func HttpdService(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*corev1.Ser
 		}
 		service.Spec.Ports[0].Name = "http"
 		service.Spec.Ports[0].Port = 8080
-		service.Spec.Selector = map[string]string{"service": "httpd"}
+		service.Spec.Selector = map[string]string{"name": "httpd"}
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


bdunne:
@abellotti This should resolve the issue that you were seeing

carbonin:
I think the problem is that the service selector used to be `name: httpd`.
It was changed here https://github.com/ManageIQ/manageiq-pods/pull/572/files#diff-5d7ac153044a8a2acaa291f0ffb4f830L426

I would rather change the services back than add extra labels to the deployment.

carbonin:
Looks like a copy and paste error from the app worker services

bdunne:
Yeah, I was trying to be consistent with the rest of the services https://github.com/ManageIQ/manageiq/blob/d67c05fd7f52e3e98c537c22891114473214ed98/app/models/miq_worker/service_worker.rb#L13

abellotti:
@bdunne this does fix the issue.  I applied this to master in my project and MiQ is now accessible. 👍

carbonin:
That was done for a specific reason. We only have it that way because we want to load balance the service across multiple deployments (workers on separate servers).

For httpd, memcached, and postgres the service will only ever point to a single deployment, so we use the deployment name.

carbonin:
Question then is do we want to change everything or leave the worker services as a separate case (they do work differently).

I'm okay with either option FWIW.

carbonin:
Can you also fix the dbus api one?

bdunne:
> Can you also fix the dbus api one?

Done

PR #582, opened 2020-07-10
[JANSA] Enhancement to optionally fetch the OIDCAuthIntrospectionURL from the Provider

abellotti:
Enhancement to optionally fetch the OIDCAuthIntrospectionURL from the Provider Metadata URL if not provided to us.

Jansa backport of #571

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/httpd.go
+++ b/manageiq-operator/pkg/helpers/miq-components/httpd.go
@@ -68,7 +68,7 @@ func NewHttpdConfigMap(cr *miqv1alpha1.ManageIQ) (*corev1.ConfigMap, error) {
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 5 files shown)


manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -72,7 +83,7 @@ func NewHttpdConfigMap(cr *miqv1alpha1.ManageIQ) *corev1.ConfigMap {
 			Labels:    labels,
 		},
 		Data: data,
-	}
+	}, nil
```
carbonin, line 86: Can you just put the config map into a local so this is a bit easier to read?

abellotti: 👍  like master, will do.

PR #585, opened 2020-07-16
[WIP] ppc64le changes

seth-priya:
Fix ppc64le builds

### DIFF
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -56,7 +56,12 @@ RUN dnf -y --disableplugin=subscription-manager install \
     if [[ "$LOCAL_RPM" = "true" ]]; then /create_local_yum_repo.sh; fi && \
     dnf -y --disableplugin=subscription-manager --setopt=tsflags=nodocs install \
       ${RPM_PREFIX}-pods          \
+      libffi-devel                \
+      libxml2-devel               \
+      libxslt-devel               \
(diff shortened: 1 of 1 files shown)


simaishi:
Suggested to add those packages in https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/python_modules.ks.erb.

https://github.com/ManageIQ/manageiq-rpm_build/issues/43#issuecomment-659375282

seth-priya:
@simaishi needed a bit of help to understand how can I test out this change in manageiq-appliance-build, I have not had to look at the appliance repositories before so just trying to figure out how this fits in , thanks in  advance!

simaishi:
Since you really just need to test this particular change, I think the easiest way would be:

1. Clone manageiq-appliance-build
1. In that repo, run bin/kickstart_build.rb - this will create kickstart files, kickstarts/generated/base-*.ks
1. Use the generated kickstart file to create a VM from it

When OS is installed, you can check /root/anaconda-post.log on the VM and you should see some errors because of missing packages (should be same/similar to what you see when you run podified build). Then, you can just run modified python_modules.ks.erb on the machine to test out your changes.

There are things that can be removed from kickstart files for testing purpose, but it's probably easier if you just take what's generated.

seth-priya:
Closing in favour of https://github.com/ManageIQ/manageiq-appliance-build/pull/432

PR #593, opened 2020-07-27
Fix incorrect command in operator README

bdunne:


abellotti:
yep, that's what I run 👍

PR #600, opened 2020-07-29
[JANSA] We should default to jansa, not master

bdunne:
I guess we should default to `jansa-latest` here.  Not sure if we should change to a release tag at some point.  @simaishi @Fryguy @abellotti thoughts?

I noticed this yesterday while smoke testing jansa-1-rc1.


Fryguy:
@simaishi Do we need some sort of `rake release` thing here in order to set the values correctly for tagged releases?

simaishi:
Yes, we should create a releas task to update the values. I assume we want to update only for the tag (e.g. change to `jansa-1-rc1` only for jansa-1-rc1 tag and put back to `latest-jansa` for jansa branch)?  That's how we update Gemfile.

Issue #529, opened 2020-06-02
Don't deploy the database/kafka if the user provides the corresponding secret


carbonin:
We shouldn't deploy our own database or kafka instance if the user has directed us to use an external one. At a minimum we should only deploy if we're creating the secret, but it might be good to go as far as reading the hostname from the secret to ensure that they're not just providing a user or password or something like that.

[ ] For PostgreSQL/Kafka - https://github.com/ManageIQ/manageiq-pods/pull/604

@miq-bot add-label enhancement

Fryguy:
@gtanzillo This needs to be P1

abellotti:
[ ] For PostgreSQL - https://github.com/ManageIQ/manageiq-pods/pull/604

PR #604, opened 2020-08-04
Do not bring up the PostgreSQL or Kafka/Zookeeper services if specifying external services

abellotti:
Do not bring up:
- the PostgreSQL service if the user is using their own external PostgreSQL server and
they specified the postgresql secret with the external PostgreSQL hostname defined there. 
- the Kafka and Zookeeper service if the user specified a Kafka secret with a Kafka messaging
service hostname defined.

Fixes: https://github.com/ManageIQ/manageiq-pods/issues/529

### DIFF
--- a/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
+++ b/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
@@ -120,23 +120,8 @@ func (r *ReconcileManageIQ) Reconcile(request reconcile.Request) (reconcile.Resu
 	if e := r.generateDefaultServiceAccount(miqInstance); e != nil {
 		return reconcile.Result{}, e
 	}
-	externalPostgreSQL := false
-	if miqInstance.Spec.DatabaseSecret != "" {
-		secretKey := types.NamespacedName{Namespace: request.Namespace, Name: miqInstance.Spec.DatabaseSecret}
-		secret := &corev1.Secret{}
-		secretErr := r.client.Get(context.TODO(), secretKey, secret)
-		if secretErr == nil {
-			hostName := string(secret.Data["hostname"])
-			if hostName != "" {
-				logger.Info("External PostgreSQL Database selected, skipping postgresql service reconciliation", "hostname", hostName)
-				externalPostgreSQL = true
-			}
-		}
-	}
-	if externalPostgreSQL == false {
-		if e := r.generatePostgresqlResources(miqInstance); e != nil {
-			return reconcile.Result{}, e
-		}
+	if e := r.generatePostgresqlResources(miqInstance); e != nil {
+		return reconcile.Result{}, e
 	}
 	if e := r.generateHttpdResources(miqInstance); e != nil {
 		return reconcile.Result{}, e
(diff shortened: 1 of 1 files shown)


chessbyte:
I see this pattern over and over in our operator.  Is it possible to make a method/function in Go to abstract this?

abellotti:
Are you talking about the new code added or the following pattern ?

```		
if e := r.generatePostgresqlResources(miqInstance); e != nil {
			return reconcile.Result{}, e
```

For the latter, I'd keep those as it's the go pattern for error check/propagation up.  For the new code added, I agree, I'd like to generalize it if possible, I'll work on the kafka side and see if we can use common code for secret checks and such.

Fryguy:
@abellotti How are you distinguishing an external postgres vs an internal postgres where we've autocreated the secret...is it the hostname being blank?

abellotti:
Yes, a user must have provided a secret with a non-blank hostname.

abellotti:
Keeping as WIP for now as it needs some more testing.

abellotti:
Successfully tested using an external Kafka server after having fun with https://github.com/ManageIQ/manageiq-pods/issues/607#issuecomment-673118131 - unwipping.

Fryguy:
> Yes, a user must have provided a secret with a non-blank hostname.

But what does the generated secret look like?  If it also has a hostname, then you can't tell them apart after the fact.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -439,3 +451,13 @@ func (r *ReconcileManageIQ) createk8sResIfNotExist(cr *miqv1alpha1.ManageIQ, res
 	}
 	return nil
 }
+
+func getSecretKeyValue(client client.Client, nameSpace string, secretName string, keyName string) string {
```
Fryguy, line 455: I think there's another file with secret methods, where this might belong better.

abellotti: Not that I saw, pod specific helpers (postgres, kafka, etc) to return the default secret name and default secret if not specified.  One possibility is to try to move this to pkg/helpers/miq-components/util.go

bdunne: I don't see a better place for this right now.

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -267,6 +267,12 @@ func (r *ReconcileManageIQ) generateMemcachedResources(cr *miqv1alpha1.ManageIQ)
 }
 
 func (r *ReconcileManageIQ) generatePostgresqlResources(cr *miqv1alpha1.ManageIQ) error {
+	hostName := getSecretKeyValue(r.client, cr.Namespace, cr.Spec.DatabaseSecret, "hostname")
+	if hostName != "" {
```
bdunne, line 271: @abellotti Should this be `if hostName != "postgresql" {`?

Fryguy: No, this is what triggers external vs internal - https://github.com/ManageIQ/manageiq-pods/pull/604#issuecomment-670514247

bdunne: It's not obvious to me why this would be empty string since either the user creates the secret with a hostname or we do... https://github.com/ManageIQ/manageiq-pods/blob/89efdbe14731a17b4eb33777725d9562aa5fac44/manageiq-operator/pkg/helpers/miq-components/postgresql.go#L19

abellotti: It should stay if hostName != "", if the user specified  a secret with a non-blank hostname we're done, we don't get to call that DefaultPostgresqlSecret (where we populate the secret for the local postgresql case), which is called on line 276 after the if clause.

Fryguy: @bdunne Yeah I had the same question - I assumed when we generate it ourselves we don't specify a hostname?

PR #615, opened 2020-08-19
ManageIQ docker container

Kuldip-Nanda:
This provides a README files and a script to develop just the manageiq container

No
@miq-bot add-label enhancement


Fryguy:
@agrare @simaishi Please review.

bin/build
```diff
@@ -102,6 +103,10 @@ pushd $IMAGE_DIR
   done
 popd
 
+if [ -n "$UI_DOCKER_ONLY" ]; then
+  exit 0
+fi
+
```
Fryguy, line 109: I think it might be better to introduce something like a "no-operator" parameter instead.  I've heard requests to avoid building the operator so that would solve 2 problems at once.  With that in place, then you don't need this because it will fall through the next 2 sections and exit as expected.

docker-assets/ManageIQContainerBuildProcess.md
```diff
@@ -0,0 +1,97 @@
+## Building and running a ManageIQ Container
```
Fryguy, line 1: For consistency please rename the file with underscore case (i.e. `manageiq_container_build_process.md`)

docker-assets/ManageIQContainerBuildProcess.md
```diff
@@ -0,0 +1,97 @@
...
+
+
+	Developer may configure bundler  with git user login and password, or personal access token to download and install **manageiq-provider-plugin** from git during **bundle install**
+	```
+    bundle config GITHUB__COM username:password USER:XXXXXXXX --global
+    bundle confg
+     Settings are listed in order of priority. The top value will be used.
+     github.com
+     Set for the current user (/root/.bundle/config): "USER:XXXXXXXX"
+   ```
```
Fryguy, line 21: I'm not sure we need deep details in this section.  We already have this documented in the guides repo, so it might be better to just say something like

```md
1. In ManageIQ core, set up your repository for plugin development.  You can read more on how to do that [here](/docs/guides/developer_setup/plugins.html).
```

Fryguy: If you want to add more specific plugin development setup for things ilke this password section, I would recommend putting it in the plugin guide: https://github.com/ManageIQ/guides/blob/master/developer_setup/plugins.md

bdunne:
It feels like our build scripts in https://github.com/ManageIQ/manageiq-appliance-build/tree/master/bin could be enhanced to support adding plugins and bundler config rather than duplicating here.

simaishi:
I think this PR is a bit confusing as this covers a few different things:
1. The build option for UI only (= no operator)
1. Adding a plugin (and create RPMs using it)
1. Use CI_USER_TOKEN
1. Create a monolithic image

How about splitting this to:
1. Add no operator option
1. Enhance the plugin guide as needed as @Fryguy suggested
1. Add a standalone script to build pods images followed by monolithic image (without plugin or any extra stuff - as I think that shouldn't be a part of the build script itself, but configuration at different stages)

The pods build script supports rpm build option (`-b`) where you can pass options.yml and build RPMs before proceeding to pods image build. We might need to enhance the build script to handle whatever is currently missing from what's done in this PR, but the steps probably can be shortened by using -b.

simaishi:
.. and I realized that I never updated "Building images" section of the README and that's outdated. Sorry about that, I'll update.

Kuldip-Nanda:
The aim of this PR was to create the docker image with user plugin. @simaishi is correct in that sense that it has may different parts, but it does present a complete beginning to end approach of creating a manageiq docker image with custom plugin. I have also tested this in both Travis and as an independent script on a host machine.

My suggestion is that I can open 3 more PR's to
1) Update plugin guide 
2) Add no operator option
3) Add a standalone script 
All the PR can be referred to in this PR and once completed can be closed.

simaishi:
Actually... we might want to consider de-coupling monolithic and podfied. The monolithic was stand alone in the beginning and we changed to base on the podified image because there were lots of duplicate work being done in both places (bundle install, compile assets, etc.). But we now use RPMs and those duplicates are gone, so we probably are not benefiting much.

Kuldip-Nanda:
- If we are going to consider decoupling monolithic and podified docker images, then the manageiq docker image can be straight away build from the RPM's. In that case I will like to
```
  - Close this PR.
  - Update the manageiq Dockerfile  in manageiq/manageiq. So it will be building manageiq/manageiq Docker from manageiq rpm straight away. I will also develop a standalone script to make the rpms. 
```
- Add no operator option
- Update plugin guide

simaishi:
@Fryguy @bdunne Thoughts on decoupling?

bdunne:
My biggest concern is the duplication of other stuff in the Dockerfiles.

bin/build
Fryguy: See #621

Fryguy:
no operator option added in #621

Fryguy:
Now that we have RPMs, I'd be +1 on decoupling, though I would like to understand better @bdunne's concerns

Fryguy:
@Kuldip-Nanda 

> I will also develop a standalone script to make the rpms. 

No need...this is all scripted in https://github.com/ManageIQ/manageiq-rpm_build

docker-assets/ManageIQContainerBuildProcess.md
bdunne: Also, I don't think this should be in the `docker-assets` directory.

Fryguy:
This is currently all scripted, so for now I'm closing this.  For plugin development we now have https://github.com/ManageIQ/guides/pull/457.  If there is more here that is needed, feel free to reopen.

PR #619, opened 2020-09-01
Add velero-backup label to important resources that need to be backed up

bdunne:
Depends On:
- [x] https://github.com/ManageIQ/manageiq-pods/pull/628
- [x] https://github.com/ManageIQ/manageiq-pods/pull/634

This allows us to simplify the listing the resources that need backing up.
`velero backup create my_backup_name -l velero-backup=t`

WIP because I need to add some logic to ensure that the label exists on some objects that we expect the deployer to create, but I wanted to consolidate the label naming discussion here.

Naming:
- Start the label name with something like `miq-` or `manageiq.org/` causes some problems for productization.
- Other apps in our namespace could use the same label if it's not namespaced.

Items that need to be backed up:

- Secrets from keys in the CR:
  - [X] databaseSecret (default: postgresql-secrets)
  - [X] httpdAuthConfig
  - [x] imagePullSecret
  - [X] kafkaSecret (default: kafka-secrets)
  - [x] oidcCaCertSecret
  - [x] oidcClientSecret
  - [X] tlsSecret (default: tls-secret)

- PVCs:
  - [X] postgresql
  - [X] kafka-data (if present)
  - [X] zookeeper-data (if present)

- Other objects:
  - [X] secrets/app-secrets
  - [x] role/manageiq-operator
  - [x] role_binding/manageiq-operator
  - [x] service_account/manageiq-operator
  - [ ] ~~crd/manageiqs.manageiq.org~~
  - [x] CR
  - [x] deployment.apps/manageiq-operator

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/util.go
+++ b/manageiq-operator/pkg/helpers/miq-components/util.go
@@ -61,6 +61,13 @@ func addAppLabel(appName string, meta *metav1.ObjectMeta) {
 	meta.Labels["app"] = appName
 }
 
+func addBackupLabel(backupLabel string, meta *metav1.ObjectMeta) {
+	if meta.Labels == nil {
+		meta.Labels = make(map[string]string)
+	}
+	meta.Labels[backupLabel] = "t"
+}
+
 func addBackupAnnotation(volumesToBackup string, meta *metav1.ObjectMeta) {
 	if meta.Annotations == nil {
 		meta.Annotations = make(map[string]string)
--- a/manageiq-operator/README.md
+++ b/manageiq-operator/README.md
@@ -199,3 +199,29 @@ spec:
   ...
   oidcCaCertSecret: <name of your openshift OIDC CA cert>
 ```
+
+### Backup and restore with Velero and Restic
+
+We will assume that Velero and Restic are already installed and working.
+The operator adds labels and/or annotations to everything that needs to be backed up so that anyone can easily and quickly backup only what is required.
+The operator will add annotations for restic volume backups i.e. `backup.velero.io/backup-volumes: miq-pgdb-volume`.
+The backup label key can be configured by adding the following to the CR, but by default it will apply the label `manageiq.org/backup=t` if nothing else is specified.
+
+```yaml
+...
+spec:
+  ...
+  backupLabelName: <name of your openshift OIDC CA cert> (default: manageiq.org/backup)
+```
+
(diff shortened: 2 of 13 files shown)


jrafanie:
> * Start the label name with something like `miq-` or `manageiq.org/` causes some problems for productization.

What works or doesn't work in productization?  It seems like `app.kubernetes.io/component` or similar naming is the convention for namespacing labels in kubernetes so I wonder what formats we can use that will get as close as possible to conventions used by other applications.  Note, the docs seem to indicate that "Shared labels and annotations share a common prefix: app.kubernetes.io. Labels without a prefix are private to users. The shared prefix ensures that shared labels do not interfere with custom user labels."

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

I guess we only need a namespaced prefix for shared labels.

> * Other apps in our namespace could use the same label if it's not namespaced.

Yeah, the eternal problem.

I need to do something similar with marking worker pod definitions for pods we want to check for failed deployments so I'd like to follow a pattern.

bdunne:
Ok, @Fryguy @gtanzillo @abellotti I think this is ready for review.  I crossed off the CRD in the list because it is needed, but it's outside of our namespace and I don't think our Service Account has permission to create/edit it.  I made a note in the restore documentation that you need to create it.

@jrafanie I set the key name to `manageiq.org/backup`.  What do you think?

gtanzillo (approved):

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -61,6 +61,13 @@ func addAppLabel(appName string, meta *metav1.ObjectMeta) {
 	meta.Labels["app"] = appName
 }
 
+func addBackupLabel(backupLabel string, meta *metav1.ObjectMeta) {
+	if meta.Labels == nil {
+		meta.Labels = make(map[string]string)
+	}
+	meta.Labels[backupLabel] = "t"
```
Fryguy, line 68: Why `t` as opposed to `true` or an actual boolean?

bdunne: You have to use a string as the label value.  Things were getting confused when I put `true` there because it thought i was putting a boolean in a string field.

Fryguy: Ok, I don't love it, but let's merge.

PR #627, opened 2020-09-17
[JANSA] Install ovirt packages before ansible venv packages

simaishi:
The Jansa podified build is hitting the same error as described in https://github.com/ManageIQ/manageiq-appliance-build/pull/440 for ansible venv setup.  That PR didn't fix the issue for podified, as podified build doesn't run post actions in the same order as appliance build:

ovirt_ansible -> python_modules (appliance)
python_modules -> ovirt_ansible (podified)

ovirt_ansible is what installs the needed package.

This is not an issue in master branch, as ovirt related packages were moved to rpm "Requires" and packages are no longer being installed as a post action. So I'm going to just swap the order for jansa podified build to match appliance build, which will eliminate the error.

### DIFF
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -63,10 +63,10 @@ RUN dnf -y --disableplugin=subscription-manager install \
     dnf clean all
 
 # Install python packages the same way the appliance does
-COPY --from=0 build/kickstarts/partials/post/python_modules.ks.erb /tmp/python_modules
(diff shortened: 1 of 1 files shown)


chessbyte:
@simaishi this is just a re-ordering of running the python install after the ansible install, right?  Assuming you have tested it, this change is simple and looks good to me.

simaishi:
Yes, just re-ordering the 2 post actions, and I've run a build and confirmed ansible venv install finish successfully with this change.

PR #632, opened 2020-09-24
[WIP] Enable PowerTools for python3-pylxca dependency

simaishi:
Replacing pip install of vspk/pylxca with rpm

Goes with https://github.com/ManageIQ/manageiq-appliance-build/pull/441 and https://github.com/ManageIQ/manageiq-rpm_build/pull/93


simaishi:
This will be a problem for s390x. Since it's noarch rpm, we can copy to manageiq yum repo and avoid using PowerTools.

PR #634, opened 2020-09-30
Store the MiqServer GUID for the orchestrator on the CR

bdunne:
Alternative to #575 since using the CR's UID is somewhat problematic.

- If the CR is recreated (backup / restore), the UID changes, so the orchestrator would exit with "New replica is not supported".  This persists the UID, so we don't have that problem.
- This also makes it easier to import a database from an appliance because you can specify the GUID from an existing appliance.


manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
```diff
@@ -158,6 +158,9 @@ spec:
             memcachedSlabPageSize:
               description: 'Memcached max item size (default: 1m, min: 1k, max: 1024m)'
               type: string
+            miqServerGuid:
```
Fryguy, line 161: let's leave out the miq for prodcutization reasons (I know it's related to MiqServer model, but end users don't see that; they only see "Server" in the UI).

Fryguy:
Aside from the naming LGTM 👍

Fryguy (approved):

PR #636, opened 2020-10-13
[WIP] Remove manageiq-appliance-build clone and customization hook

simaishi:
The last use of manageiq-appiance-build repo is removed and we no longer need to multi stage pods build to clone appliance build repo, etc.

Since no source clone is done in pods build, customization hook is no loner needed either.

Depends on:
- Build manageiq-ansible-venv rpm for supported architecture and add 'Requires' to rpm spec file

For https://github.com/ManageIQ/manageiq-appliance-build/issues/423


Fryguy:
Closing in favor of #721

PR #643, opened 2020-10-27
[JANSA] Test with ruby 2.6.6 on Jansa

jrafanie:
Test with ruby 2.6.6 on Jansa


jrafanie:
Failure with blaster outer... starting over.

PR #646, opened 2020-10-28
Update master to use manageiq 12-lasker yum repo

simaishi:


images/manageiq-base/container-assets/create_local_yum_repo.sh
```diff
@@ -12,4 +12,4 @@ enabled=1
 gpgcheck=0
 EOF
 
-dnf config-manager --setopt=manageiq-11-kasparov*.exclude=manageiq-* --save
+dnf config-manager --setopt=manageiq-*.exclude=manageiq-* --save
```
chessbyte, line 15: should this be `--setopt=manageiq-12-lasker*.exclude=manageiq-*` or are you changing the format of the setopt to be more open?

simaishi: Yes, I'm making it more open, so we don't need to change on every release (although we have to change others anyway...)

PR #649, opened 2020-11-04
Add Ingress Network Policies to the operator

### LINKED ISSUE #590: Use NetworkPolicy to restrict traffic between pods
Create [NetworkPolicy](https://docs.openshift.com/container-platform/4.5/networking/network_policy/creating-network-policy.html) objects in our namespace to define the expected channels of communication between pods so traffic cannot come from an unexpected location.

This generally improves security.

bdunne:
Fixes #590 

Allow the following ingress:
- default deny all
- external traffic -> httpd
- httpd -> api
- httpd -> ui
- orchestrator & orchestrated-by -> postgres
- orchestrator & orchestrated-by -> memcached

TODO:
- [x] refactoring common patterns to make it less verbose
- [x] testing

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/network_policies.go
+++ b/manageiq-operator/pkg/helpers/miq-components/network_policies.go
@@ -40,21 +40,14 @@ func NetworkPolicyAllowInboundHttpd(cr *miqv1alpha1.ManageIQ, scheme *runtime.Sc
 		networkPolicy.Spec.PodSelector.MatchLabels = map[string]string{"name": "httpd"}
 
 		ensureIngressRule(networkPolicy)
+		setFirstIngressTCPPort(networkPolicy, 8080)
 		if len(networkPolicy.Spec.Ingress[0].From) != 1 {
 			networkPolicy.Spec.Ingress[0].From = []extensionsv1beta1.NetworkPolicyPeer{
 				extensionsv1beta1.NetworkPolicyPeer{},
 			}
 		}
-		if len(networkPolicy.Spec.Ingress[0].Ports) != 1 {
-			networkPolicy.Spec.Ingress[0].Ports = []extensionsv1beta1.NetworkPolicyPort{
-				extensionsv1beta1.NetworkPolicyPort{},
-			}
-		}
-		tcp := corev1.ProtocolTCP
 		networkPolicy.Spec.Ingress[0].From[0].IPBlock = &extensionsv1beta1.IPBlock{}
 		networkPolicy.Spec.Ingress[0].From[0].IPBlock.CIDR = "0.0.0.0/0"
-		networkPolicy.Spec.Ingress[0].Ports[0].Protocol = &tcp
-		networkPolicy.Spec.Ingress[0].Ports[0].Port = &intstr.IntOrString{Type: intstr.Int, IntVal: 8080}
 
 		return nil
 	}
@@ -75,21 +68,14 @@ func NetworkPolicyAllowHttpdApi(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme
 		networkPolicy.Spec.PodSelector.MatchLabels = map[string]string{"service": "web-service"}
 
 		ensureIngressRule(networkPolicy)
+		setFirstIngressTCPPort(networkPolicy, 3000)
 		if len(networkPolicy.Spec.Ingress[0].From) != 1 {
 			networkPolicy.Spec.Ingress[0].From = []extensionsv1beta1.NetworkPolicyPeer{
 				extensionsv1beta1.NetworkPolicyPeer{},
 			}
 		}
-		if len(networkPolicy.Spec.Ingress[0].Ports) != 1 {
-			networkPolicy.Spec.Ingress[0].Ports = []extensionsv1beta1.NetworkPolicyPort{
-				extensionsv1beta1.NetworkPolicyPort{},
-			}
-		}
-		tcp := corev1.ProtocolTCP
 		networkPolicy.Spec.Ingress[0].From[0].PodSelector = &metav1.LabelSelector{}
 		networkPolicy.Spec.Ingress[0].From[0].PodSelector.MatchLabels = map[string]string{"name": "httpd"}
-		networkPolicy.Spec.Ingress[0].Ports[0].Protocol = &tcp
-		networkPolicy.Spec.Ingress[0].Ports[0].Port = &intstr.IntOrString{Type: intstr.Int, IntVal: 3000}
 
 		return nil
 	}
@@ -110,21 +96,14 @@ func NetworkPolicyAllowHttpdUi(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme)
 		networkPolicy.Spec.PodSelector.MatchLabels = map[string]string{"service": "ui"}
 
 		ensureIngressRule(networkPolicy)
+		setFirstIngressTCPPort(networkPolicy, 3000)
 		if len(networkPolicy.Spec.Ingress[0].From) != 1 {
 			networkPolicy.Spec.Ingress[0].From = []extensionsv1beta1.NetworkPolicyPeer{
 				extensionsv1beta1.NetworkPolicyPeer{},
 			}
 		}
-		if len(networkPolicy.Spec.Ingress[0].Ports) != 1 {
-			networkPolicy.Spec.Ingress[0].Ports = []extensionsv1beta1.NetworkPolicyPort{
-				extensionsv1beta1.NetworkPolicyPort{},
-			}
-		}
-		tcp := corev1.ProtocolTCP
 		networkPolicy.Spec.Ingress[0].From[0].PodSelector = &metav1.LabelSelector{}
 		networkPolicy.Spec.Ingress[0].From[0].PodSelector.MatchLabels = map[string]string{"name": "httpd"}
-		networkPolicy.Spec.Ingress[0].Ports[0].Protocol = &tcp
-		networkPolicy.Spec.Ingress[0].Ports[0].Port = &intstr.IntOrString{Type: intstr.Int, IntVal: 3000}
 
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


bdunne:
Ok @gtanzillo @abellotti I think this is ready for review.  I deployed it and I was able to browse the UI and ping the API.  Do you think there is any additional testing we should do before merge?
cc @Fryguy

manageiq-operator/pkg/helpers/miq-components/network_policies.go
```diff
@@ -0,0 +1,237 @@
...
+		return nil
+	}
+
+	return networkPolicy, f
+}
+
+func newNetworkPolicy(cr *miqv1alpha1.ManageIQ, name string) *extensionsv1beta1.NetworkPolicy {
+	return &extensionsv1beta1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cr.Spec.AppName + name,
```
Fryguy, line 194: What is name here?  Even so, doesn't this need something like a `-` between the name and app name?

manageiq-operator/pkg/helpers/miq-components/network_policies.go
```diff
@@ -0,0 +1,237 @@
...
+			Namespace: cr.ObjectMeta.Namespace,
+		},
+	}
+}
+
+func orchestratorPod(c client.Client) *corev1.Pod {
+	podList := &corev1.PodList{}
+	c.List(context.TODO(), podList)
```
Fryguy, line 202: TODO?

manageiq-operator/pkg/helpers/miq-components/network_policies.go
Fryguy: or appname + -networkpolicy-  + name?

manageiq-operator/pkg/helpers/miq-components/network_policies.go
```diff
@@ -0,0 +1,237 @@
...
+
+func setFirstIngressTCPPort(networkPolicy *extensionsv1beta1.NetworkPolicy, port int32) {
+	if len(networkPolicy.Spec.Ingress[0].Ports) != 1 {
+		networkPolicy.Spec.Ingress[0].Ports = []extensionsv1beta1.NetworkPolicyPort{
+			extensionsv1beta1.NetworkPolicyPort{},
+		}
+	}
+	tcp := corev1.ProtocolTCP
+	networkPolicy.Spec.Ingress[0].Ports[0].Protocol = &tcp
+	networkPolicy.Spec.Ingress[0].Ports[0].Port = &intstr.IntOrString{Type: intstr.Int, IntVal: port}
```
Fryguy, line 236: Can we DRY up some of this code, with something like -

```go
ports := networkPolicy.Spec.Ingress[0].Ports
if len(ports) != 1 {
  ...

ports[0].Protocol = &tcp
...
```

manageiq-operator/pkg/helpers/miq-components/network_policies.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func setFirstIngressTCPPort(networkPolicy *extensionsv1beta1.NetworkPolicy, port int32) {
+	if len(networkPolicy.Spec.Ingress[0].Ports) != 1 {
+		networkPolicy.Spec.Ingress[0].Ports = []extensionsv1beta1.NetworkPolicyPort{
+			extensionsv1beta1.NetworkPolicyPort{},
+		}
+	}
+	tcp := corev1.ProtocolTCP
+	networkPolicy.Spec.Ingress[0].Ports[0].Protocol = &tcp
```
Fryguy, line 235: I'm not a Go write, but can this be:

```suggestion
	networkPolicy.Spec.Ingress[0].Ports[0].Protocol = &(corev1.ProtocolTCP)
```

Fryguy (approved):
Some minor comments above, but overall LGTM.

two questions
- Will this be backportable to Jansa if necessary? (not saying we will, just want to know if it's an option)
- Some names are hardcoded - "orcehstrator", "httpd" - should those be
  a. extracted to constants
  b. isolated to only those related to the operator that deployed them?  For example if there are multiple instance of MIQ, _which_ httpd has the access rights?

chessbyte:
@bdunne Why does the name of the PR start with `Initial commit`?  Are you planning more PRs in this area?

manageiq-operator/pkg/helpers/miq-components/network_policies.go
bdunne: The `-` was in the caller, I moved it here.

Here's an example of the names:
```
$ oc get networkpolicy
NAME                           POD-SELECTOR          AGE
manageiq-allow-httpd-api       service=web-service   55s
manageiq-allow-httpd-ui        service=ui            55s
manageiq-allow-inbound-httpd   name=httpd            55s
manageiq-allow-memcached       name=memcached        55s
manageiq-allow-postgres        name=postgresql       55s
manageiq-default-deny          <none>                55s
```

manageiq-operator/pkg/helpers/miq-components/network_policies.go
bdunne: https://golang.org/pkg/context/#TODO

My understanding is that we don't need to pass a context here, but you shouldn't pass in a `nil`.

manageiq-operator/pkg/helpers/miq-components/network_policies.go
bdunne: Unfortunately not...
```
INFO[0000] Running the operator locally in namespace bdunne-miq. 
# github.com/ManageIQ/manageiq-pods/manageiq-operator/pkg/helpers/miq-components
pkg/helpers/miq-components/network_policies.go:234:52: cannot take the address of "k8s.io/api/core/v1".ProtocolTCP
FATA[0000] Failed to run operator locally: failed to build operator to run locally: failed to exec []string{"go", "build", "-o", "build/_output/bin/manageiq-operator-local", "github.com/ManageIQ/manageiq-pods/manageiq-operator/cmd/manager"}: exit status 2 
```

manageiq-operator/pkg/helpers/miq-components/network_policies.go
bdunne: I don't think so.  On initial pass (network policy creation), the ports is then empty.  If we then added `networkPolicy.Spec.Ingress[0].Ports = ports` to the end it could work, but that typically causes problems in the reconcile loop sending unnecessary updates when nothing has changed.

manageiq-operator/pkg/helpers/miq-components/network_policies.go
bdunne: BTW... I'm not an expert either and this also bothers me so if anyone else knows of a way to do it, I'd be happy to change it.

bdunne:
> Some minor comments above, but overall LGTM.
> 
> two questions
> 
>     * Will this be backportable to Jansa if necessary? (not saying we will, just want to know if it's an option)

Um... I think it could be backported without too much effort, but I haven't tried.

> 
>     * Some names are hardcoded - "orcehstrator", "httpd" - should those be
>       a. extracted to constants

I think this could be done, but I'd prefer not to do it in this PR.  There are many other references like this in the other components and I would prefer to update those at the same time.

>       b. isolated to only those related to the operator that deployed them?  For example if there are multiple instance of MIQ, _which_ httpd has the access rights?

These Network Policies are scoped to the namespace (that's the default).  At the moment you can only have one ManageIQ deployed in a namespace and you can't have multiple deployments with the same name, so I don't think this would be a problem.

bdunne:
> @bdunne Why does the name of the PR start with `Initial commit`? Are you planning more PRs in this area?

Changed.  Network Policies can cover ingress and/or egress traffic.  In the future we may also want to add egress rules, but these ingress rules protect us from any unexpected incoming traffic.

manageiq-operator/pkg/helpers/miq-components/network_policies.go
abellotti: what about `networkPolicy.Spec.Ingress[0].Ports[0].Protocol = &corev1.ProtocolTCP{}` ?

bdunne: Nope.
```
INFO[0000] Running the operator locally in namespace bdunne-miq. 
# github.com/ManageIQ/manageiq-pods/manageiq-operator/pkg/helpers/miq-components
pkg/helpers/miq-components/network_policies.go:234:71: "k8s.io/api/core/v1".ProtocolTCP is not a type
FATA[0000] Failed to run operator locally: failed to build operator to run locally: failed to exec []string{"go", "build", "-o", "build/_output/bin/manageiq-operator-local", "github.com/ManageIQ/manageiq-pods/manageiq-operator/cmd/manager"}: exit status 2 
```

abellotti: Ok, It's just a constant https://github.com/kubernetes/api/blob/5dbd1534041f2d0d080a2cf3b124fb095312c948/core/v1/types.go#L928 we can't take an address of a constant thus the assignment to the tcp var needed.

manageiq-operator/pkg/helpers/miq-components/network_policies.go
Fryguy: 😑

manageiq-operator/pkg/helpers/miq-components/network_policies.go
Fryguy: I don't follow.... I'm talking within the method...does `.Ports` change within the pass of the method?

bdunne: This method is called repeatedly, once on operator startup then every reconcile and its results are passed to `CreateOrUpdate`.

For the initial creation, if we added the setter `networkPolicy.Spec.Ingress[0].Ports = ports` that would work fine.  That will create the network policy as expected.  In the process, kubernetes will fill in any other default fields to define the object.

Next time reconcile is called, that would cause problems because that line will overwrite the definition of the ports (without any of the kubernetes defaults).  This will send a patch to the API, unnecessarily updating the NetworkPolicy definition.  Kubernetes will fill in the defaults again, and every time reconcile is called, this will happen again.

If we put a watch on the NetworkPolicies, this will put the operator into a never-ending reconcile .loop

Fryguy: Is `networkPolicy.Spec.Ingress[0].Ports` a method-with-side-effects or a property?  I'm really lost, and I suspect it's because I don't understand Go.

abellotti: Golang student here too.

All of these usually access and update the kubernetes objects (json) using well defined structure definitions which otherwise would be accessing generic unstructured hash/maps.

Here you can follow the NetworkPolicy definitions down to the specific Ports being accessed/updated.



https://godoc.org/k8s.io/api/extensions/v1beta1#NetworkPolicy
https://godoc.org/k8s.io/api/extensions/v1beta1#NetworkPolicySpec
https://godoc.org/k8s.io/api/extensions/v1beta1#NetworkPolicyIngressRule
https://godoc.org/k8s.io/api/extensions/v1beta1#NetworkPolicyPort

Fryguy: The problem for me isn't the structures, it's the code itself, or perhaps, what you can do with structs in the code...

I guess my question is, at a fundamental go level...what is the difference between:

```go
ports := networkPolicy.Spec.Ingress[0].Ports
ports[0].Protocol = &foo
ports[0].Port = &bar
```

and

```go
networkPolicy.Spec.Ingress[0].Ports[0].Protocol = &foo
networkPolicy.Spec.Ingress[0].Ports[0].Port = &bar
```

Fryguy: I played around in the Go playground for a bit, and I think I understand the difference in that 

```go
ports := networkPolicy.Spec.Ingress[0].Ports
```

creates a _copy_ of the struct, so changing it does nothing to the final object.  However if you take a reference to it, then you can dry it up...

```go
ports := &networkPolicy.Spec.Ingress[0].Ports
ports[0].Protocol = &foo
ports[0].Port = &bar
```

abellotti (approved):

manageiq-operator/pkg/helpers/miq-components/network_policies.go
bdunne: Done

PR #653, opened 2020-11-19
Prefer recreate as the deployment strategy for many services.

bdunne:
Rolling deployments require that a new pod be available before the old pod is killed for availability reasons.  However we have many services that should only ever have a single replica and having a second pod can actually cause some strange bugs.  This went unnoticed for many of our services since nothing would prevent a second pod from coming up, but it was found on the postgresql deployment since it needs to mount the data PVC and with most backing storages are `ReadWriteOnce` and the pod can not mount the PVC if another pod has it mounted.

I switched postgresql, kafka, zookeeper, memcached and the orchestrator to `Recreate` since there should only ever be one replica of those.  ~~I didn't see a problem with having multiple replicas of httpd, so I left that as `Rolling`.~~

EDIT: I switched httpd to `Recreate` as well until we have time to test whether or not it is safe to have multiple pods running concurrently.

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/httpd.go
+++ b/manageiq-operator/pkg/helpers/miq-components/httpd.go
@@ -361,6 +361,9 @@ func HttpdDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*appsv1.
 		addAppLabel(cr.Spec.AppName, &deployment.ObjectMeta)
 		var repNum int32 = 1
 		deployment.Spec.Replicas = &repNum
+		deployment.Spec.Strategy = appsv1.DeploymentStrategy{
+			Type: "Recreate",
+		}
 		deployment.Spec.Template.Spec.Containers = []corev1.Container{container}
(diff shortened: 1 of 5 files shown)


gtanzillo (approved):

simaishi:
Kasparov backport details:

```
$ git log -1
commit 016330ef2ccd8758068d0ba976f489eedadb71b6
Author: Gregg Tanzillo <gtanzill@redhat.com>
Date:   Thu Nov 19 17:56:15 2020 -0500

    Merge pull request #653 from bdunne/use_recreate_strategy

    Prefer recreate as the deployment strategy for many services.

    (cherry picked from commit 4d04b16bbb053308b7967faefbb4766b99a415a7)
```

simaishi:
@bdunne Backporting to jansa conflicts on all files. Please create a separate PR.

simaishi:
@bdunne Just checking to see if we still want to backport this PR (in which case, we'll need a separate PR).

bdunne:
Everything except for httpd and memcached are already using `Recreate` strategy on jansa, so we can skip that backport.

PR #655, opened 2020-12-01
Set the TLS certificate CN to the application domain

bdunne:
This is still a self-signed certificate, but now at least it matches the name on the ingress route.


abellotti:
was looking at the different fields for the subject in https://tools.ietf.org/rfc/rfc5280.txt, common name seems the appropriate one for the fqdn.

abellotti (approved):

gtanzillo (approved):

PR #656, opened 2020-12-02
Drop the old templates since the operator is stable

bdunne:
I believe all use cases of the templates are now covered by the operator.

Since all efforts are only focused on the operator development, we should remove the old templates to avoid confusion. :scissors: :fire: :toilet:


jrafanie (approved):
😂 😂 😂 😂 😂 😂 
👍 👍 👍 👍 👍 👍 
🍪 🍪 🍪 🍪 🍪 🍪 

Yes, please.  I keep using `git grep` and these files are always coming up.  Out with the old.

PR #658, opened 2020-12-03
Typos: Other limit variables use CpuLimit, not Cpulimit.

jrafanie:


Fryguy:
@bdunne Do we have upgrade concerns with something like this?

bdunne:
> @bdunne Do we have upgrade concerns with something like this?

Yep

kbrock:
This looks still to be the case:
- The yaml uses `kafkaCpulimit`.
- The go code references `kafkaCpuLimit`.
- the go serialization code references `kafkaCpulimit`.

Do we still want this PR to use camel case?

jrafanie:
> This looks still to be the case:
> 
> * The yaml uses `kafkaCpulimit`.
> * The go code references `kafkaCpuLimit`.
> * the go serialization code references `kafkaCpulimit`.
> 
> Do we still want this PR to use camel case?

Preferably we can normalize them to the same convention.  If any change is risky, we can just close this.

Issue #538, opened 2020-06-05
Document how to rebuild images with non-redistributable VMware libs


carbonin:
We should write some documentation about how a user could rebuild the images to include vmware vddk libs for smartstate and webmks for remote console access.

Right now both of those features are broken (as we can't include the required libs) and the standard solution (install them to the appliance) doesn't apply to pods as those changes would be lost on the first re-deploy.

PR #666, opened 2020-12-21
Add simple installation process for the VMware VDDK

bdunne:
Fixes: https://github.com/ManageIQ/manageiq-pods/issues/538

I'm not sure if we want to go this route or have a dedicated `Dockerfile` that builds on top of the worker base image, but since it's used for both scanning and webmks, I thought this might be easier.  Either way, the code would be the same.

### DIFF
--- a/images/manageiq-base-worker/Dockerfile
+++ b/images/manageiq-base-worker/Dockerfile
@@ -1,6 +1,13 @@
 ARG FROM_REPO=manageiq
 ARG FROM_TAG=latest
 
+FROM ${FROM_REPO}/manageiq-base:${FROM_TAG} AS vddk
+
+COPY container-assets/ /vddk/
+
+RUN /vddk/extract-vmware-vddk
+
+
 FROM ${FROM_REPO}/manageiq-base:${FROM_TAG}
 MAINTAINER ManageIQ https://manageiq.org
 
@@ -9,6 +16,10 @@ LABEL name="manageiq-base-worker" \
 
 COPY container-assets/entrypoint /usr/local/bin
 COPY container-assets/manageiq_liveness_check /usr/local/bin
+
+COPY --from=vddk /vddk/vmware-vix-disklib-distrib/ /usr/lib/vmware-vix-disklib/
+COPY container-assets/install-vmware-vddk /tmp/
+RUN /tmp/install-vmware-vddk
 
 RUN source /etc/default/evm && /usr/bin/generate_rpm_manifest.sh
 
--- a/images/manageiq-base-worker/container-assets/extract-vmware-vddk
+++ b/images/manageiq-base-worker/container-assets/extract-vmware-vddk
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+cd /vddk
+filename=$(ls VMware-*)
+
+# Skip the install if the VDDK library is not found
+[[ -f $filename ]] && tar -zxvf VMware-* || mkdir vmware-vix-disklib-distrib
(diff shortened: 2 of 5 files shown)


images/manageiq-base-worker/Dockerfile
```diff
@@ -9,6 +9,17 @@ LABEL name="manageiq-base-worker" \
 
 COPY container-assets/entrypoint /usr/local/bin
 COPY container-assets/manageiq_liveness_check /usr/local/bin
+COPY container-assets/VMware-* /tmp/
+
+RUN cd /tmp && \
```
Fryguy, line 14: Can you add a comment here, something like:

# Install the VDDK if found in `/tmp/VMware-*` ?  (Aside, we may want to provide an ARG, defaulted to `/tmp/VMware-*` instead of guessing the location)

cc @agrare

Fryguy: Oh sorry...I misunderstood that the caller puts it in the container-assets, and it's copied _into_ /tmp inside the container.  This is fine.

Would be nice to avoid the layer that has the VDDK in it however I'm not sure how.

images/manageiq-base-worker/Dockerfile
```diff
@@ -9,6 +9,17 @@ LABEL name="manageiq-base-worker" \
...
+    filename=$(ls VMware-*) && \
+    [[ -f $filename ]] && \
+    echo "VDDK library found: $filename, installing..." && \
+    tar -zxvf VMware-* && \
+    cp -rf vmware-vix-disklib-distrib/ /usr/lib/vmware-vix-disklib/ && \
+    for SO in $(ls /usr/lib/vmware-vix-disklib/lib64/libvixDiskLib.so*); do ln -s $SO /usr/lib/$(basename $SO); done && \
+    ldconfig && \
+    rm -rf /tmp/VMware-* /tmp/vmware-*
```
agrare, line 22: @bdunne could this live in a helper script that also lives in `container-assets/` like `container-assets/install-vmdk` so that we don't have to one-line this relatively complex bash statement?

images/manageiq-base-worker/Dockerfile
Fryguy: Maybe use a multi-stage build and COPY it into the first stage?

images/manageiq-base-worker/Dockerfile
chessbyte: should it not be `container-assets/install-vddk` or `container-assets/install-vmware-vddk`?

agrare: :+1: yes `s/vmdk/vddk/` was a typo

images/manageiq-base-worker/container-assets/extract-vmware-vddk
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+cd /vddk
+filename=$(ls VMware-*)
+
+# Skip the install if the VDDK library is not found
+[[ -f $filename ]] && tar -zxvf VMware-* || mkdir vmware-vix-disklib-distrib
```
agrare, line 7: Looked weird to create an empty directory if there were no vddk files but I see you need it so that https://github.com/ManageIQ/manageiq-pods/pull/666/files#diff-cd386fc86f89ae5eef0d1a52e06f621b5d8286f39ddc055e1cfaa7229b6316d9R20 will work

bdunne: Yeah, I don't really like it either, but it seemed like the cleanest approach

agrare (approved):
:+1: LGTM

PR #670, opened 2021-01-18
Switch NetworkPolicies from extensions/v1beta1 to networking/v1

bdunne:
Fixes `no matches for kind "NetworkPolicy" in version "extensions/v1beta1"`


abellotti (approved):

abellotti:
LGTM!!  Tested against my 4.5.20 cluster. 👍

Fryguy:
Was this never working, or it worked on an older Kube?  Do we have to support multiple/old kube versions (or openshift which lags)?

abellotti:
@bdunne I think we also need to update the role, networkpolicies is under the apiGroups extensions.

abellotti:
@Fryguy this worked fine before.  The networking api existed in both extensions/v1beta1 as well as networking/v1.  It's just that in the newer OCP clusters, they removed it from the extensions/v1beta1.   (lucky me kicking the tire with master and OCP 4.5.20).  Also, @bdunne found out that the api was in both for a long time (4 years +-) so it's safe to just shift to the new one.

PR #671, opened 2021-01-22
Added support for Managed OIDC Configurations

abellotti:
Added Support for Managed OIDC Configurations.

With Managed OIDC configurations, parent managing applications, can rewire and reconfigure all OIDC client registrations against an identity provider, then provide us the updated parameters via the HttpdAuthConfig secret.

With this PR, the operator would define the following httpd pod environment variable:

- MANAGED_HTTPD_CFG_VERSION

to reflect the latest resourceVersion of the provided HttpdAuthConfig secret. 
Upon an update to the secret, the Httpd pod environment will get updated and then restarted by Openshift for Apache to honor the new configuration.

- [x] Tested

<strike>Added support for Managed OIDC Configurations.

Managed environments provide the configured OIDC parameters as secret key/value pairs in the HttpdAuthConfig secret
as follows:

- managed.cfg.client_id       (OIDC Client ID)
- managed.cfg.client_secret   (OIDC Client Secret)
- managed.cfg.app_domain      (Application Domain for ManageIQ)
- managed.cfg.mgmt_domain     (Domain for the parent management application)

These configurations get defined as httpd pod environment variables:

- MANAGED_CFG_CLIENT_ID
- MANAGED_CFG_CLIENT_SECRET
- MANAGED_CFG_APP_DOMAIN
- MANAGED_CFG_MGMT_DOMAIN

Parent managing application can rewire and reconfigure all OIDC client registrations against an identity provider and then provide the updated parameters via the HttpdAuthConfig secret.

Upon updating these pod environment variables, the HTTPD pod will automatically be restarted by Openshift for Apache to honor the new OIDC configuration.
</strike>

### DIFF
--- a/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
+++ b/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
@@ -245,7 +245,7 @@ func (r *ReconcileManageIQ) generateHttpdResources(cr *miqv1alpha1.ManageIQ) err
 		}
 	}
 
-	httpdDeployment, mutateFunc, err := miqtool.HttpdDeployment(cr, r.scheme)
+	httpdDeployment, mutateFunc, err := miqtool.HttpdDeployment(r.client, cr, r.scheme)
 	if err != nil {
 		return err
 	}
@@ -289,7 +289,7 @@ func (r *ReconcileManageIQ) generateMemcachedResources(cr *miqv1alpha1.ManageIQ)
 }
 
 func (r *ReconcileManageIQ) generatePostgresqlResources(cr *miqv1alpha1.ManageIQ) error {
-	hostName := getSecretKeyValue(r.client, cr.Namespace, cr.Spec.DatabaseSecret, "hostname")
+	hostName := miqtool.GetSecretKeyValue(r.client, cr.Namespace, cr.Spec.DatabaseSecret, "hostname")
 	if hostName != "" {
 		logger.Info("External PostgreSQL Database selected, skipping postgresql service reconciliation", "hostname", hostName)
 		return nil
@@ -336,7 +336,7 @@ func (r *ReconcileManageIQ) generatePostgresqlResources(cr *miqv1alpha1.ManageIQ
 }
 
 func (r *ReconcileManageIQ) generateKafkaResources(cr *miqv1alpha1.ManageIQ) error {
-	hostName := getSecretKeyValue(r.client, cr.Namespace, cr.Spec.KafkaSecret, "hostname")
+	hostName := miqtool.GetSecretKeyValue(r.client, cr.Namespace, cr.Spec.KafkaSecret, "hostname")
 	if hostName != "" {
 		logger.Info("External Kafka Messaging Service selected, skipping kafka and zookeeper service reconciliation", "hostname", hostName)
 		return nil
@@ -590,14 +590,4 @@ func (r *ReconcileManageIQ) createk8sResIfNotExist(cr *miqv1alpha1.ManageIQ, res
 		return err
 	}
 	return nil
-}
-
-func getSecretKeyValue(client client.Client, nameSpace string, secretName string, keyName string) string {
-	secretKey := types.NamespacedName{Namespace: nameSpace, Name: secretName}
-	secret := &corev1.Secret{}
-	secretErr := client.Get(context.TODO(), secretKey, secret)
-	if secretErr != nil {
-		return ""
-	}
-	return string(secret.Data[keyName])
 }
--- a/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
+++ b/manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
@@ -2,6 +2,7 @@ package manageiq
 
 import (
 	"context"
+	"strings"
 
 	miqv1alpha1 "github.com/ManageIQ/manageiq-pods/manageiq-operator/pkg/apis/manageiq/v1alpha1"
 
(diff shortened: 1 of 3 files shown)


manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -176,6 +177,18 @@ func addOIDCEnv(secretName string, podSpec *corev1.PodSpec) {
 	podSpec.Containers[0].Env = append(podSpec.Containers[0].Env, clientSecret)
 }
 
+func configureHttpdOIDCEnv(podSpec *corev1.PodSpec, managed_client_id string, managed_client_secret string, managed_app_domain string, managed_mgmt_domain string) {
+	managedCfgClientId := corev1.EnvVar{Name: "MANAGED_CFG_CLIENT_ID", Value: managed_client_id}
```
bdunne, line 181: I think we should we be using `ValueFrom` here instead to pull the ENV var from the secret rather than creating a clear text ENV var with the data. (I'm assuming that some of this is sensitive data, like `managed_client_secret`)

bdunne (changes requested):

manageiq-operator/pkg/helpers/miq-components/httpd.go
abellotti: Yeah, I had ValueFrom originally, however, Kubernetes does not see that as a change when the value is updated in the secret, so the Httpd pod will not restart. Those values accessible to an admin as they exist in the /etc/httpd/user-conf.d/oidc.conf file.

bdunne: It's clear text to anyone that can `kubectl describe pod httpd-zzzzz` as well.

abellotti: @bdunne new code pushed up, commit https://github.com/ManageIQ/manageiq-pods/pull/671/commits/464c84284efe94b0b85ece3cf432f2be0c5ba4d7 has the new logic for the secret watcher. I've added  predicate functions on the events to only trigger reconciliation on the httpd auth secret since we skip non-CR request reconciliations.

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -175,6 +178,23 @@ func addOIDCEnv(secretName string, podSpec *corev1.PodSpec) {
...
+		if secretErr == nil {
+			httpd_auth_config_version = string(secret.GetObjectMeta().GetResourceVersion())
+		}
+	}
+	return httpd_auth_config_version
+}
+
+func setManagedHttpdCfgVersion(httpdAuthConfigVersion string, podSpec *corev1.PodSpec) {
+	managedHttpdCfgRevision := corev1.EnvVar{Name: "MANAGED_HTTPD_CFG_VERSION", Value: httpdAuthConfigVersion}
+	podSpec.Containers[0].Env = append(podSpec.Containers[0].Env, managedHttpdCfgRevision)
```
bdunne, line 195: Should we add a comment like `// This is not used by the pod, only to trigger a redeployment if the secret has changed`?

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -540,6 +577,52 @@ func (r *ReconcileManageIQ) manageCR(cr *miqv1alpha1.ManageIQ) error {
...
+				Namespace: cr.Namespace,
+				Name:      cr.Spec.HttpdAuthConfig,
+			},
+		}}, &handler.EnqueueRequestsFromMapFunc{
+			ToRequests: mapFn,
+		}, predicateFn)
+
+		if err != nil {
+			return err
+		}
```
bdunne, line 621: What do you think of something like this:
```suggestion
    authConfigSecret := &corev1.Secret{
      ObjectMeta: metav1.ObjectMeta{
        Namespace: cr.Namespace,
        Name:      cr.Spec.HttpdAuthConfig,
      },
    }

    mapFunc := &handler.EnqueueRequestsFromMapFunc{
      ToRequests: mapFn,
    }

    if err := controller_manageiq.Watch(&source.Kind{Type: authConfigSecret}, mapFunc, predicateFn); err != nil {
      return err
    }
```

manageiq-operator/pkg/controller/manageiq/manageiq_controller.go
```diff
@@ -101,6 +106,26 @@ func (r *ReconcileManageIQ) Reconcile(request reconcile.Request) (reconcile.Resu
...
 
+	if strings.HasPrefix(request.Name, "httpd-auth-config-secret-") {
+		// Mapped secret name coming as httpd-auth-config-secret-<miq_instance_name>-<secret_name>
+		rString := strings.TrimPrefix(request.Name, "httpd-auth-config-secret-")
+		nSlices := strings.SplitN(rString, "-", 2)
+		// Fetch the ManageIQ instance
+		miqInstance := &miqv1alpha1.ManageIQ{}
+		err := r.client.Get(context.TODO(), types.NamespacedName{
+			Namespace: request.Namespace,
+			Name:      nSlices[0],
```
bdunne, line 117: What do you think of this?
```suggestion
		nSlices := strings.SplitN(rString, "-", 6)
		// Fetch the ManageIQ instance
		miqInstance := &miqv1alpha1.ManageIQ{}
		err := r.client.Get(context.TODO(), types.NamespacedName{
			Namespace: request.Namespace,
			Name:      nSlices[4],
```

bdunne (approved):

Fryguy:
Backported to `lasker` in commit 32b7e4d92213cb16b10dde0e684e0813c7cf3c04.

```text
commit 32b7e4d92213cb16b10dde0e684e0813c7cf3c04
Date:   Thu Feb 25 14:49:26 2021 -0500

    Merge pull request #671 from abellotti/support_managed_oidc_configuration
    
    Added support for Managed OIDC Configurations
    
    (cherry picked from commit 2523904e5814d042721198b82d64aa2ec426eb25)
```

Issue #652, opened 2020-11-11
postgresql secret needs to support SSL certs


Fryguy:
If connecting to an external Postgres, there seems to be no way to specify SSL certs for the connection.  If possible, can we extend the secret to include certs, or perhaps another secret that holds the certs?

For example, if I want to point to an IBM Cloud database, then the parameters that you need for connecting and using IBM Databases for PostgreSQL

- Hostname/address
- Port
- Maintenance database : postgres
- Username
- Password
- SSL mode : Verify-Full
- Root/SSL/TLS certificate

All of the above is readily available once you provision IBM Databases for PostgreSQL service.
_________________
The current data section of the **PostgreSQL secret** contains 
-   dbname:
-   hostname: 
-   password: 
-   port: 
-   username: 

Support would be needed for the remaining Parameters. 

Doc Link:- https://cloud.ibm.com/docs/databases-for-postgresql?topic=databases-for-postgresql-getting-started

@miq-bot add-label enhancement

PRs:
- [ ] Operator configure the orchestrator #674 
- [ ] Orchestrator configure the workers https://github.com/ManageIQ/manageiq/pull/21005

NickLaMuro:
So I did a little research into adding SSL cert support in our pods, and here is what I found:

For connecting to the database, we override the `config/database.yml` using the `DATABASE_URL` environment variable, configured here:

https://github.com/ManageIQ/manageiq-pods/blob/b856d2d3/images/manageiq-base/container-assets/container_env#L12

So any change necessary to support this will most likely entirely done on the `manageiq-pods` repo, with some supporting work via the operator to write the cert to a file system that the Rails process has access to (which has been done for other items from what I have been told).

The two attributes that need to be add to the `DATABASE_URL` are:

- `sslmode`
- `sslrootcert` (based on the ibm docs)

The `sslmode` is something specific to PG, and _should_ get passed along to the `pg` gem from the `DATABASE_URL` settings that are parsed or the `database.yml` (preference first given to `DATABASE_URL`, which is what we use in pods).

Specifically, the value for `sslmode` is assumed to be `verify-full` in this configuration, as defined by the PG docs:

https://www.postgresql.org/docs/10/libpq-ssl.html#LIBQ-SSL-CERTIFICATES

which is capitalized in the IBM docs when working with `pgAdmin`, but I assume that doesn't matter.

[...]

PR #674, opened 2021-02-02
Support Postgres with SSL certificates

bdunne:
part of #652

### DIFF
--- a/images/manageiq-base/container-assets/container_env
+++ b/images/manageiq-base/container-assets/container_env
@@ -9,4 +9,7 @@ function urlescape() {
 safeuser=$(urlescape ${DATABASE_USER})
 safepass=$(urlescape ${DATABASE_PASSWORD})
 
-export DATABASE_URL="postgresql://${safeuser}:${safepass}@${DATABASE_HOSTNAME}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5"
+database_url="postgresql://${safeuser}:${safepass}@${DATABASE_HOSTNAME}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5&sslmode=${DATABASE_SSL_MODE:-prefer}"
+[[ -f /.postgresql/postgresql.crt ]] && database_url="$database_url&sslrootcert=/.postgresql/postgresql.crt"
+
+export DATABASE_URL=$database_url
--- a/manageiq-operator/README.md
+++ b/manageiq-operator/README.md
@@ -110,7 +110,17 @@ $ podman push docker.io/example/manageiq-bundle:0.0.1
 
 ### Configuring external Postgres
 
-ManageIQ can be run with an external Postgres server. To do this, create the required OpenShift secret with the correct parameters using the template for [Postgres](templates/app/postgresql-secrets.yaml) and provide those secret names as `databaseSecret` in `manageiq.org_v1alpha1_manageiq_cr.yaml`.
+ManageIQ can be run with an external Postgres server. To do this, manually create the postgresql secret:
+```
+$ oc create secret generic postgresql-secrets \
+  --from-literal=dbname=vmdb_production \
+  --from-literal=hostname=YOUR_HOSTNAME \
+  --from-literal=port=5432 \
+  --from-literal=password=YOUR_PASSWORD_HERE \
+  --from-literal=username=YOUR_USERNAME_HERE \
+  --from-literal=sslmode=verify-full \ # optional
+  --from-file=rootcertificate=path/to/optional/certificate.pem # optional
+```
 
 ### Configuring external messaging
 
--- a/manageiq-operator/pkg/helpers/miq-components/orchestrator.go
+++ b/manageiq-operator/pkg/helpers/miq-components/orchestrator.go
@@ -159,6 +159,16 @@ func addMessagingEnv(cr *miqv1alpha1.ManageIQ, c *corev1.Container) {
 	return
 }
 
+func addPostgresEnv(cr *miqv1alpha1.ManageIQ, c *corev1.Container) {
+	c.Env = append(c.Env, corev1.EnvVar{Name: "DATABASE_REGION",   Value: cr.Spec.DatabaseRegion})
+
+	c.Env = append(c.Env, corev1.EnvVar{Name: "DATABASE_HOSTNAME", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: postgresqlSecretName(cr)}, Key: "hostname"}}})
+	c.Env = append(c.Env, corev1.EnvVar{Name: "DATABASE_NAME",     ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: postgresqlSecretName(cr)}, Key: "dbname"}}})
+	c.Env = append(c.Env, corev1.EnvVar{Name: "DATABASE_PASSWORD", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: postgresqlSecretName(cr)}, Key: "password"}}})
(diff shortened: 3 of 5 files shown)


manageiq-operator/README.md
```diff
@@ -110,7 +110,17 @@ $ podman push docker.io/example/manageiq-bundle:0.0.1
...
+```
+$ oc create secret generic postgresql-secrets \
+  --from-literal=dbname=vmdb_production \
+  --from-literal=hostname=YOUR_HOSTNAME \
+  --from-literal=port=5432 \
+  --from-literal=password=YOUR_PASSWORD_HERE \
+  --from-literal=username=YOUR_USERNAME_HERE \
+  --from-literal=sslmode=verify-full \ # optional
+  --from-file=rootcertificate=path/to/optional/certificate.pem # optional
+  ```
```
Fryguy, line 123: Minor, but the triple backticks are not aligned.

manageiq-operator/pkg/helpers/miq-components/orchestrator.go
```diff
@@ -159,6 +160,35 @@ func addMessagingEnv(cr *miqv1alpha1.ManageIQ, c *corev1.Container) {
...
+	d.Spec.Template.Spec.Containers[0].Env = append(d.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: "DATABASE_USER", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: postgresqlSecretName(cr)}, Key: "username"}}})
+
+	pgSecret := postgresqlSecret(cr, client)
+	if pgSecret.Data["sslmode"] != nil {
+		d.Spec.Template.Spec.Containers[0].Env = append(d.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: "DATABASE_SSL_MODE", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: postgresqlSecretName(cr)}, Key: "sslmode"}}})
+	}
+	if pgSecret.Data["rootcertificate"] != nil {
+		d.Spec.Template.Spec.Containers[0].VolumeMounts = []corev1.VolumeMount{corev1.VolumeMount{Name: "miq-pg-root-certificate", MountPath: "/.postgresql", ReadOnly: true}}
```
Fryguy, line 177: Does the `miq-` bit have to be in the name? (or is that the volume name from within the container)?

manageiq-operator/pkg/helpers/miq-components/orchestrator.go
```diff
@@ -159,6 +160,35 @@ func addMessagingEnv(cr *miqv1alpha1.ManageIQ, c *corev1.Container) {
...
+	}
+}
+
+func postgresqlSecret(cr *miqv1alpha1.ManageIQ, client client.Client) *corev1.Secret {
+	secretKey := types.NamespacedName{Namespace: cr.Namespace, Name: postgresqlSecretName(cr)}
+	secret := &corev1.Secret{}
+	client.Get(context.TODO(), secretKey, secret)
+
+	return secret
+}
```
Fryguy, line 190: Should this be in the other file where we define all of the secrets?

Fryguy:
Overall LGTM...just some minor things

images/manageiq-base/container-assets/container_env
```diff
@@ -9,4 +9,8 @@ function urlescape() {
 safeuser=$(urlescape ${DATABASE_USER})
 safepass=$(urlescape ${DATABASE_PASSWORD})
 
-export DATABASE_URL="postgresql://${safeuser}:${safepass}@${DATABASE_HOSTNAME}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5"
+ssl_mode=${DATABASE_SSL_MODE:-prefer}
+database_url="postgresql://${safeuser}:${safepass}@${DATABASE_HOSTNAME}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5&sslmode=${ssl_mode}"
+[[ -f /.postgresql/postgresql.crt ]] && database_url="$database_url&sslrootcert=/.postgresql/postgresql.crt"
```
NickLaMuro, line 14: To avoid extra "`if`" statements in this file (which I find hard to parse\*), does it make sense try using what was already available for environment variables from postgres?

> The location of the root certificate file and the CRL can be changed by setting the connection parameters `sslrootcert` and `sslcrl` or the environment variables `PGSSLROOTCERT` and `PGSSLCRL`.

(info pull from here:  https://www.postgresql.org/docs/10/libpq-ssl.html#LIBQ-SSL-CERTIFICATES )

I think there is a case to be made that it makes it more clear to do all variables in one fashion, so this isn't a hard stance I am making.  That said, it seems like instead of adding more logic here, it makes more sense to just add another environment variable to the process if `"rootcertificate"` is detected in the `orchestrator.go ` file.

* * *

If we do go this route, we could add a comment here mentioning that root certs are defined via the `orchestrator.go`, and that will set `PGSSLROOTCERT` for you.  Less places the path for the cert is hard coded as well (would only be in the `orchestrator` that way.

NickLaMuro (commented):
I think this is fine based on what little I know about the `orchestrator.go` code base.

I really can't comment on elongated lines like this:

```diff
+ d.Spec.Template.Spec.Containers[0].Env = append(d.Spec.Template.Spec.Containers[0].Env, corev1.EnvVar{Name: "DATABASE_HOSTNAME", ValueFrom: &corev1.EnvVarSource{SecretKeyRef: &corev1.SecretKeySelector{LocalObjectReference: corev1.LocalObjectReference{Name: postgresqlSecretName(cr)}, Key: "hostname"}}})
```

Since I have next to zero knowledge about the libraries and facilities they are calling into, but seems it is mostly like it is just some refactoring, and a bit of IO tweaking for how we are setting configuring the orchestrator (correct me if I am wrong)

images/manageiq-base/container-assets/container_env
```diff
@@ -9,4 +9,8 @@ function urlescape() {
 safeuser=$(urlescape ${DATABASE_USER})
 safepass=$(urlescape ${DATABASE_PASSWORD})
 
-export DATABASE_URL="postgresql://${safeuser}:${safepass}@${DATABASE_HOSTNAME}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5"
+ssl_mode=${DATABASE_SSL_MODE:-prefer}
+database_url="postgresql://${safeuser}:${safepass}@${DATABASE_HOSTNAME}:${DATABASE_PORT}/${DATABASE_NAME}?encoding=utf8&pool=5&wait_timeout=5&sslmode=${ssl_mode}"
```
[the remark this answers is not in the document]
NickLaMuro: At a minimum, we probably should move this up to the be next to the `safeuser` and `safepass` variable declarations.

images/manageiq-base/container-assets/container_env
bdunne: I'm guessing we don't need to set `PGSSLROOTCERT` since I'm dropping it in the default location.

I figured that if this were on an appliance we would set this in `database.yml` anyway, so I put it in `DATABASE_URL` for consistency.

manageiq-operator/pkg/helpers/miq-components/orchestrator.go
bdunne: Moved

manageiq-operator/pkg/helpers/miq-components/orchestrator.go
bdunne: Done.

manageiq-operator/README.md
bdunne: Fixed

images/manageiq-base/container-assets/container_env
NickLaMuro: > I figured that if this were on an appliance we would set this in `database.yml` anyway, so I put it in `DATABASE_URL` for consistency.

This rational makes sense, and I am cool with that.  I wasn't hard set on my position here anyway, so I am fine with this staying as it is.

NickLaMuro (approved):
My concerns have been addressed, and everything else is probably out of my wheelhouse to be able to comment on properly.  :shipit:

PR #675, opened 2021-02-02
Adding support for custom images

abellotti:
Adding support for custom image specifications via environment variables:
      - HTTPD_IMAGE
      - KAFKA_IMAGE
      - ZOOKEEPER_IMAGE
      - MEMCACHED_IMAGE
      - BASEWORKER_IMAGE
      - WEBSERVERWORKER_IMAGE
      - UIWORKER_IMAGE
      - ORCHESTRATOR_IMAGE
      - POSTGRESQL_IMAGE

Memcached and Postgresql PodSpec now define the default ServiceAccountName

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/httpd.go
+++ b/manageiq-operator/pkg/helpers/miq-components/httpd.go
@@ -13,6 +13,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	intstr "k8s.io/apimachinery/pkg/util/intstr"
 	"net/http"
+	"os"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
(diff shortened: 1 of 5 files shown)


abellotti:
Hi @bdunne, I found this set of change in the downstream operator that were added. They seem benign to bring back to upstream and help us minimizing the differences.  Let me know.  Thanks.

Fryguy:
Big :+1: from me for custom images!

Fryguy:
Can you update the PR title to reflect the actual change (and not the "why" you made the change)?

Fryguy (approved):

bdunne (changes requested):
So, I'm 👍🏻 for bringing as much as possible to upstream, but  👎🏻 on this particular change.

What is the use case for this?  I see too many problems with using ENV vars:
- I could see users being extremely confused if they specify a value for images in the CR and then get something else when the operator creates the deployments. (because of an ENV var on the operator)
- I'd prefer to have a single place to specify the image names (CR).
- For example: If you want to update a single deployment (say httpd) in this scenario, you would have to edit the operator deployment which would bounce the operator pod which would then deploy the new httpd pod. Whereas, the CR could be edited and the operator would update the httpd deployment (no need to edit the operator deployment and bounce the operator pod).

abellotti:
Closing this, as it would only be needed in an installer and not the manageiq-operator.

PR #677, opened 2021-02-04
Added support for AppAnnotations

abellotti:
Adding support for AppAnnotations

- An optional String map object,  If specified, the annotations will be applied
  to the Httpd, Kafka, Memcached, Orchestrator and PostgresQL Pods.


manageiq-operator/pkg/helpers/miq-components/memcached.go
```diff
@@ -80,6 +80,9 @@ func NewMemcachedDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*
 			},
 		},
 	}
+	if len(cr.Spec.AppAnnotations) > 0 {
```
bdunne, line 83: Should this be in the reconcile loop like the others?

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -233,6 +233,9 @@ func PostgresqlDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*ap
 			},
 		},
 	}
+	if len(cr.Spec.AppAnnotations) > 0 {
```
bdunne, line 236: Should this be in the reconcile loop like the others?

manageiq-operator/pkg/helpers/miq-components/kafka.go
```diff
@@ -336,6 +336,9 @@ func ZookeeperDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*app
 		deployment.Spec.Strategy = appsv1.DeploymentStrategy{
 			Type: "Recreate",
 		}
+		if len(cr.Spec.AppAnnotations) > 0 {
```
bdunne, line 339: This is going to overwrite the backup annotations on https://github.com/ManageIQ/manageiq-pods/pull/677/files#diff-e23faf1d310cc4c5d8c16370e2cf9c9544e328f6dd349865dcbf585914565d51R331

abellotti: Good Catch 👍. I was only looking at the openshift.io/scc annotation which was there regardless so I thought I wasn't overwriting.    I'll update the PR to do a merge instead.  Thanks!!

abellotti: Ok, fix in this commit https://github.com/ManageIQ/manageiq-pods/pull/677/commits/ece18bd5433e970f98e8877a58f081886c16095c. I followed the pattern of the other methods, add*, and generalized it for Annotations. So it can also be used for other cases.   Thanks for catching this.

bdunne (approved):

Issue #551, opened 2020-06-17
Image parameters for the images deployed by the operator should be consistent


Fryguy:
While not a bug, per se, the inconsistency is a little strange.  Sometimes you specify the namespace separate, sometimes it's together, and sometimes it has docker.io as the default whereas other times it doesn't.  Not sure what the _right_ consistency would look like whether a single string we parse, or every possible piece as a separate value.

- [httpd](https://github.com/ManageIQ/manageiq-pods/blob/1e9c2d46f80a06d7cd2d609006249776eca2fae4/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go#L288-L294)
- [memcached](https://github.com/ManageIQ/manageiq-pods/blob/1e9c2d46f80a06d7cd2d609006249776eca2fae4/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go#L309-L315)
- [orchestrator itself](https://github.com/ManageIQ/manageiq-pods/blob/1e9c2d46f80a06d7cd2d609006249776eca2fae4/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go#L329-L339)
- [postgres](https://github.com/ManageIQ/manageiq-pods/blob/1e9c2d46f80a06d7cd2d609006249776eca2fae4/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go#L345-L351)
- [kafka](https://github.com/ManageIQ/manageiq-pods/blob/1e9c2d46f80a06d7cd2d609006249776eca2fae4/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go#L345-L351)
- [zookeeper](https://github.com/ManageIQ/manageiq-pods/blob/1e9c2d46f80a06d7cd2d609006249776eca2fae4/manageiq-operator/pkg/apis/manageiq/v1alpha1/manageiq_types.go#L377-L383)

@miq-bot add-label enhancement

jrafanie:
Along these lines, I had to push images and try to deploy them using my CR and it was trial and error to get it working and making these consistent would make this a lot easier.  

Additionally, extracting the default values out of the go code and into a template CR file that you can make a copy and edit would make the process of customizing the images you want to deploy much easier.

Here's the CR that I was able to get to work:

```
apiVersion: manageiq.org/v1alpha1
kind: ManageIQ
metadata:
  name: miq
spec:
[...]

PR #680, opened 2021-02-08
Adding support for optional Image References

abellotti:
Adding support for optional Image References instead of the ImageName and ImageTag parameter pairs. If not specified,
then the default ImageName:ImageTag image reference is used.

New optional CR parameters:

  - PostgresqlImage
  - HttpdImage
  - OrchestratorImage
  - MemcachedImage
  - KafkaImage
  - ZookeeperImage

Todo:

- [x] Test with Meta operator

### DIFF
--- a/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
+++ b/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
@@ -88,7 +88,7 @@ spec:
               description: 'Httpd deployment CPU request (default: no request)'
               type: string
             httpdImage:
-              description: 'Image string used for the httpd deployment (default: <HttpdImageNamespace>:<HttpdImageTag>)'
+              description: 'Image string used for the httpd deployment (default: <HttpdImageNamespace>/httpd[-init]:<HttpdImageTag>)'
               type: string
             httpdImageNamespace:
               description: 'Image namespace used for the httpd deployment (default:
@@ -199,7 +199,7 @@ spec:
               type: string
             orchestratorImage:
               description: 'Image string used for the orchestrator deployment (default:
-                <OrchestratorImageName>:<OrchestratorImageTag>)'
+                <OrchestratorImageNamespace>/<OrchestratorImageName>:<OrchestratorImageTag>)'
               type: string
             orchestratorImageName:
               description: 'Image name used for the orchestrator deployment (default:
--- a/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
+++ b/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
@@ -88,8 +88,7 @@ spec:
               description: 'Httpd deployment CPU request (default: no request)'
               type: string
             httpdImage:
-              description: Image string used for the httpd deployment By default HttpdImageNamespace:HttpdImageTag
-                takes effect
+              description: 'Image string used for the httpd deployment (default: <HttpdImageNamespace>:<HttpdImageTag>)'
               type: string
             httpdImageNamespace:
               description: 'Image namespace used for the httpd deployment (default:
@@ -120,8 +119,7 @@ spec:
               description: 'Kafka deployment CPU limit (default: no limit)'
               type: string
             kafkaImage:
-              description: Image string used for the kafka deployment By default KafkaImageName:KafkaImageTag
-                takes effect
+              description: 'Image string used for the kafka deployment (default: <KafkaImageName>:<KafkaImageTag>)'
               type: string
             kafkaImageName:
               description: 'Image used for the kafka deployment (default: docker.io/bitnami/kafka)'
@@ -149,8 +147,8 @@ spec:
               description: 'Memcached deployment CPU request (default: no request)'
               type: string
             memcachedImage:
-              description: Image string used for the memcached deployment By default
-                MemcachedImageName:MemcachedImageTag takes effect
+              description: 'Image string used for the memcached deployment (default:
+                <MemcachedImageName>:<MemcachedImageTag>)'
               type: string
             memcachedImageName:
               description: 'Image used for the memcached deployment (default: manageiq/memcached)'
@@ -200,8 +198,8 @@ spec:
               description: 'Orchestrator deployment CPU request (default: no request)'
               type: string
             orchestratorImage:
-              description: Image string used for the orchestrator deployment By default
-                OrchestratorImageName:OrchestratorImageTag takes effect
+              description: 'Image string used for the orchestrator deployment (default:
+                <OrchestratorImageName>:<OrchestratorImageTag>)'
               type: string
             orchestratorImageName:
               description: 'Image name used for the orchestrator deployment (default:
@@ -232,8 +230,8 @@ spec:
               description: 'PostgreSQL deployment CPU request (default: no request)'
               type: string
             postgresqlImage:
-              description: Image string used for the postgresql deployment By default
-                PostgresqlImageName:PostGresqlImageTag takes effect
+              description: 'Image string used for the postgresql deployment (default:
+                <PostgresqlImageName>:<PostgresqlImageTag>)'
               type: string
             postgresqlImageName:
               description: 'Image used for the postgresql deployment (Default: docker.io/manageiq/postgresql)'
@@ -279,8 +277,8 @@ spec:
               description: 'Zookeeper deployment CPU limit (default: no limit)'
               type: string
             zookeeperImage:
-              description: Image string used for the zookeeper deployment By default
-                ZookeeperImageName:ZookeeperImageTag takes effect
+              description: 'Image string used for the zookeeper deployment (default:
+                <ZookeeperImageName>:<ZookeeperImageTag>)'
               type: string
             zookeeperImageName:
               description: 'Image used for the zookeeper deployment (default: docker.io/bitnami/zookeeper)'
--- a/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
+++ b/manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
@@ -87,6 +87,10 @@ spec:
             httpdCpuRequest:
               description: 'Httpd deployment CPU request (default: no request)'
               type: string
+            httpdImage:
+              description: Image string used for the httpd deployment By default HttpdImageNamespace:HttpdImageTag
+                takes effect
+              type: string
             httpdImageNamespace:
               description: 'Image namespace used for the httpd deployment (default:
                 manageiq) Note: the exact image will be determined by the authentication
@@ -115,6 +119,10 @@ spec:
             kafkaCpulimit:
               description: 'Kafka deployment CPU limit (default: no limit)'
               type: string
+            kafkaImage:
+              description: Image string used for the kafka deployment By default KafkaImageName:KafkaImageTag
+                takes effect
+              type: string
             kafkaImageName:
               description: 'Image used for the kafka deployment (default: docker.io/bitnami/kafka)'
               type: string
@@ -139,6 +147,10 @@ spec:
               type: string
             memcachedCpuRequest:
               description: 'Memcached deployment CPU request (default: no request)'
+              type: string
+            memcachedImage:
+              description: Image string used for the memcached deployment By default
+                MemcachedImageName:MemcachedImageTag takes effect
               type: string
             memcachedImageName:
               description: 'Image used for the memcached deployment (default: manageiq/memcached)'
@@ -187,6 +199,10 @@ spec:
             orchestratorCpuRequest:
               description: 'Orchestrator deployment CPU request (default: no request)'
               type: string
+            orchestratorImage:
+              description: Image string used for the orchestrator deployment By default
+                OrchestratorImageName:OrchestratorImageTag takes effect
+              type: string
             orchestratorImageName:
               description: 'Image name used for the orchestrator deployment (default:
                 manageiq-orchestrator)'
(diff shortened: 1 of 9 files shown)


bdunne:
@abellotti This moves toward fixing https://github.com/ManageIQ/manageiq-pods/issues/551 right?

abellotti:
wow, didn't even know about this issue, but yes, it would solve it.

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -243,8 +243,13 @@ func configureHttpdAuth(spec *miqv1alpha1.ManageIQSpec, podSpec *corev1.PodSpec)
 	}
 }
 
-func httpdImage(namespace, tag string, privileged bool) string {
+func httpdImage(imageref string, namespace string, tag string, privileged bool) string {
```
bdunne, line 246: Can this logic be moved into `cr.go`?  Then all callers can use `cr.Spec.HttpdImage`

manageiq-operator/pkg/helpers/miq-components/kafka.go
```diff
@@ -174,9 +174,14 @@ func KafkaDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*appsv1.
 		"app":  cr.Spec.AppName,
 	}
 
+	kafkaImage := cr.Spec.KafkaImage
+	if kafkaImage == "" {
```
bdunne, line 178: Same here

manageiq-operator/pkg/helpers/miq-components/kafka.go
```diff
@@ -279,9 +284,14 @@ func ZookeeperDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*app
 		"app":  cr.Spec.AppName,
 	}
 
+	zookeeperImage := cr.Spec.ZookeeperImage
+	if zookeeperImage == "" {
```
bdunne, line 288: And here...

manageiq-operator/pkg/helpers/miq-components/memcached.go
```diff
@@ -11,9 +11,14 @@ import (
 )
 
 func NewMemcachedDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*appsv1.Deployment, controllerutil.MutateFn, error) {
+	memcachedImage := cr.Spec.MemcachedImage
+	if memcachedImage == "" {
```
bdunne, line 15: And here

manageiq-operator/pkg/helpers/miq-components/orchestrator.go
```diff
@@ -213,9 +213,14 @@ func OrchestratorDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme, cl
 		"app":  cr.Spec.AppName,
 	}
 
+	orchestratorImage := cr.Spec.OrchestratorImage
+	if orchestratorImage == "" {
```
bdunne, line 217: and here

manageiq-operator/pkg/helpers/miq-components/postgresql.go
```diff
@@ -149,9 +149,14 @@ func PostgresqlDeployment(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*ap
 	}
 	var initialDelaySecs int32 = 60
 
+	postgresqlImage := cr.Spec.PostgresqlImage
+	if postgresqlImage == "" {
```
bdunne, line 153: and here

bdunne (changes requested):

manageiq-operator/pkg/helpers/miq-components/kafka.go
abellotti: As an example, you mean have cr.go define cr.Spec.KafkaImage if not specified and built it based on the image parts ?  i.e. cr.Spec.KafkaImage = cr.Spec.KafkaImageName + ":" + cr.Spec.KafkaImageTag ?   So code here below would just use cr.Spec.KafkaImage (now guaranteed to be defined) ?

bdunne: exactly

manageiq-operator/pkg/helpers/miq-components/cr.go
```diff
@@ -298,26 +354,38 @@ func ManageCR(cr *miqv1alpha1.ManageIQ, c *client.Client) (*miqv1alpha1.ManageIQ
 		cr.Spec.HttpdAuthenticationType = httpdAuthenticationType(cr)
 		cr.Spec.HttpdImageNamespace = httpdImageNamespace(cr)
 		cr.Spec.HttpdImageTag = httpdImageTag(cr)
+
+		privileged, err := PrivilegedHttpd(cr.Spec.HttpdAuthenticationType)
```
bdunne, line 358: Should this be in the httpdImage method?

manageiq-operator/pkg/helpers/miq-components/cr.go
```diff
@@ -97,6 +97,22 @@ func httpdImageTag(cr *miqv1alpha1.ManageIQ) string {
...
 
+func httpdImage(cr *miqv1alpha1.ManageIQ, privileged bool) string {
+	if cr.Spec.HttpdImage == "" {
+		var image string
+
+		if privileged {
+			image = "httpd-init"
+		} else {
+			image = "httpd"
+		}
```
bdunne, line 108: Would this be simpler?
```suggestion
		if privileged {
			image := "httpd-init"
		} else {
			image := "httpd"
		}
```

manageiq-operator/pkg/helpers/miq-components/cr.go
abellotti: meh, if we do, then httpdImage would need to return string, error and check for that here, so it won't be the one liners like all others here.

manageiq-operator/pkg/helpers/miq-components/cr.go
abellotti: Actually that won't work,  undefined: image because the image := are only defined within the scope of the if blocks.  we need the above var declaration. I miss Ruby.

bdunne: Oh, right. I miss Ruby too.

manageiq-operator/pkg/helpers/miq-components/cr.go
bdunne: Why would `httpImage` need to return error?

abellotti: httpdImage would need to call PriviledgedHttpd to get that privileged value for its logic and that method could return an error.

bdunne: Wouldn't we ignore the error in the `httpdImage`function just like we're ignoring it here?

bdunne: I think the "validation" that happens in `PrivilegedHttpd` which could return an error should ideally be part of the CRD Schema (although that doesn't appear to be supported) or our custom validation logic.

manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
```diff
@@ -215,6 +231,10 @@ spec:
             postgresqlCpuRequest:
               description: 'PostgreSQL deployment CPU request (default: no request)'
               type: string
+            postgresqlImage:
+              description: Image string used for the postgresql deployment By default
+                PostgresqlImageName:PostGresqlImageTag takes effect
```
Fryguy, line 236: Typo
```suggestion
                PostgresqlImageName:PostgresqlImageTag takes effect
```

(There are other locations below with a similar typo)

manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
```diff
@@ -215,6 +231,10 @@ spec:
             postgresqlCpuRequest:
               description: 'PostgreSQL deployment CPU request (default: no request)'
               type: string
+            postgresqlImage:
+              description: Image string used for the postgresql deployment By default
```
Fryguy, line 235: The pattern elsewhere is `<description> (Default: <default value>)`, so can we change it to that for consistency?  If not, there should at least be a period or comma between the description and `By default`.

Fryguy:
> @abellotti This moves toward fixing #551 right?

> wow, didn't even know about this issue, but yes, it would solve it.

I don't think it would solve it completely, but it would provide some more consistency.  To solve it completely, I think we'd have to remove all of the other component bits for each image (e.g. postgresqlImageName, postgresqlImageTag) OR support all of the same components bits for each image.

manageiq-operator/pkg/helpers/miq-components/cr.go
```diff
@@ -97,6 +97,22 @@ func httpdImageTag(cr *miqv1alpha1.ManageIQ) string {
...
+
+		return cr.Spec.HttpdImageNamespace + "/" + image + ":" + cr.Spec.HttpdImageTag
+	} else {
+		return cr.Spec.HttpdImage
+	}
+}
```
Fryguy, line 114: I feel like this would be cleaner if written in guard clause style:

```suggestion
	if cr.Spec.HttpdImage != "" {
		return cr.Spec.HttpdImage
	}
		
	var image string

	if privileged {
		image = "httpd-init"
	} else {
		image = "httpd"
	}

	return cr.Spec.HttpdImageNamespace + "/" + image + ":" + cr.Spec.HttpdImageTag
}
```

Fryguy: Same comment for all of the other places as well.

manageiq-operator/pkg/helpers/miq-components/cr.go
```diff
@@ -298,26 +354,38 @@ func ManageCR(cr *miqv1alpha1.ManageIQ, c *client.Client) (*miqv1alpha1.ManageIQ
...
+		privileged, err := PrivilegedHttpd(cr.Spec.HttpdAuthenticationType)
+		if err != nil {
+			return err
+		}
+
+		cr.Spec.HttpdImage = httpdImage(cr, privileged)
 		cr.Spec.ImagePullSecret = imagePullSecretName(cr, *c)
 		cr.Spec.KafkaImageName = kafkaImageName(cr)
 		cr.Spec.KafkaImageTag = kafkaImageTag(cr)
+		cr.Spec.KafkaImage = kafkaImage(cr)
```
Fryguy, line 367: From a consistency perspective, in the other places, KafkaImage is sorted before KafkaImageName, so we should do the same here (unless there's an actual code reason).

Same goes for the rest.

Fryguy (commented):
Overall LGTM...almost all of my comments are cosmetic.

manageiq-operator/pkg/helpers/miq-components/cr.go
abellotti: The reason I put them after the ImageName/ImageTag is that we want to default/define those first since we use them to build the *Image ones if not specified.

manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
abellotti: both good, because the default is based on the other parameters, I didn't have a value.  would something like `description (default: <postgresqlImageName>:<postgresqlImageTag>)` work ?

manageiq-operator/pkg/helpers/miq-components/cr.go
Fryguy: Ah ok...makes sense.

manageiq-operator/pkg/helpers/miq-components/cr.go
abellotti: 👍  will update for httpd because of added logic, the others follow the pattern of all others in that file, so I prefer to keep them the same for consistency.

manageiq-operator/deploy/olm-catalog/manageiq-operator/0.0.1/manageiq.org_manageiqs_crd.yaml
```diff
@@ -187,6 +197,10 @@ spec:
             orchestratorCpuRequest:
               description: 'Orchestrator deployment CPU request (default: no request)'
               type: string
+            orchestratorImage:
+              description: 'Image string used for the orchestrator deployment (default:
+                <OrchestratorImageName>:<OrchestratorImageTag>)'
```
Fryguy, line 202: I think this is slightly wrong.

```suggestion
                <OrchestratorImageNamespace>/<OrchestratorImageName>:<OrchestratorImageTag>)'
```

manageiq-operator/deploy/crds/manageiq.org_manageiqs_crd.yaml
```diff
@@ -87,6 +87,9 @@ spec:
             httpdCpuRequest:
               description: 'Httpd deployment CPU request (default: no request)'
               type: string
+            httpdImage:
+              description: 'Image string used for the httpd deployment (default: <HttpdImageNamespace>:<HttpdImageTag>)'
```
Fryguy, line 91: This one is slightly wrong too...

```suggestion
              description: 'Image string used for the httpd deployment (default: <HttpdImageNamespace>/httpd:<HttpdImageTag>)'
```

or perhaps

```suggestion
              description: 'Image string used for the httpd deployment (default: <HttpdImageNamespace>/httpd[-init]:<HttpdImageTag>)'
```

bdunne (approved):

PR #685, opened 2021-02-16
Enhance Operator documentation

bdunne:
- Add an install step for creating a namespace is the user doesn't have one.
- Add uninstall steps


manageiq-operator/README.md
```diff
@@ -215,3 +226,45 @@ spec:
...
+
+### Step 3. Remove the RBAC that was added in Step 3
+```bash
+$ oc delete -f deploy/role.yaml
+$ oc delete -f deploy/role_binding.yaml
+$ oc delete -f deploy/service_account.yaml
+```
+
+### Step 4. Remove the Namespace that was created in Step 2 (optional)
+
```
gtanzillo, line 259: It may be worth it to add a warning here that if an existing, shared namespace was used, don't remove it

gtanzillo (approved):
Looks good, just one suggestion

Fryguy (approved):

PR #687, opened 2021-02-23
We only need the ruby environment, not full container_env

jrafanie:
7d841daecccaacbbe made this work but container_env does expensive URL
escaping of the DATABASE_USER and PASSWORD to build DATABASE_URL which
is not needed in the liveness check. Since all pods run this liveness check
many times per minutes, it has to be as fast as possible.

This drops a few 100ms from the liveness check depending on the environment.

### DIFF
--- a/images/manageiq-base-worker/container-assets/manageiq_liveness_check
+++ b/images/manageiq-base-worker/container-assets/manageiq_liveness_check
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-source ${APP_ROOT}/container_env
+[[ -s /etc/default/evm ]] && source /etc/default/evm
 ruby ${APP_ROOT}/lib/workers/bin/heartbeat_check.rb --heartbeat-file=${APP_ROOT}/tmp/worker.hb


chessbyte:
Is there any way to convert the Ruby liveness check code to something faster in another language like Go or sh?

jrafanie:
> Is there any way to convert the Ruby liveness check code to something faster in another language like Go or sh?

yes, for sure... it's just doing mtime on files, reading a possible time string, parsing it and comparing it to the current time and the timeout and exiting if it the current time exceeds the timeout time.  It's leveraging shared code for the hb file location and various timeouts.

NickLaMuro:
@jrafanie @chessbyte I made a quick POC PR to address the "just use `Go`" comment, since this script really doesn't need a new language introduced to function "fast".

https://github.com/ManageIQ/manageiq/pull/21077

Fryguy:
I think sh is the easiest, as in the end you are basically just comparing 2 times, so it's a glorified `[ $(date +%s) -ge $(cat hbfile) ]`, which is probably blazing fast.  Might need some defensive params for the file existence, possibly.

I spoke with @jrafanie offline, and he was gonna experiment with that.

Fryguy:
@jrafanie I'm gonna merge this as-is...it's a quick win for an immediate 200-300 ms, which is like 15%-25% faster already.  We can do follow up PRs to convert to base or remove AS or whatever's best.

jrafanie:
Thanks for the motivation all, while it's shell, this isn't too bad but it's so much faster... 7-26 milliseconds in bash (vs 500-600ms at best in ruby since the latter still would need to source /etc/default/evm or we'd have to figure out what parts of that it needs)...

Either way, this script assumes we don't have to parse the contents of the heartbeat file by assuming it's in unix time... therefore I'll need to change the heartbeat writing code to write unix time.  Additionally, this script can be greatly simplified if we always write the unix time in the file, as we can avoid doing mtime work and don't need a hardcoded 2 minute heartbeat time value.

but anyway, here's what I have so far... I'll open a pr on manageiq to always write unix time and one here to something like this without the extra conditionals.

```bash
#!/bin/bash

# hb_file needs to either exist or only contain a time in unix time
# TODO: Simplify: only write unix time in file (don't rely on mtime)
hb_file="${APP_ROOT}/tmp/worker.hb"
if [[ -f ${hb_file} ]]; then
  current_time=$(date +%s)
  custom_timeout=$(cat ${hb_file})
  if [[ -z $custom_timeout ]]; then
    # TODO: This goes away if we always write unix time in hb file
    # mtime=$(stat -f "%m" ${hb_file}) # for testing on mac
    mtime=$(stat -c "%Y" ${hb_file})   # for linux

    # TODO: this hardcoded timeout goes away if we always write unix time in hb file
    timeout=$((${mtime} + 120))
  else
    timeout=$custom_timeout
  fi

  if [ $current_time -le $timeout ]; then
    exit 0
  else
    exit 1
  fi
else
  exit 1
fi

```

Fryguy:
Backported to `lasker` in commit 7472958f43bb90ddf0d6b98d1720ff3cf8107ac3.

```text
commit 7472958f43bb90ddf0d6b98d1720ff3cf8107ac3
Date:   Tue Feb 23 17:14:41 2021 -0500

    Merge pull request #687 from jrafanie/faster_liveness_check
    
    We only need the ruby environment, not full container_env
    
    (cherry picked from commit 4cb84d3caff6a19aafd7fb9d0e8c38dac5b19ff6)
```

PR #688, opened 2021-02-24
Fasterer liveness check

jrafanie:
Replace [heartbeat_check.rb and friends](https://github.com/ManageIQ/manageiq/blob/c43b2093199bb6942318c3102371b79c03f95473/lib/workers/bin/heartbeat_check.rb) with a simple shell script that assumes the heartbeat file now contains a timeout in unix time seconds since epoch.

Depends on the core change to write these files in unix time. 

- [x] Must be merged with this [core change](https://github.com/ManageIQ/manageiq/pull/21079)

Preliminary results:

### BEFORE
kasparov-1 environment running mostly idle

```
$ oc adm top pods
NAME                                 CPU(cores)   MEMORY(bytes)
1-event-handler-5dd956d75f-fr25h     74m          227Mi
1-generic-5865997d56-2xzrt           54m          322Mi
1-generic-5865997d56-fzwvz           72m          348Mi
1-priority-576657869b-9r4qf          71m          317Mi
1-priority-576657869b-swgdp          61m          297Mi
1-remote-console-5dff596449-ll2j5    35m          199Mi
1-reporting-755b878d7b-jpbzd         59m          189Mi
1-reporting-755b878d7b-vg7c6         59m          229Mi
1-schedule-dd5945764-vvfpb           78m          278Mi
1-ui-7654578488-t9s85                49m          361Mi
1-web-service-cdfc455b7-gbq5x        63m          336Mi
httpd-b89bbdff6-js9qv                1m           20Mi
manageiq-operator-769b4b48bb-dljm9   0m           30Mi
memcached-84987cbdf5-g42zs           0m           8Mi
orchestrator-fc86cbf66-vjzrj         21m          339Mi
postgresql-867fcb7d8-b65xn           9m           274Mi
```

### AFTER
Master, mostly idle.  Compare the CPU column with the same column above:

```
$ oc adm top pods
NAME                                 CPU(cores)   MEMORY(bytes)
1-event-handler-6fb6ff5995-drzw4     3m           177Mi
1-generic-7c87dc84f8-6xgxk           4m           228Mi
1-generic-7c87dc84f8-skp5p           3m           212Mi
1-priority-5df7c8b7f4-hl8jk          12m          228Mi
1-priority-5df7c8b7f4-rgfcl          12m          214Mi
1-remote-console-6f478b75b6-gxh7m    1m           186Mi
1-reporting-79799898-ct72b           5m           177Mi
1-reporting-79799898-knrmb           9m           177Mi
1-schedule-5d569cb977-qtrbr          2m           215Mi
1-ui-74f785fb99-w7q5j                2m           276Mi
1-web-service-5d47b5b5c8-tvjfv       8m           263Mi
httpd-b89bbdff6-tpp5r                0m           14Mi
manageiq-operator-57b78b6c6b-5jrjx   0m           23Mi
memcached-84987cbdf5-qxpjb           0m           1Mi
orchestrator-7f96f4964d-c8vng        16m          345Mi
postgresql-867fcb7d8-sz247           9m           196Mi
```

### DIFF
--- a/images/manageiq-base-worker/container-assets/manageiq_liveness_check
+++ b/images/manageiq-base-worker/container-assets/manageiq_liveness_check
@@ -1,4 +1,27 @@
 #!/bin/bash
 
-[[ -s /etc/default/evm ]] && source /etc/default/evm
-ruby ${APP_ROOT}/lib/workers/bin/heartbeat_check.rb --heartbeat-file=${APP_ROOT}/tmp/worker.hb
+# hb_file needs to either exist or only contain a time in unix time
+# TODO: Simplify: only write unix time in file (don't rely on mtime)
+hb_file="${APP_ROOT}/tmp/worker.hb"
+if [[ -f ${hb_file} ]]; then
+  current_time=$(date +%s)
+  custom_timeout=$(cat ${hb_file})
+  if [[ -z $custom_timeout ]]; then
+    # TODO: This goes away if we always write unix time in hb file
+    # mtime=$(stat -f "%m" ${hb_file}) # for testing on mac
+    mtime=$(stat -c "%Y" ${hb_file})   # for linux
+
+    # TODO: this hardcoded timeout goes away if we always write unix time in hb file
+    timeout=$((${mtime} + 120))
+  else
+    timeout=$custom_timeout
+  fi
+
+  if [ $current_time -le $timeout ]; then
+    exit 0
+  else
+    exit 1
+  fi
+else
+  exit 1
+fi
--- a/images/manageiq-base-worker/container-assets/manageiq_liveness_check
+++ b/images/manageiq-base-worker/container-assets/manageiq_liveness_check
@@ -1,27 +1,15 @@
 #!/bin/bash
 
-# hb_file needs to either exist or only contain a time in unix time
-# TODO: Simplify: only write unix time in file (don't rely on mtime)
-hb_file="${APP_ROOT}/tmp/worker.hb"
-if [[ -f ${hb_file} ]]; then
-  current_time=$(date +%s)
-  custom_timeout=$(cat ${hb_file})
-  if [[ -z $custom_timeout ]]; then
-    # TODO: This goes away if we always write unix time in hb file
-    # mtime=$(stat -f "%m" ${hb_file}) # for testing on mac
-    mtime=$(stat -c "%Y" ${hb_file})   # for linux
+# We're alive if the heartbeat_file exists and has a unix time
+# "newer" / greater than the current time.
+heartbeat_file="${APP_ROOT}/tmp/worker.hb"
+if [[ ! -f ${heartbeat_file} ]]; then
+  exit 1
+fi
 
-    # TODO: this hardcoded timeout goes away if we always write unix time in hb file
-    timeout=$((${mtime} + 120))
-  else
-    timeout=$custom_timeout
-  fi
+current_time=$(date +%s)
+not_responsive_threshold=$(cat ${heartbeat_file})
 
-  if [ $current_time -le $timeout ]; then
-    exit 0
-  else
-    exit 1
-  fi
-else
+if [ $current_time -gt $not_responsive_threshold ]; then
   exit 1
 fi


images/manageiq-base-worker/container-assets/manageiq_liveness_check
```diff
@@ -1,4 +1,17 @@
 #!/bin/bash
 
-[[ -s /etc/default/evm ]] && source /etc/default/evm
-ruby ${APP_ROOT}/lib/workers/bin/heartbeat_check.rb --heartbeat-file=${APP_ROOT}/tmp/worker.hb
+# We're alive if the hb_file exists and has a unix time "newer" / greater
+# than the current time.
+hb_file="${APP_ROOT}/tmp/worker.hb"
+if [[ -f ${hb_file} ]]; then
+  current_time=$(date +%s)
+  custom_timeout=$(cat ${hb_file})
```
chessbyte, line 8: why call it `custom_timeout` instead of `heartbeat_time`?

images/manageiq-base-worker/container-assets/manageiq_liveness_check
```diff
@@ -1,4 +1,17 @@
 #!/bin/bash
 
-[[ -s /etc/default/evm ]] && source /etc/default/evm
-ruby ${APP_ROOT}/lib/workers/bin/heartbeat_check.rb --heartbeat-file=${APP_ROOT}/tmp/worker.hb
+# We're alive if the hb_file exists and has a unix time "newer" / greater
+# than the current time.
+hb_file="${APP_ROOT}/tmp/worker.hb"
```
chessbyte, line 5: suggest `heartbeat_file` instead of `hb_file`

images/manageiq-base-worker/container-assets/manageiq_liveness_check
Fryguy: it's sort of neither.  The value inside is the custom threshold (specifically, `Time.now.to_i + heartbeat_threshold`)

Fryguy: Maybe just call it `threshold`?

images/manageiq-base-worker/container-assets/manageiq_liveness_check
```diff
@@ -1,4 +1,17 @@
...
+if [[ -f ${hb_file} ]]; then
```
Fryguy, line 6: Minor, but I'm more of a guard clause kind of guy, so I'd rather see

```suggestion
if [[ ! -f ${hb_file} ]]; then
  exit 1
fi
```

images/manageiq-base-worker/container-assets/manageiq_liveness_check
```diff
@@ -1,4 +1,17 @@
...
+  current_time=$(date +%s)
+  custom_timeout=$(cat ${hb_file})
+
+  if [ $current_time -le $custom_timeout ]; then
+    exit 0
+  else
+    exit 1
+  fi
```
Fryguy, line 14: Similarly here

```suggestion
  if [ $current_time -gt $custom_timeout ]; then
    exit 1
  fi
```

The exit 0 will then naturally fall out.

images/manageiq-base-worker/container-assets/manageiq_liveness_check
jrafanie: sure, it was named to match the name of the file.

images/manageiq-base-worker/container-assets/manageiq_liveness_check
jrafanie: Good question.  It's because we store the [current time + the worker's timeout or the default timeout of 2 minutes](https://github.com/ManageIQ/manageiq/blob/de4c23643816b171243b26c5d48dc03add943873/app/models/miq_worker/runner.rb#L316).  We're storing the "custom timeout", or not the actual last heartbeat.  It's also not the [heartbeat timeout](https://github.com/ManageIQ/manageiq/blob/de4c23643816b171243b26c5d48dc03add943873/lib/workers/miq_defaults.rb#L5) as that's a constant 2 minutes.  We have a custom value that could be the current time + that 2 minutes or a timeout from that worker's settings. 

Of note, we can't even see the worker heartbeat files from the "server" process in podified unless they're on the same filesystem, which isn't the case in pods.  The server assumes the worker's are alive or dead based on the liveness check.  If the worker row goes away, because it was killed due to not heartbeating, we won't update the worker's last heartbeat.  If it does exist, we assume it's heartbeating and set the last heartbeat to the current time.  See [here](https://github.com/ManageIQ/manageiq/pull/21079/files#diff-7872eccf60832f6d08361bdb1a0cc4ce6ea7482b5c48b17b6b145af33c78e2b0R7)

images/manageiq-base-worker/container-assets/manageiq_liveness_check
jrafanie: haha, I wanted to demonstrate the "normal" case exiting properly instead of a side effect of not exiting early.

jrafanie: I can change it... I'm currently doing a build to make sure the two changes work together, so I can make any suggestions a bit later.

images/manageiq-base-worker/container-assets/manageiq_liveness_check
jrafanie: > Maybe just call it `threshold`?

I'm not really a fan of threshold since that's also used for memory_threshold and it's not really conveying a meaning.  I think it's very much a custom timeout or maybe just a timeout.  We overloaded terms with heartbeat timeout so we can't really use that.  I'm open to other ideas.

chessbyte: I think the part that is confusing is comparing `current_time` to a `timeout`.  The former feels like a particular time and the latter like a length of time (i.e. 5 minutes)

jrafanie: We have agreed that `not_responsive_threshold` sounds better for the purposes of what the liveness check is doing.

images/manageiq-base-worker/container-assets/manageiq_liveness_check
jrafanie: haha, I just noticed the comment at the top will now match the code you're suggesting...

Nice work

```
# We're alive if the heartbeat_file exists and has a unix time "newer" / greater
# than the current time.
```

Fryguy (approved):

images/manageiq-base-worker/container-assets/manageiq_liveness_check
jrafanie: Ok, changed.

chessbyte:
> Preliminary results

@jrafanie very impressive performance improvements

Fryguy:
@jrafanie Do you have a PR that rm's all of the original heartbeat check code or is that going to a follow up?

Fryguy:
@jrafanie Super minor, but can we rename to not have manageiq in the name?

jrafanie:
> @jrafanie Do you have a PR that rm's all of the original heartbeat check code or is that going to a follow up?

Yes, will do a followup.

jrafanie:
> > Preliminary results
> 
> @jrafanie very impressive performance improvements

Thanks for the suggestion...this is a clear case where this needs to be fast and should be hyper optimized.

bdunne (approved):
👍🏻 LGTM

Fryguy:
Backported to `lasker` in commit 1c08ae764e91a86549f1bdaf5cfe8486eb88f020.

```text
commit 1c08ae764e91a86549f1bdaf5cfe8486eb88f020
Date:   Wed Feb 24 17:28:37 2021 -0500

    Merge pull request #688 from jrafanie/fasterer_liveness_check
    
    Fasterer liveness check
    
    (cherry picked from commit 5609c81c53d8a4264b37f6d2d38f5e7e0de2d436)
```

PR #702, opened 2021-03-23
Clear caches better for smaller images

Fryguy:
These changes cause the base-worker image to drop from 2.2GB to 1.99GB
(~200MB / 9%) and the ui-worker image to drop from 2.51GB to 2.18GB
(~330MB / 13%).

Notable savings are:
- When running the generate_rpm_manifest, the dnf commands fill the
  cache, which wasn't being cleaned. This particulary affects the
  manifest run immediately after installing manageiq-ui dropping that
  layer size from 123MB to 6.6MB.
- When installing the pythin dependencies, we run dnf install, which
  wasn't being cleaned. This layer dropped from 604MB to 435MB
- The chgrp commands are done in a separate layer from the layer that
  installs those files, causing an extra 39MB layer which is
  unnecessary.

### DIFF
--- a/images/manageiq-base-worker/Dockerfile
+++ b/images/manageiq-base-worker/Dockerfile
@@ -21,6 +21,9 @@ COPY --from=vddk /vddk/vmware-vix-disklib-distrib/ /usr/lib/vmware-vix-disklib/
 COPY container-assets/install-vmware-vddk /tmp/
 RUN /tmp/install-vmware-vddk
 
-RUN source /etc/default/evm && /usr/bin/generate_rpm_manifest.sh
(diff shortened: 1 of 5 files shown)


Fryguy:
Backported to `lasker` in commit 106c197d53547a8372208deda0dbc65652250e98.

```text
commit 106c197d53547a8372208deda0dbc65652250e98
Date:   Tue Mar 23 11:47:47 2021 -0400

    Merge pull request #702 from Fryguy/smaller_images
    
    Clear caches better for smaller images
    
    (cherry picked from commit d426e10bcf4a9ec00cba8a58f8f382314027c8ce)
```

PR #706, opened 2021-04-01
[WIP] Replace Controller references for any objects that we are supposed to own

bdunne:
Super WIP right now.  I have to figure out how to test this.


bdunne:
Decided to take a different route to fix this.

PR #723, opened 2021-06-04
Allow running with TLS/SSL

bdunne:
- Ensure that local CAs are trusted
- Install mod_ssl

### DIFF
--- a/images/manageiq-base/container-assets/container_env
+++ b/images/manageiq-base/container-assets/container_env
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+update-ca-trust
+
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 
 function urlescape() {
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -30,6 +30,9 @@ LABEL name="manageiq-base" \
       io.k8s.description="ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures." \
       io.openshift.tags="ManageIQ,miq,manageiq"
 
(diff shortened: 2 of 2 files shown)


images/manageiq-ui-worker/Dockerfile
```diff
@@ -18,6 +18,7 @@ COPY container-assets/manageiq-http.conf /etc/httpd/conf.d
 
 RUN source /etc/default/evm && \
     /usr/bin/generate_rpm_manifest.sh && \
+    dnf install -y mod_ssl && \
```
Fryguy, line 21: Can we put this in a separate section on its own (or even in the base image)?  Reason I'm asking is because this section is copy-pasted across the images, so for consistent updates I'd like to keep them all the same.

Fryguy: Additionally, we should install mod_ssl _before_ we generate the manifest that will need to include it.

images/manageiq-base/container-assets/container_env
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+update-ca-trust
```
Fryguy, line 3: What provides this command?  If it's mod_ssl, and the change to add mod_ssl is only in the ui worker image, then the other images will fail.

bdunne: It's part of `ca-certificates-2020.2.41-80.0.el8_2.noarch` which comes with UBI

images/manageiq-ui-worker/Dockerfile
bdunne: Moved to https://github.com/ManageIQ/manageiq-rpm_build/pull/163 where we install httpd

PR #728, opened 2021-06-16
Expand LimitRequestFieldSize in the internal httpd

Fryguy:
LimitRequestFieldSize: Expand this to a large number to allow pass-through.
This does not introduce a potential DoS, because the value is validated by
the httpd container first.  However, if a user changes this value in the
httpd container, we need to be able to accomodate that value here also.

512K was chosen as the size because that is also the maximum limit of
the OIDCCacheShmEntrySizeMax value for OIDC.

@bdunne Please review.

### DIFF
--- a/images/manageiq-ui-worker/container-assets/manageiq-http.conf
+++ b/images/manageiq-ui-worker/container-assets/manageiq-http.conf
@@ -10,6 +10,12 @@ ServerTokens Prod
 RewriteEngine On
 Options SymLinksIfOwnerMatch
 
+# LimitRequestFieldSize: Expand this to a large number to allow pass-through.
+#   This does not introduce a potential DoS, because the value is validated by
+#   the httpd container first.  However, if a user changes this value in the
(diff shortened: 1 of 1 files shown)


ag-ibm:
Hi Jason, thanks a million for addressing this, is there anyway we can try the fix on the client's cp4mcm 2.2 cluster? Also we are currently installing the new cluster using 2.3. Please let me know how to add the fix into the installation.

Fryguy:
Backported to `lasker` in commit 2c4882b0c368a75aae277ab7c084b1b79086f81a.

```text
commit 2c4882b0c368a75aae277ab7c084b1b79086f81a
Date:   Wed Jun 16 12:14:19 2021 -0400

    Merge pull request #728 from Fryguy/expand_limit_request_field_size
    
    Expand LimitRequestFieldSize in the internal httpd
    
    (cherry picked from commit 7683b9239069c3b3572e4495e0a09f8da58e2659)
```

PR #739, opened 2021-08-18
SSL Part 1

bdunne:
- Introduce the secret to contain all internal certificates
- Secure the connection between workers and postgresql

Depends on:
- https://github.com/ManageIQ/container-postgresql/pull/27


kbrock:
I thought we were pointing the ssl certificate into the `/var/www/vmdb/certs` area for a while.
Not that I like that solution, but if we want to use the standard location for the postgresql certs, then lets keep that in mind when we notice that documentation for the appliance differs from the pod

bdunne:
I don't think pods has ever used `/var/www/vmdb/certs` for trusted CAs or postgres connections.

chessbyte:
> I don't think pods has ever used /var/www/vmdb/certs for trusted CAs or postgres connections.

@bdunne but the question is `should we?` for consistency between appliance and podified?

bdunne:
> > I don't think pods has ever used /var/www/vmdb/certs for trusted CAs or postgres connections.
> 
> @bdunne but the question is `should we?` for consistency between appliance and podified?

For CA or server public certificates that we use to verify connections to an endpoint (what we're doing here), I think they belong in `/etc/pki/ca-trust/source/anchors` so that they are globally accessible.

If we were to run rails with SSL, I would expect the private keys to be in `/var/www/vmdb/certs`

If the rails app is authenticating to services using certificates, I would expect those certificates to also live in `/var/www/vmdb/certs`

manageiq-operator/pkg/helpers/miq-components/util.go
```diff
@@ -85,3 +91,68 @@ func addAnnotations(annotations map[string]string, meta *metav1.ObjectMeta) {
...
+	secret := &corev1.Secret{}
+	client.Get(context.TODO(), secretKey, secret)
+
+	return secret
+}
+
+func addInternalCertificate(cr *miqv1alpha1.ManageIQ, d *appsv1.Deployment, client client.Client, name string, mountPoint string) {
+	secret := InternalCertificatesSecret(cr, client)
+	if secret.Data[fmt.Sprintf("%s_crt", name)] != nil && secret.Data[fmt.Sprintf("%s_key", name)] != nil {
+		volumeName := fmt.Sprintf("internal-%s-certificate", name)
```
Fryguy, line 108: what if the postgres is external - does this name then make it confusing?  is it better to just say "%s-certificate"?

bdunne: Done, btw... it's only used to link the volumeMount to the volumeSource, but I see the point.

bdunne:
@Fryguy Any remaining concerns?

Issue #724, opened 2021-06-04
[RFE] Allow SSL/TLS communication between pods


bdunne:
- [x] Route -> HTTPD https://github.com/ManageIQ/manageiq-pods/pull/745
- [x] HTTPD -> UI
  - [x] httpd Container https://github.com/ManageIQ/container-httpd/pull/61
  - [x] UI Container https://github.com/ManageIQ/manageiq-pods/pull/723
  - [x] Operator / orchestrator UI pod httpd config
    - [x] Orchestrator: https://github.com/ManageIQ/manageiq/pull/21527
    - [x] Operator: https://github.com/ManageIQ/manageiq-pods/pull/768
- [x] HTTPD -> API
  - [x]  Enable httpd on on the API pod https://github.com/ManageIQ/manageiq-pods/pull/742 & https://github.com/ManageIQ/manageiq/pull/21395
- [x] ALL -> Postgres (internal)
  - [x]  Operator https://github.com/ManageIQ/manageiq-pods/pull/739
  - [x] container-postgresql https://github.com/ManageIQ/container-postgresql/pull/27
- [x] ALL -> Memcached
  - [x] https://github.com/ManageIQ/manageiq/pull/21280 ManageIQ app changes for Dalli v3 and enabling SSL
  - [x] https://github.com/ManageIQ/manageiq-pods/pull/767 Turn on SSL connections if the required certificates exist
- [x] Documentation (WIP) https://github.com/ManageIQ/manageiq-documentation/pull/1574
  - [x] Certificate generation tool https://github.com/ManageIQ/manageiq-pods/pull/772
- [ ] ~~ALL -> Kafka~~ Followup in https://github.com/ManageIQ/manageiq-pods/issues/789

bdunne:
Closing, implemented for all default services.  Followup for kafka (optional) in https://github.com/ManageIQ/manageiq-pods/issues/789

PR #741, opened 2021-08-20
Move UI httpd config to a config map and include optional configs

bdunne:
This allows people to customize the httpd config in the UI worker

PR #742, opened 2021-08-24
Add httpd to the API pod and create a configMap for the config

bdunne:
Part of https://github.com/ManageIQ/manageiq-pods/issues/724


Fryguy:
Is the content the same as https://github.com/ManageIQ/manageiq-pods/pull/741/files#diff-a4b5e916d0a1ed6717db85cad449a7c0b23d4d42edc2a116c343934b93c3ebe6R384 ? If so, can we reuse that method?

manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
```diff
@@ -432,3 +433,57 @@ LimitRequestFieldSize 524288
...
+  ServerName %s://web-service
+  DocumentRoot /var/www/miq/vmdb/public
+
+  RewriteCond %{REQUEST_URI}     ^/ws/notifications [NC]
+  RewriteCond %{HTTP:UPGRADE}    ^websocket$ [NC]
+  RewriteCond %{HTTP:CONNECTION} ^Upgrade$   [NC]
+  RewriteRule .* ws://localhost:3001%{REQUEST_URI}  [P,QSA,L]
+  ProxyPassReverse /ws/notifications ws://localhost:3001/ws/notifications
+
+  RewriteRule ^/web-service/service(?!/(assets|images|img|styles|js|fonts|vendor|gettext)) /web-service/service/index.html [L]
```
Fryguy, line 469: I think this is the only line that's different?  But this isn't valid because /web-service/service isn't a path.  The other one is for the SUI.

manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
```diff
@@ -432,3 +433,57 @@ LimitRequestFieldSize 524288
...
+# LimitRequestFieldSize: Expand this to a large number to allow pass-through.
+#   This does not introduce a potential DoS, because the value is validated by
+#   the httpd container first.  However, if a user changes this value in the
+#   httpd container, we need to be able to accomodate that value here also.
+LimitRequestFieldSize 524288
+
+<VirtualHost *:3000>
+  IncludeOptional conf.d/*_config
+
+  ServerName %s://web-service
```
Fryguy, line 460: Oh this line is different too.  Wonder if it's better to have a method that takes the server name and builds this string, if this is the only change.

Fryguy:
Partially backported to `morphy` via https://github.com/ManageIQ/manageiq-pods/pull/752

PR #745, opened 2021-09-03
[WIP] Secure inbound connections to httpd

bdunne:
Part of: https://github.com/ManageIQ/manageiq-pods/issues/724

- Switch from Ingress to Route to handle reencryption
- Run httpd with the provided certificates
- Reuse addOrUpdateEnvVar

WIP because I need to re-test after the rebase


manageiq-operator/cmd/manager/main.go
```diff
@@ -29,6 +29,7 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/manager/signals"
+	routev1 "github.com/openshift/api/route/v1"
```
Fryguy, line 32: Will this prevent us from running on plain Kubernetes?  If so, I don't think we can use this.

manageiq-operator/cmd/manager/main.go
```diff
@@ -113,6 +114,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	//Add route scheme
```
Fryguy, line 117:
```suggestion
	// Add route scheme
```

manageiq-operator/pkg/helpers/miq-components/httpd.go
```diff
@@ -42,6 +43,62 @@ func HttpdServiceAccount(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*cor
...
+
+		var public = tlsSecret(cr, client)
+		route.Spec.TLS.Certificate = string(public.Data["tls.crt"])
+		route.Spec.TLS.Key = string(public.Data["tls.key"])
+		// route.Spec.TLS.CACertificate
+
+		internalCerts := InternalCertificatesSecret(cr, client)
+		route.Spec.TLS.DestinationCACertificate = string(internalCerts.Data["root_crt"])
+
+
```
Fryguy, line 85:
```suggestion

```

Fryguy: Oh sorry - didn't realize this was WIP 😅

manageiq-operator/cmd/manager/main.go
bdunne: Yes, this part is OpenShift only.  It's conditional, so we only try to create a route if we have a certificate for the ingress.  The default ingress can not handle re-encryption.

bdunne: Loading the library shouldn't make us OpenShift only though

manageiq-operator/cmd/manager/main.go
```diff
@@ -113,6 +114,12 @@ func main() {
...
+	if err := routev1.AddToScheme(mgr.GetScheme()); err != nil {
+		log.Error(err, "")
+		os.Exit(1)
```
Fryguy, line 120: I think it was this part that made me concerned about locking us to openshift

PR #754, opened 2021-09-23
Add basic framework for Go tests

Fryguy:
@bdunne Please review.

### DIFF
--- a/lib/tasks/spec.rake
+++ b/lib/tasks/spec.rake
@@ -2,4 +2,10 @@ require 'rspec/core/rake_task'
 
 RSpec::Core::RakeTask.new(:spec)
 
-task :default => :spec
+desc "Run Go tests"
+task :go_test do
+  sh "go test -v ./...", :chdir => File.expand_path("../../manageiq-operator", __dir__)
+  puts
+end
+
+task :default => [:spec, :go_test]
--- a/manageiq-operator/pkg/helpers/miq-components/generators_test.go
+++ b/manageiq-operator/pkg/helpers/miq-components/generators_test.go
@@ -1,0 +1,15 @@
+package miqtools
+
+import (
+	"testing"
+)
+
+func TestRandomBytes(t *testing.T) {
+	expected := 10
+	bytes := randomBytes(expected)
+	actual := len(bytes)
+
+	if actual != expected {
+		t.Fatalf(`Expected %d bytes, but got %d bytes.`, expected, actual)
+	}
+}


manageiq-operator/pkg/helpers/miq-components/generators_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"testing"
+)
+
+func TestRandomBytes(t *testing.T) {
+	expected := 10
+	bytes := randomBytes(expected)
+	actual := len(bytes)
+
+	if actual != expected {
+		t.Fatalf(`Expected %d bytes, but got %d bytes.`, expected, actual)
```
NickLaMuro, line 13: Spacing here seems bigger than what you used previously (4 instead of 2).

NickLaMuro: Also, wouldn't this (very infrequently) fail when `randomBytes` produces the same as actual?  Or is that not mathematically possible?

lib/tasks/spec.rake
```diff
@@ -2,4 +2,12 @@ require 'rspec/core/rake_task'
 
 RSpec::Core::RakeTask.new(:spec)
 
-task :default => :spec
+desc "Run Go tests"
+task :go_test do
+  cmd = "go test -v ./..."
+  puts cmd
+  exit $?.exitstatus unless system(cmd, :chdir => File.expand_path("../../manageiq-operator", __dir__))
+  puts
```
NickLaMuro, line 10: This can be simplified to just use the `rake` builtin `sh`:

```ruby
task :go_test do
  sh "go test -v ./..."
end
```

NickLaMuro (changes requested):

lib/tasks/spec.rake
NickLaMuro: Okay, I lied a bit here because of the need to do the `:chdir`... let me look up and see if that is supported with `sh`.  Otherwise you could wrap in a `chdir { }`:

```
task :go_test do
  chdir File.expand_path("../../manageiq-operator", __dir__)) do
    sh "go test -v ./..."
  end
end

NickLaMuro: Yeah, it does support `:chdir => path`, because it just wraps `system`:

https://github.com/ruby/rake/blob/5c60da8644a9e4f655e819252e3b6ca77f42b7af/lib/rake/file_utils.rb#L54

bdunne:
What do you think of using [Ginkgo](https://onsi.github.io/ginkgo/)?  It reads like RSpec (much cleaner)

manageiq-operator/pkg/helpers/miq-components/generators_test.go
Fryguy: Go uses tabs, and GitHub doesn't render the first tab properly.

Notice that actual is the length of the array, which should always be 10

lib/tasks/spec.rake
Fryguy: Does it exit with that same exit code on failure?

Fryguy:
@bdunne yeah I can give that a try

lib/tasks/spec.rake
NickLaMuro: > Does it exit with that same exit code on failure?

Does that matter?  I don't see why it would, but no, it doesn't:

https://github.com/ruby/rake/blob/5c60da8644a9e4f655e819252e3b6ca77f42b7af/lib/rake/file_utils.rb#L66-L68

Simply calls `fail` and prints the exit status of the command.

manageiq-operator/pkg/helpers/miq-components/generators_test.go
NickLaMuro: Okay, yeah, viewed it without the "diff view", and yeah it looks fine.  That is annoying, but good to know.

And I see now I was just not understanding the test.  Whoops.

lib/tasks/spec.rake
Fryguy: That's probably fine - my usual pattern for wrapping shell commands tries to exit with the same exit code.  I just couldn't follow the rake code to understand what it was doing on failure.

Fryguy: Done.

Fryguy:
@bdunne Thoughts on merging this first, and then I can follow up with ginkgo?

Fryguy:
Closing in favor of the tests that will be added in the operator-sdk upgrade #831

PR #770, opened 2021-11-01
[WIP] Move certificates from /root to /etc/pki/tls/

bdunne:
Requires: https://github.com/ManageIQ/container-postgresql/pull/32

Based on discussion in: https://github.com/ManageIQ/manageiq/pull/21527#discussion_r740367755


kbrock:
thanks for putting this together.
Is there anything I can do to help with this?

kbrock:
@bdunne darn. Could this be redirected?

I guess my main desire for this was to have the credentials available for non-root users on the appliance. So this would be more of a consistency thing.

You thinking punt is better?

kbrock:
Is this still viable?

agrare:
:+1: this is a great idea, wonder if we could do the same on appliances (move to a more standard location)

PR #772, opened 2021-11-10
Add a tool for generating certificates to enable SSL/TLS between pods

bdunne:
Based on discussion in https://github.com/ManageIQ/manageiq-documentation/pull/1574#discussion_r746073542

### DIFF
--- a/tools/cert_generator.rb
+++ b/tools/cert_generator.rb
@@ -1,0 +1,77 @@
+#!/usr/bin/env ruby
+
+require 'fileutils'
+require 'openssl'
+require 'openssl-extensions/all'
+require 'pathname'
+
+class CertGenerator
+  attr_accessor :root_cert, :root_key
+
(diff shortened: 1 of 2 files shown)


tools/cert_generator.rb
```diff
@@ -0,0 +1,77 @@
...
+end
+
+c = CertGenerator.new
+c.generate_cert("httpd")
+c.generate_cert("kafka")
+c.generate_cert("memcached")
+c.generate_cert("postgresql")
+
+c.generate_cert("api", "your_applicationDomain_hostname")
+c.generate_cert("ui", "your_applicationDomain_hostname")
```
Fryguy, line 77:
```suggestion
c.generate_cert("api", application_domain)
c.generate_cert("ui", application_domain)
```

bdunne: Fixed

PR #812, opened 2022-02-09
Switch to GitHub Actions

bdunne:


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,28 @@
...
+      with:
+        ruby-version: ${{ matrix.ruby-version }}
+        bundler-cache: true
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: '1.14'
+    - name: Run tests
+      run:
+      - travis_wait curl -L https://github.com/operator-framework/operator-sdk/releases/download/v0.18.2/operator-sdk-v0.18.2-$(uname
```
Fryguy, line 26: travis_wait is not here

Fryguy: Also can we move this into a bin/before_install for consistency with other repos?  before_install should be a shell script that checks for CI and installs any system packages and whatnot (Example: https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/bin/before_install).  before_install should be before the `Set up Ruby` step (and so should be a pure shell script (Example: https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/.github/workflows/ci.yaml#L30-L31)

Something like

```sh

bdunne: 👍🏻 Fixed

Fryguy: Can you do the bin/before_install script as well?

chessbyte: @bdunne ^^

bdunne: Done

chessbyte:
@Fryguy seems like this ready for review

PR #682, opened 2021-02-09
Ensure images and operator dirs are relative to the base dir

Fryguy:
Without this PR, the operator dir is not relativel to the base dir, and is assuming that `pwd` is the base dir, so when it tries to pushd to it, it fails

Note that this PR changes the meaning of `-d`, so it is a breaking change and will require changes on the build machine.

PR #814, opened 2022-02-17
Fix `bin/build -d` in README

agrare:
The -d option specifies the base_dir, the images_dir is relative to the
base_dir.

When using the commands in the README I get the following error:
```
bin/build -d images -r manageiq -l
bin/build: line 65: pushd: images/images: No such file or directory
```

What works is
```
bin/build -d . -r manageiq -l
```


Fryguy:
Oh thanks I forgot to update after https://github.com/ManageIQ/manageiq-pods/pull/682

PR #823, opened 2022-03-08
adding secure headers to apache

kbrock:
- these are immutable static assets
- adding packs which are also immutable static assets
- adding Cache-Control public to allow proxy servers to cache these files

This is the pod version of https://github.com/ManageIQ/manageiq-appliance/pull/358

### DIFF
--- a/manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
+++ b/manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
@@ -432,9 +432,27 @@ LimitRequestFieldSize 524288
   ProxyPreserveHost on
   <Location /assets/>
     Header unset ETag
+    Header set Content-Security-Policy            "default-src 'self'; child-src 'self'; connect-src 'self'; font-src 'self' fonts.gstatic.com; script-src 'self'; style-src 'self'; report-uri /dashboard/csp_report"
+    Header set X-Content-Type-Options             "nosniff"
+    Header set X-Frame-Options                    "SAMEORIGIN"
+    Header set X-Permitted-Cross-Domain-Policies  "none"
+    Header set X-XSS-Protection                   "1; mode=block"
     FileETag None
     ExpiresActive On
     ExpiresDefault "access plus 1 year"
+    Header merge Cache-Control public
+  </Location>
+  <Location /packs/>
+    Header unset ETag
+    Header set Content-Security-Policy            "default-src 'self'; child-src 'self'; connect-src 'self'; font-src 'self' fonts.gstatic.com; script-src 'self'; style-src 'self'; report-uri /dashboard/csp_report"
+    Header set X-Content-Type-Options             "nosniff"
+    Header set X-Frame-Options                    "SAMEORIGIN"
+    Header set X-Permitted-Cross-Domain-Policies  "none"
+    Header set X-XSS-Protection                   "1; mode=block"
+    FileETag None
+    ExpiresActive On
+    ExpiresDefault "access plus 1 year"
+    Header merge Cache-Control public
   </Location>
   <Location /proxy_pages/>
     ErrorDocument 403 /error/noindex.html


manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
```diff
@@ -425,9 +425,27 @@ LimitRequestFieldSize 524288
   ProxyPreserveHost on
   <Location /assets/>
     Header unset ETag
+    Header set Content-Security-Policy            "default-src 'self'; child-src 'self'; connect-src 'self'; font-src 'self' fonts.gstatic.com; script-src 'unsafe-eval' 'unsafe-inline' 'self'; style-src 'unsafe-inline' 'self'; report-uri /dashboard/csp_report"
```
Fryguy, line 428: A lot of these don't make sense...for example, I don't think we need `font-src fonts.gstatic.com` on the static assets, but maybe I'm wrong.  Same for things like unsafe-eval and unsafe-inline.

kbrock: I was avoiding picking and choosing. And went the blind approach.

I understand how these make sense from `.js`, but they don't make sense to me on `.jpg`. Unfortunately, I think that is what was requested.

kbrock: ok. I used the header that is set in `secure_headers.rb` for the values that I set into apache. Specifically, this is the C-S-P from `/ops/explorer`.

I had originally thought we could setup one C-S-P and put it into apache, removing it from rails all together.
My second thought was to have one C-S-P and duplicate it across `secure_headers.rb`, apache, and pods. There are a few controllers that tweak the policy as needed (i.e.: consoles and websockets).

Please confirm: You want our dynamic C-S-P set in `secure_headers.rb`
and a different static C-S-P set in apache/pods (synchronized)?

We do have an outstanding request to change the policy we use, but we stated that our login requires our current policy.

kbrock:
update:
- using `self` for all
- tested with a static file with these changes (from guides oidc) and no errors were introduced
- this matches the changes to the apache appliance

I did notice that we are not using a single stylesheet page. Which is wrong. So when we do introduce a stylesheet, I'm guessing we will need the special fonts link. I'm a little annoyed just how many fonts we are shipping with our product and we need to go and fetch more from google. But addressing that complaint is not the goal of this PR.

kbrock:
@Fryguy are you good with this PR

manageiq-operator/pkg/helpers/miq-components/httpd_conf.go
```diff
@@ -425,9 +425,27 @@ LimitRequestFieldSize 524288
...
     Header unset ETag
+    Header set Content-Security-Policy            "default-src 'self'; child-src 'self'; connect-src 'self'; font-src 'self' fonts.gstatic.com; script-src 'self'; style-src 'self'; report-uri /dashboard/csp_report"
+    Header set X-Content-Type-Options             "nosniff"
+    Header set X-Frame-Options                    "SAMEORIGIN"
+    Header set X-Permitted-Cross-Domain-Policies  "none"
+    Header set X-XSS-Protection                   "1; mode=block"
     FileETag None
     ExpiresActive On
     ExpiresDefault "access plus 1 year"
+    Header merge Cache-Control public
```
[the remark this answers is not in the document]
kbrock: pretty sure expires default is the expires header, while the other just merges the cache control header.

But yes, I made that change

Fryguy:
Backported to `najdorf` in commit c1fa94c48f28d0d56ff47c0aec23b39e764407ae.

```text
commit c1fa94c48f28d0d56ff47c0aec23b39e764407ae
Date:   Fri May 20 09:23:38 2022 -0400

    Merge pull request #823 from kbrock/oidc_same_site
    
    adding secure headers to apache
    
    (cherry picked from commit 4bcd909e1baa97424fd938cf4264345ff66d9bd9)
```

Fryguy:
Backported to `morphy` in commit 48f8382889bccfdba75c2ae5ac5478b8fbed8c33.

```text
commit 48f8382889bccfdba75c2ae5ac5478b8fbed8c33
Date:   Fri May 20 09:23:38 2022 -0400

    Merge pull request #823 from kbrock/oidc_same_site
    
    adding secure headers to apache
    
    (cherry picked from commit 4bcd909e1baa97424fd938cf4264345ff66d9bd9)
```

PR #825, opened 2022-03-10
Allow httpGet health probes using the /ping endpoint

bdunne:
Enables: https://github.com/ManageIQ/manageiq/pull/21766


Fryguy:
I'd like to discuss this before merge...something about this feels dangerous, but maybe I'm overly concerned.

PR #826, opened 2022-03-10
Fix escaping

bdunne:
`RewriteRule ^/ping http://localhost:3001%!{(string=http)REQUEST_URI} [P,QSA,L]`


Fryguy:
We do this so often I feel like we need a test for checking these 😂

PR #833, opened 2022-04-08
Pull mod_ssl from UBI, not CentOS Appstream

bdunne:
Copy of https://github.com/ManageIQ/container-httpd/pull/74


bdunne:
Removed `najdorf/yes?` since https://github.com/ManageIQ/manageiq-pods/pull/834 has already been backported and is a better solution.

PR #836, opened 2022-04-29
Handle a missing database in orchestrator entrypoint

agrare:
Handle rake evm:deployment_status returning "no_database" and sleep for a bit then retry

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21837

### DIFF
--- a/images/manageiq-orchestrator/container-assets/entrypoint
+++ b/images/manageiq-orchestrator/container-assets/entrypoint
@@ -16,6 +16,57 @@ function check_svc_status() {
   echo "${SVC_NAME}:${SVC_PORT} - accepting connections"
 }
 
+function check_deployment_status() {
+  echo "== Checking deployment status =="
+
+  cd ${APP_ROOT}
+  while true; do
+    bin/rake evm:deployment_status
+    case $? in
+      3) # new_deployment
+        echo "== Starting New Deployment =="
+
+        pushd ${APP_ROOT}
+          set -e
+          REGION=${DATABASE_REGION} bin/rake db:migrate
+          REGION=${DATABASE_REGION} bin/rake db:seed
+          create_initial_group
+        popd
+
+        return 0
+      ;;
+      4) # new_replica
+        echo "New replica is not supported, exiting..."
+
+        return 1
+      ;;
+      5) # redeployment
+        echo "== Starting Re-deployment =="
+
+        return 0
+      ;;
+      6) # upgrade
+        echo "== Starting Upgrade =="
+        pushd ${APP_ROOT}
+          bin/rake db:migrate
+          [ "$?" -ne "0" ] && echo "ERROR: Failed to migrate database" && return 1
+        popd
+
+        return 0
+      ;;
+      7)
+        echo "Database has not been created yet, retrying..."
+
+        sleep 5
+      ;;
+      *)
+        echo "Could not find a suitable deployment type, exiting..."
+
+        return 1
+    esac
+  done
+}
+
 function update_auth_settings() {
   echo "== Updating auth settings =="
   script=$(cat <<- EOS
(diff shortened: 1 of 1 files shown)


agrare:
WIP as this is untested

images/manageiq-orchestrator/container-assets/entrypoint
```diff
@@ -16,6 +16,54 @@ function check_svc_status() {
...
+        break
+      ;;
+      6) # upgrade
+        echo "== Starting Upgrade =="
+        pushd ${APP_ROOT}
+          bin/rake db:migrate
+          [ "$?" -ne "0" ] && echo "ERROR: Failed to migrate database" && return 1
+        popd
+
+        break
```
Fryguy, line 54: does this break the loop or the case?  breaks are so confusing to me 😆

agrare: Ha yeah this will break the loop, that's `C` where you need the break so it doesn't skip to the next case

agrare: bash switch/case syntax always confuses me, the `;;` and `esac` :scream:

agrare: We could `return 0` from there though since we're doing `return 1` for the errors, might be more obvious

Fryguy: yeah return 0 might be much clearer

images/manageiq-orchestrator/container-assets/entrypoint
```diff
@@ -16,6 +16,54 @@ function check_svc_status() {
...
+      ;;
+      7)
+        echo "Database has not been created yet, retrying..."
+        sleep 5
```
Fryguy, line 58: Only problem here is that nothing is going to recreate the database.  But for now, this is better than straight blowing up.

agrare: Yeah exactly figured at least this would show in the log and not have the pod constantly failing and restarting

images/manageiq-orchestrator/container-assets/entrypoint
```diff
@@ -16,6 +16,57 @@ function check_svc_status() {
...
+          set -e
+          REGION=${DATABASE_REGION} bin/rake db:migrate
+          REGION=${DATABASE_REGION} bin/rake db:seed
+          create_initial_group
+        popd
+
+        return 0
+      ;;
+      4) # new_replica
+        echo "New replica is not supported, exiting.."
```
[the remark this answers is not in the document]
agrare: :laughing: I actually had my addition below as `...` but then saw this and thought consistency I guess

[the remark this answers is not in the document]
agrare: Done

Fryguy:
Code-wise looks good to me

agrare:
:+1: I'll build some images and test this locally on crc, if it works I'll remove from WIP

jrafanie:
LGTM.  I guess the orchestrator will just spin, and report it's missing a database until it's resolved, then, right?  As opposed to crashing and restarting...  Sounds better to me.

agrare:
Okay I deployed a custom built operator based on this PR to CRC and the initial setup still completes successfully
```
postgresql:5432 - accepting connections
== Writing encryption key ==
{"@timestamp":"2022-05-02T15:01:38.600256 ","hostname":"orchestrator-6fd89c9d45-kvrll","pid":40,"tid":"97b8","service":"evm","level":"info","message":"MIQ(Vmdb::Initializer.log_db_connectable) Successfully connected to the database."}
Deployment status is new_deployment
== Starting New Deployment ==
== 20171030131403 CollapsedInitialMigration: migrating ========================
-- table_exists?("schema_migrations")
   -> 0.0007s
etc...
```

Fryguy:
Backported to `najdorf` in commit 8438de992bcc66304cf82d047e21e6881ad20650.

```text
commit 8438de992bcc66304cf82d047e21e6881ad20650
Date:   Mon May 2 13:52:45 2022 -0400

    Merge pull request #836 from agrare/handle_missing_database_orchestrator_entrypoint
    
    Handle a missing database in orchestrator entrypoint
    
    (cherry picked from commit ee806cf910691bba18df4f443d16388f7ccfaeb5)
```

PR #837, opened 2022-05-26
Don't remove subscription-manager on s390x

Minal-Mahor:
Red Hat Enterprise Linux 8 repo will be removed if we remove **subscription-manager** on s390x due to which createrepo_c and other packages were failed to install within manageiq-base image.

### DIFF
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -12,7 +12,7 @@ RUN mkdir build && \
 FROM registry.access.redhat.com/ubi8/ubi
 MAINTAINER ManageIQ https://manageiq.org
 
-ARG ARCH=x86_64
+ARG ARCH=$(uname -m)
 ARG LOCAL_RPM
 ARG RELEASE_BUILD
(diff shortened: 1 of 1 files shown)


images/manageiq-base/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN mkdir build && \
 FROM registry.access.redhat.com/ubi8/ubi
 MAINTAINER ManageIQ https://manageiq.org
 
-ARG ARCH=x86_64
+ARG ARCH=$(uname -m)
```
Fryguy, line 15: This will not work - ARCH should be passed from the caller as an ARG if it's needed

Fryguy: Added a `RUN echo $ARCH` right after this and I get

```
#5 [stage-1  2/12] RUN echo $(uname
#5 sha256:d7be9006ff8aec6f61879ceb483f7d0e3c9cd732e46e8de4ff9c9fe198688a53
#5 0.220 $(uname
#5 DONE 0.2s
```

PR #840, opened 2022-06-07
Add format docker for s390x builds using podman

Fryguy:
This fixes issues where the images created by podman on s390x are not in v2 image format, which causes issues later when we create multi-arch image manifests.

@miq-bot add-label bug

I think we also have to update the operator-sdk build, but I'm really not sure how.


SriramaSharma (changes requested):

SriramaSharma (approved):
LGTM.

jrafanie (approved):

PR #844, opened 2022-07-08
Clean more files after dnf installing

Fryguy:
This change reduces the UI worker image size from 2.063GB to 2.018GB
(44MB, 2.1%)

- Add the removal of dnf logs, dnf history, and rpm __db lock files.
  This eliminates roughly 2-6MB per invocation.
- Move the cleaning of dnf and rpm assets into a script, which makes it
  easier to use across layers and images.
- Remove redundant rebuild of the rpm manifests, since ther are
  unnecessary unless rpm changes are made.
- Collapse the modifications of the httpd package into the same layer as
  when it's installed.

Part of #736

EDIT: I forgot to mention this also has significant time savings as well.  Locally, the rpm manifest generation steps can take  ~100s (1m40s) x 5 images = ~500s (8m20s).  After this PR, there are only 2 calls, so that's 200s (3m20s) for a savings of ~5min per build

### DIFF
--- a/images/manageiq-base-worker/Dockerfile
+++ b/images/manageiq-base-worker/Dockerfile
@@ -7,6 +7,7 @@ COPY container-assets/ /vddk/
 
 RUN /vddk/extract-vmware-vddk
 
+################################################################################
 
 FROM ${FROM_REPO}/manageiq-base:${FROM_TAG}
 MAINTAINER ManageIQ https://manageiq.org
@@ -20,10 +21,5 @@ COPY container-assets/manageiq_liveness_check /usr/local/bin
 COPY --from=vddk /vddk/vmware-vix-disklib-distrib/ /usr/lib/vmware-vix-disklib/
 COPY container-assets/install-vmware-vddk /tmp/
 RUN /tmp/install-vmware-vddk
-
-RUN source /etc/default/evm && \
-    /usr/bin/generate_rpm_manifest.sh && \
-    dnf clean all && \
-    rm -rf /var/cache/dnf
 
 CMD ["entrypoint"]
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -9,6 +9,8 @@ ARG GIT_AUTH
 RUN mkdir build && \
     if [[ -n "$GIT_AUTH" ]]; then GIT_HOST=${GIT_AUTH}@${GIT_HOST}; fi && curl -L https://${GIT_HOST}/${BUILD_ORG}/${CORE_REPO_NAME}-appliance-build/tarball/${BUILD_REF} | tar vxz -C build --strip 1
 
+################################################################################
+
 FROM registry.access.redhat.com/ubi8/ubi
 MAINTAINER ManageIQ https://manageiq.org
 
@@ -33,8 +35,13 @@ LABEL name="manageiq-base" \
 RUN chmod -R g+w /etc/pki/ca-trust && \
     chmod -R g+w /usr/share/pki/ca-trust-legacy
 
+# Install dumb-init to be used as the entrypoint
+RUN curl -L -o /usr/bin/dumb-init https://github.com/Yelp/dumb-init/releases/download/v1.2.2/dumb-init_1.2.2_${ARCH} && \
+    chmod +x /usr/bin/dumb-init
+
 COPY rpms/* /tmp/rpms/
 COPY container-assets/create_local_yum_repo.sh /
+COPY container-assets/clean_dnf_rpm /usr/local/bin/
 
 RUN curl -L https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo > /etc/yum.repos.d/ansible-runner.repo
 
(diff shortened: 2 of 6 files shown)


images/manageiq-base-worker/Dockerfile
```diff
@@ -7,6 +7,7 @@ COPY container-assets/ /vddk/
 
 RUN /vddk/extract-vmware-vddk
 
+################################################################################
```
Fryguy, line 10: This makes it easier to see various multi-stage builds - I can remove from this PR if you want, but I figured it was fine.

images/manageiq-base/Dockerfile
```diff
@@ -35,6 +37,7 @@ RUN chmod -R g+w /etc/pki/ca-trust && \
 
 COPY rpms/* /tmp/rpms/
 COPY container-assets/create_local_yum_repo.sh /
+COPY container-assets/clean_dnf_rpm /usr/local/bin/
```
Fryguy, line 40: I'm open to renaming this script.

images/manageiq-ui-worker/Dockerfile
```diff
@@ -6,10 +6,3 @@ MAINTAINER ManageIQ https://manageiq.org
 
 LABEL name="manageiq-ui-worker" \
       summary="ManageIQ user interface worker image"
-
-RUN source /etc/default/evm && \
-    /usr/bin/generate_rpm_manifest.sh && \
-    dnf clean all && \
-    rm -rf /var/cache/dnf
-
-RUN rm -rf /tmp/rpms /create_local_yum_repo.sh /etc/yum.repos.d/local_rpm.repo
```
Fryguy, line 15: I wasn't sure why this was done here.  Removing these doesn't change anything with respect to the file size, and one can get at them from one of the parent images.

images/manageiq-base/Dockerfile
```diff
@@ -69,17 +70,16 @@ COPY --from=appliance_build build/kickstarts/partials/post/python_modules.ks.erb
 RUN bash /tmp/python_modules && \
     rm -f /tmp/python_modules && \
     rm -rf /root/.cache/pip && \
-    dnf clean all && \
-    rm -rf /var/cache/dnf
+    clean_dnf_rpm
+
+# Build the RPM manifest
```
bdunne, line 75: I thought we were keeping this as close as possible to the end in case any other steps were added with `dnf` commands

Fryguy: I was actually considering moving the dumb-init and ADD calls _up_ the file.  If you're ok with that for this PR I can do that, and then this is at the "end" of the file again.

Fryguy: Even so, the only calls afterwards are an ADD, and the curl of dumb-init, so this is technically at the "end" of any dnf commands

images/manageiq-webserver-worker/Dockerfile
```diff
@@ -9,23 +9,24 @@ ARG RPM_PREFIX=manageiq
...
+    # Remove httpd default settings
+    rm -f /etc/httpd/conf.d/* && \
+    sed -i '/^Listen 80/d' /etc/httpd/conf/httpd.conf && \
+    # Configure httpd to run without root privileges
+    chgrp root /var/run/httpd && \
+    chmod g+rwx /var/run/httpd && \
+    chgrp root /var/log/httpd && \
+    chmod g+rwx /var/log/httpd
+
+# Build the RPM manifest
```
bdunne, line 26: Any reason to merge the other `RUN` steps but not this one?

images/manageiq-base/container-assets/clean_dnf_rpm
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
bdunne, line 1: Not for this PR, but I wonder if we should put this script in one of our RPMs and run it on the appliance to reduce image size there too

images/manageiq-base/Dockerfile
bdunne: I'm fine with moving them up, they're only used at runtime, not build time so the order doesn't matter.

bdunne: Not for this PR, but dumb-init is up to v1.2.5 now, so we should probably update.  Also, should we rely on pulling it from Github at every build or do you think we should keep a copy somewhere else (package it in an RPM)?

images/manageiq-base/Dockerfile
bdunne: Naming is hard 😄

images/manageiq-base/Dockerfile
Fryguy: Oh yes I saw that - will have a follow-up PR

images/manageiq-base/container-assets/clean_dnf_rpm
Fryguy: Oh that's an interesting idea - agreed that we can do that in a follow up

images/manageiq-webserver-worker/Dockerfile
Fryguy: It doesn't save any space to split them up, though admittedly it would drop the need to duplicate the clean_dnf_rpm call.  I'd like to leave it for now.

images/manageiq-base/Dockerfile
Fryguy: @bdunne Updated.  I moved up both commands

Fryguy:
I forgot to mention this also has significant time savings as well.  Locally, the rpm manifest generation steps can take  ~100s x 5 image = ~500s (8m20s).  After this PR, there are only 2 calls, so that's 200s (3m20s) for a savings of ~5min per build

Updated the OP with this information.

images/manageiq-ui-worker/Dockerfile
bdunne: I think we used to install something here, but it was moved up in the layers.

PR #846, opened 2022-07-08
Avoid installing weak dependencies 

Fryguy:
Depends on
- [x] https://github.com/ManageIQ/manageiq-rpm_build/pull/289

The following weak dependencies will end up not being installed.

```
apr-util-bdb
apr-util-openssl
geolite2-city
geolite2-country
libmaxminddb
perl-Digest
perl-Digest-MD5
perl-IO-Socket-IP
perl-IO-Socket-SSL
perl-libnet
perl-Mozilla-CA
perl-Net-SSLeay
perl-URI
python38-pip
python3-jmespath
rubygem-bundler
rubygem-json
rubygem-rdoc
```

This drops the image size roughly from 2.063GB to 1.943GB (120MB, 5.85%).  Of note, the geolite2 packages are somewhat large, so part of #736 

---

Packages sizes sorted by size:

```
56549434  geolite2-city
 7784945  python38-pip
 3424334  geolite2-country
 1714611  rubygem-rdoc
 1359290  rubygem-bundler
 1325271  perl-Net-SSLeay
  618705  perl-IO-Socket-SSL
  277135  perl-libnet
  216452  perl-URI
  119891  python3-jmespath
   99525  perl-IO-Socket-IP
   97777  rubygem-json
   56718  perl-Digest-MD5
   54260  libmaxminddb
   26685  perl-Digest
   20448  apr-util-openssl
   11920  apr-util-bdb
    5716  perl-Mozilla-CA
```

PR #847, opened 2022-07-11
Set dnf install_weak_deps=False and nodocs globally

bdunne:
Followup to https://github.com/ManageIQ/manageiq-pods/pull/846#pullrequestreview-1034724336


bdunne:
Backport to morphy requires https://github.com/ManageIQ/manageiq-rpm_build/pull/473

PR #859, opened 2022-07-29
OPERATOR_NAME is no longer used

bdunne:
It was used for leader election lock naming in older versions of the operator-sdk.
Dropped by https://github.com/ManageIQ/manageiq-pods/pull/831/files#diff-963be88783f1e1465a03f13199442714132415353ddb038a0f52f6442dde70ceL87-L97


Fryguy:
YOLO!

PR #861, opened 2022-08-02
Rely on make docker-build for consistency

bdunne:
- After upgrading to operator-sdk 1.x we use make, not operator-sdk to build the operator image
- Also move options to Makefile

Fixes: `must run command in project root dir: project structure requires build/Dockerfile`

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -106,26 +106,7 @@ popd
 
 if [ -z "$NO_OPERATOR" ]; then
   pushd "$OPERATOR_DIR"
-    echo "Building manageiq-operator manifest..."
-    echo "$(date +"%Y%m%d%H%M%S")-$(git rev-parse HEAD)" > ./BUILD
-    cat ./BUILD
-    echo
-
-    image_build_args="--image-build-args --pull"
-    image_builder=""
-
-    if [ "$ARCH" = "s390x" ]; then
-      # HACK: Using buildah as, under the covers, this uses --format docker,
-      # which is needed to build v2 manifest builds. Normally, we'd pass
-      # --format docker as above as another arg to image-build-args, however,
-      # there is a bug in operator-sdk v0.18.2 where it cannot accept multiple
-      # args in image-build-args.
-      #
-      # See also https://github.com/operator-framework/operator-sdk/blob/v0.18.2/cmd/operator-sdk/build/cmd.go#L75-L99
-      image_builder="--image-builder buildah"
-    fi
-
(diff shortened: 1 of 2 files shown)


manageiq-operator/Makefile
```diff
@@ -119,7 +119,8 @@ run: manifests generate fmt vet ## Run a controller from your host.
 .PHONY: docker-build
 docker-build: test ## Build docker image with the manager.
 	echo "$(date +"%Y%m%d%H%M%S")-$(git rev-parse HEAD)" > ./BUILD
-	docker build -t ${IMG} .
+	cat ./BUILD
+	docker build -t ${IMG} . --pull --format docker
```
Fryguy, line 123: This might affect s390x where we need buildah format

bdunne: Why do we need buildah format on s390x?

bdunne: Also, a few lines above we're tacking on `cmd+=" --format docker"` for s390x builds of the other images.

Fryguy: Ah ok then I mixed it up... We tried a number of passes before finding the one that works

bin/build
```diff
@@ -106,26 +106,7 @@ popd
...
-    if [ "$ARCH" = "s390x" ]; then
-      # HACK: Using buildah as, under the covers, this uses --format docker,
-      # which is needed to build v2 manifest builds. Normally, we'd pass
-      # --format docker as above as another arg to image-build-args, however,
-      # there is a bug in operator-sdk v0.18.2 where it cannot accept multiple
-      # args in image-build-args.
-      #
-      # See also https://github.com/operator-framework/operator-sdk/blob/v0.18.2/cmd/operator-sdk/build/cmd.go#L75-L99
-      image_builder="--image-builder buildah"
-    fi
```
Fryguy, line 126: This is needed and is what I meant by needing `buildah` (I thought it was --format buildah, but it's --image-builder buildah).

Note that the new operator-sdk might _not_ need this as it was a hack to begin with as per the comment.

Fryguy: Ok, as discussed... this hack was because we _couldn't_ pass `--format docker` and `-image-builder buildah` was our way to hack that value in because operator-sdk was constructing the docker build command line for us.  Since the makefile calls docker build directly, we don't have that restriction anymore.

bdunne: https://github.com/ManageIQ/manageiq-pods/pull/862

PR #862, opened 2022-08-02
Always build the manifest in docker format

bdunne:
Followup to https://github.com/ManageIQ/manageiq-pods/pull/861

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -57,6 +57,7 @@ fi
 
 pushd $IMAGE_DIR
   cmd="docker build --tag $REPO/manageiq-base:$TAG \
+                    --format docker
                     --pull \
                     --build-arg BUILD_REF=$BUILD_REF \
                     --build-arg BUILD_ORG=$BUILD_ORG \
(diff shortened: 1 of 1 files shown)


bin/build
```diff
@@ -57,6 +57,7 @@ fi
 
 pushd $IMAGE_DIR
   cmd="docker build --tag $REPO/manageiq-base:$TAG \
+                    --format docker
```
Fryguy, line 60: You can rm lines 82-84 below.

bin/build
```diff
@@ -89,16 +90,12 @@ pushd $IMAGE_DIR
...
   for image in $build_images; do
     cmd="docker build --tag $REPO/$image:$TAG \
+                      --format docker \
                       --build-arg FROM_REPO=$REPO \
                       --build-arg FROM_TAG=$TAG \
                       --build-arg RPM_PREFIX=$RPM_PREFIX \
                       $image"
 
-    if [ "$ARCH" = "s390x" ]; then
-      # --format docker is needed to ensure the builds are built with docker v2 manifests
```
Fryguy, line 98: can we keep this comment somewhere?  If not in code then at least the commit message?

Issue #695, opened 2021-03-10
container_env is relatively slow


jrafanie:
Using the following script to profile bash, we can see 3 commands make container_env slow:

```
PS4='+ $(date "+%s.%N")\011 '
set -x

source /var/www/miq/vmdb/container_env

set +x
```
https://gist.github.com/jrafanie/56273d25ac84fefa54f8c2299c79ac7e

EDIT: this brilliant hack came from this: https://stackoverflow.com/a/5015179

1) gem env path 0.32888 from the gemset enable
2) 2 calls to CGI escape the DB user and password, roughly 0.14 seconds each

![Screen Shot 2021-03-10 at 4 20 56 PM](https://user-images.githubusercontent.com/19339/110703225-80880080-81c1-11eb-9f00-a34a43e052db.png)

Note, this is much less of a problem now that the liveness probe [no longer calls container_env every 15 seconds](https://github.com/ManageIQ/manageiq/issues/21090), but this was opened so we can investigate solutions if it's needed to be run more often than 1) once during the pod entrypoint and 2) whenever I need to rsh in and load the ruby environment.

@miq-bot add-label bug

Fryguy:
The `gem env path` comes from here: https://github.com/ManageIQ/manageiq-rpm_build/blob/8f66a73a18dffbc73af613e4ce5dd611e7d75c1e/rpm_spec/manageiq.spec.in#L54 .  The purpose of this enable script is when the gemset rpm is enabled, we need to prepend our gemset gem path to the existing gem path.  However, at this point, GEM_PATH is not yet defined, so we need to figure out what the current one is.  There might be a more efficient method to this than shelling out.  Since this is for containers we _might_ be able to just hardcode the path, but if things change, we're going to have to remember to change this as well.

Fryguy:
For the CGI.escape, https://github.com/ManageIQ/manageiq-pods/pull/694 will cut the time a bit.  If we really care to make it faster we can also just smash the 2 separate calls into a single call which should give us another 2x (since the expensive part is starting ruby itself).  Alternatively, there might be a shell function we can grab to avoid ruby altogether.

Fryguy:
@jrafanie For the gem env path, I'm seeing roughly 50% improvement locally if I do this:

```
$ time gem env path
/Users/jfrey/.gem/ruby/2.6.6:/Users/jfrey/.rubies/ruby-2.6.6/lib/ruby/gems/2.6.0
[...]

PR #871, opened 2022-08-11
Simplify DATABASE_URL generation

Fryguy:
This PR simplifies the creation of the DATABASE_URL by only shelling out to Ruby once instead of twice to build the userinfo.

Using the script from https://github.com/ManageIQ/manageiq-pods/issues/695 for timing, this eliminates one of the Ruby runs, dropping the time (for those 2 calls) from ~0.06s to 0.03s

### DIFF
--- a/images/manageiq-base/container-assets/container_env
+++ b/images/manageiq-base/container-assets/container_env
@@ -4,19 +4,8 @@ update-ca-trust
 
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 
-function urlescape() {
-  PAYLOAD="$1" ruby --disable-gems -rcgi -e "puts CGI.escape(ENV['PAYLOAD'])"
-}
-
-safeuser=$(urlescape ${DATABASE_USER})
-safepass=$(urlescape ${DATABASE_PASSWORD})
-if [ -n "$safeuser" -a -n "$safepass" ]; then
-  safeuserinfo="${safeuser}:${safepass}@"
-else
-  safeuserinfo=""
-fi
-
-database_url="postgresql://${safeuserinfo}${DATABASE_HOSTNAME:-localhost}:${DATABASE_PORT:-5432}/${DATABASE_NAME:-db_unknown}?encoding=utf8&pool=5&wait_timeout=5&sslmode=${DATABASE_SSL_MODE:-prefer}"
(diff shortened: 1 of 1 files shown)


kbrock:
Not sure if it affects us, but this will call out to ruby instead of using bash.
And we got hit pretty hard by that before.

Do we think this will get called that often?

images/manageiq-base/container-assets/container_env
```diff
@@ -4,19 +4,8 @@ update-ca-trust
...
-fi
-
-database_url="postgresql://${safeuserinfo}${DATABASE_HOSTNAME:-localhost}:${DATABASE_PORT:-5432}/${DATABASE_NAME:-db_unknown}?encoding=utf8&pool=5&wait_timeout=5&sslmode=${DATABASE_SSL_MODE:-prefer}"
-[[ -f /.postgresql/root.crt ]] && database_url="$database_url&sslrootcert=/.postgresql/root.crt"
-
-export DATABASE_URL=$database_url
+DATABASE_USERINFO=""
+[[ -n "$DATABASE_USER" && -n "$DATABASE_PASSWORD" ]] && DATABASE_USERINFO="$(ruby --disable-gems -rcgi -e "puts \"#{CGI.escape(ENV['DATABASE_USER'])}:#{CGI.escape(ENV['DATABASE_PASSWORD'])}@\"")"
+DATABASE_QUERY="encoding=utf8&pool=5&wait_timeout=5&sslmode=${DATABASE_SSL_MODE:-prefer}"
+[[ -f /.postgresql/root.crt ]] && DATABASE_QUERY="${DATABASE_QUERY}&sslrootcert=/.postgresql/root.crt"
```
kbrock, line 10: @bdunne is this where we put the root cert on pods?

Fryguy:
> Not sure if it affects us, but this will call out to ruby instead of using bash.

Old code called out twice, new code calls out once.  Either way it was using Ruby.

images/manageiq-base/container-assets/container_env
jrafanie: I'll merge anyway... this is unchanged... we can fix it in a followup if there's a concern.

jrafanie (approved):
LGTM, we can followup as I think the code is unchanged, just one less call to ruby.

PR #872, opened 2022-08-11
Check for the presence of the certs before calling update-ca-trust

Fryguy:
Built on #871 

When certs are not present, `time source container_env` drops from
0m0.266s to 0m0.011s

Closes https://github.com/ManageIQ/manageiq-pods/issues/695

@jrafanie Please review

### DIFF
--- a/images/manageiq-base/container-assets/container_env
+++ b/images/manageiq-base/container-assets/container_env
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-update-ca-trust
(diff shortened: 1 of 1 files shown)


jrafanie:
@bdunne can you review?  Is this a valid assumption that we can skip it if the file doesn't exist?

jrafanie (approved):
LGTM, but am unsure if this assumption is valid.  Want confirmation.

PR #874, opened 2022-08-16
Add app label to the operator pod

bdunne:
This will apply the default-deny network policy to the operator pod

Before:
$ curl http://ip_of_operator_pod:8080/abc
404 page not found

After:
$ curl http://ip_of_operator_pod:8080/abc
no response


Fryguy:
Skipping backport to `oparin`, because it is already in the branch.

PR #875, opened 2022-08-16
go get -t -u ./... && go mod tidy

bdunne:


Fryguy:
Does it make sense to automate this, say, weekly, similar to how we do the AWS instance type updates?

bdunne:
https://github.com/ManageIQ/manageiq-pods/pull/878

PR #877, opened 2022-08-17
Fix make manifests

bdunne:
- Fix roleName
- drop creationTimestamp and namespace from role.yaml
- remove status fron CRD since we don't set status yet

### DIFF
--- a/manageiq-operator/Makefile
+++ b/manageiq-operator/Makefile
@@ -88,7 +88,8 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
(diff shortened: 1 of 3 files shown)


manageiq-operator/Makefile
```diff
@@ -88,7 +88,8 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manageiq-operator crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	sed -i "/creationTimestamp\|namespace/d" config/rbac/role.yaml
```
Fryguy, line 92: This is going to fail in MacOS.  I'll show you offline, but the fix is obnoxious.

bdunne: Fixed

Fryguy (approved):

PR #879, opened 2022-08-24
Switch to UBI 8 minimal

bdunne:


manageiq-operator/Dockerfile
```diff
@@ -17,9 +17,7 @@ COPY controllers/ controllers/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux go build -a -o manager main.go
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
```
jrafanie, line 20: Will this change what versions of packages will be on the operator? Do we need to run builds with this before merge?

bdunne: Yes, it will change a bunch of OS packages, but those aren't used by the operator code.
We used to use this image, but the operator-sdk upgrade repaced it with the distroless image. I don't think we need to do any special testing, it just runs the prebuilt binary.

jrafanie (approved):

PR #880, opened 2022-08-24
Pass the ApplicationDomain to the orchestrator

bdunne:
First step to getting the remote_*_url methods working in pods.  I'm thinking that the orchestrator would then stick this into the region settings.

### DIFF
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/orchestrator.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/orchestrator.go
@@ -184,6 +184,7 @@ func addPostgresConfig(cr *miqv1alpha1.ManageIQ, d *appsv1.Deployment, client cl
 func updateOrchestratorEnv(cr *miqv1alpha1.ManageIQ, c *corev1.Container) {
 	c.Env = addOrUpdateEnvVar(c.Env, corev1.EnvVar{Name: "ADMIN_GROUP", Value: cr.Spec.InitialAdminGroupName})
 	c.Env = addOrUpdateEnvVar(c.Env, corev1.EnvVar{Name: "APP_NAME", Value: cr.Spec.AppName})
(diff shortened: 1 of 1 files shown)


manageiq-operator/api/v1alpha1/helpers/miq-components/orchestrator.go
```diff
@@ -184,6 +184,7 @@ func addPostgresConfig(cr *miqv1alpha1.ManageIQ, d *appsv1.Deployment, client cl
 func updateOrchestratorEnv(cr *miqv1alpha1.ManageIQ, c *corev1.Container) {
 	c.Env = addOrUpdateEnvVar(c.Env, corev1.EnvVar{Name: "ADMIN_GROUP", Value: cr.Spec.InitialAdminGroupName})
 	c.Env = addOrUpdateEnvVar(c.Env, corev1.EnvVar{Name: "APP_NAME", Value: cr.Spec.AppName})
+	c.Env = addOrUpdateEnvVar(c.Env, corev1.EnvVar{Name: "APPLICATION_DOMAIN", Value: cr.Spec.ApplicationDomain})
```
jrafanie, line 187: is this run once per orchestrator startup or each iteration of the reconcile loop?  Is this a value that could change and we'd need to deal with updating here and then syncing in the region settings later?

jrafanie: in other words, if someone were to try to change this, is this doable and will it be picked up or would they have to bounce the orchestrator and all the pods?

bdunne: Yes, if someone updates the CR changing the ApplicationDomain, the ENV will be updated causing the orchestrator to get redeployed.

jrafanie (approved):

PR #883, opened 2022-08-30
 Use podman if available, fall back to docker 

bdunne:
If both podman and docker executables are installed (GHA ubuntu-latest), we were adding `--format docker` and sending it to docker which is a problem since that's not an option on Docker.

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -40,7 +40,7 @@ BASE_DIR=${BASE_DIR:-$PWD}
 IMAGE_DIR="$BASE_DIR/images"
 OPERATOR_DIR="$BASE_DIR/manageiq-operator"
 
-HAS_PODMAN="$(which podman &>/dev/null && echo "true")"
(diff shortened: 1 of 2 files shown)


bdunne:
@Fryguy Any thoughts here?

Fryguy:
yeah I'm not sure - it feels weird that we built a specific script that wraps it, when the existing docker command on podman already wraps it.

Fryguy:
I'm wondering if maybe we name the script `docker`, or even better, just have a docker function that we source in which overrides things. I have something similar locally that works well (I use it for custom docker commands:

```bash
docker () {
	custom_docker_cmd="docker-$1"
	if [ -x "$(which "$custom_docker_cmd")" ]
	then
		shift
		$custom_docker_cmd "$@"
	else
		command docker "$@"
	fi
}
```

Fryguy (approved):

PR #889, opened 2022-09-19
Add a build action

bdunne:
- [x] Depends on: https://github.com/ManageIQ/manageiq-rpm_build/pull/303

~~TODO:~~
~~- [ ] Tag builds~~ We'll do this on the build machines for now

### DIFF
--- a/.github/workflows/build_pods.yaml
+++ b/.github/workflows/build_pods.yaml
@@ -1,0 +1,19 @@
+name: Build Pods
+on:
+  repository_dispatch:
+    types:
+    - build
+jobs:
+  build_pods:
+    if: github.repository_owner == 'ManageIQ'
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version-file: 'manageiq-operator/go.mod'
+    - name: Docker login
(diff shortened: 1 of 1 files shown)


.github/workflows/build_pods.yaml
```diff
@@ -0,0 +1,21 @@
...
+  repository_dispatch:
+    types:
+    - build
+jobs:
+  build_pods:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        repository: ManageIQ/manageiq-pods
```
Fryguy, line 12: This shouldn't be necessary as it's the implied default (and probably makes forks harder)

.github/workflows/build_pods.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - name: Set up Go
+      uses: actions/setup-go@v3
+      with:
+        go-version-file: 'manageiq-operator/go.mod'
+    - name: Docker login
+      if: ${{ github.event_name != 'pull_request' }}
```
Fryguy, line 18: This will still trigger on branch forks, which we probably don't want?

.github/workflows/build_pods.yaml
bdunne: Done

.github/workflows/build_pods.yaml
bdunne: Done

PR #896, opened 2022-10-06
Enforce and default Kafka configuration

nasark:
- Enforce Kafka by making the orchestrator wait for Kafka service to be ready (requires Kafka network policy)
- default `deployMessagingService` to true from CR as Kafka will be the default configuration -> this will retain the option to disable Kafka in case things go wrong

Depends on:
- [x] https://github.com/ManageIQ/manageiq-pods/pull/935
- [x] https://github.com/ManageIQ/manageiq-pods/pull/936

Needed for:
- [ ] https://github.com/ManageIQ/manageiq-pods/pull/1005

Ref:
- https://github.com/ManageIQ/manageiq/issues/22225

@miq-bot assign @Fryguy 
@miq-bot add_reviewer @agrare

### DIFF
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/cr.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/cr.go
@@ -51,7 +51,7 @@ func databaseVolumeCapacity(cr *miqv1alpha1.ManageIQ) string {
 
 func deployMessagingService(cr *miqv1alpha1.ManageIQ) bool {
 	if cr.Spec.DeployMessagingService == nil {
-		return false
+		return true
 	} else {
 		return *cr.Spec.DeployMessagingService
 	}
--- a/manageiq-operator/api/v1alpha1/manageiq_types.go
+++ b/manageiq-operator/api/v1alpha1/manageiq_types.go
@@ -64,7 +64,7 @@ type ManageIQSpec struct {
 	// +optional
 	DatabaseVolumeCapacity string `json:"databaseVolumeCapacity,omitempty"`
 
-	// Flag to indicate if Kafka and Zookeeper should be deployed (default: false)
+	// Deprecated: Flag to indicate if Kafka and Zookeeper should be deployed (default: true)
 	// +optional
 	DeployMessagingService *bool `json:"deployMessagingService,omitempty"`
 
--- a/images/manageiq-orchestrator/container-assets/entrypoint
+++ b/images/manageiq-orchestrator/container-assets/entrypoint
@@ -117,6 +117,10 @@ EOS
 check_svc_status ${MEMCACHED_SERVICE_HOST} ${MEMCACHED_SERVICE_PORT}
 check_svc_status ${database_hostname} ${database_port}
 
+if [ -n "$MESSAGING_HOSTNAME" ] && [ -n "$MESSAGING_PORT" ]; then
+  check_svc_status ${MESSAGING_HOSTNAME} ${MESSAGING_PORT}
+fi
+
 check_deployment_status || exit 1
 
 pushd ${APP_ROOT}
--- a/manageiq-operator/config/crd/bases/manageiq.org_manageiqs.yaml
+++ b/manageiq-operator/config/crd/bases/manageiq.org_manageiqs.yaml
@@ -68,8 +68,8 @@ spec:
                 description: 'Database volume size (default: 15Gi)'
                 type: string
               deployMessagingService:
-                description: 'Flag to indicate if Kafka and Zookeeper should be deployed
-                  (default: false)'
+                description: 'Deprecated: Flag to indicate if Kafka and Zookeeper
+                  should be deployed (default: true)'
                 type: boolean
               enableApplicationLocalLogin:
                 description: 'Flag to allow logging into the application without SSO


Fryguy:
@bdunne Please review.

manageiq-operator/api/v1alpha1/manageiq_types.go
```diff
@@ -63,10 +63,6 @@ type ManageIQSpec struct {
 	// +optional
 	DatabaseVolumeCapacity string `json:"databaseVolumeCapacity,omitempty"`
 
-	// Flag to indicate if Kafka and Zookeeper should be deployed (default: false)
-	// +optional
-	DeployMessagingService *bool `json:"deployMessagingService,omitempty"`
```
Fryguy, line 68: Was this ever shipped? or was this transient?  We may need to keep it since it might break backward compat.

nasark: I see, would it better to keep the `deployMessagingService` option and set its default value to `true` as opposed to the current `false` then?

agrare:
@nasark so this removes the option to deploy kafka, does this actually do the deployment?

nasark:
> @nasark so this removes the option to deploy kafka, does this actually do the deployment?

Actually this removes the flag that would disable Kafka by default. Removing the flag altogether makes it so that Kafka is deployed alongside the other services by default.

agrare:
Awesome thanks @nasark can you tell what it deploys is it ManageIQ/container-kafka ?

nasark:
If removing the flag altogether is a concern for backwards compatibility, then we could just set the default value for `deployMessagingService` to `true` as well. Some clarification and/or elaboration regarding https://github.com/ManageIQ/manageiq-pods/pull/896#discussion_r995099738 would be appreciated @Fryguy

nasark:
> Awesome thanks @nasark can you tell what it deploys is it ManageIQ/container-kafka ?

The resources deployed are handled here https://github.com/ManageIQ/manageiq-pods/blob/30480277dff0accfb679bef93b9d206257f7e0f3/manageiq-operator/controllers/manageiq_controller.go#L364-L429

Fryguy:
> If removing the flag altogether is a concern for backwards compatibility, then we could just set the default value for `deployMessagingService` to `true` as well. Some clarification/elaboration regarding [#896 (comment)](https://github.com/ManageIQ/manageiq-pods/pull/896#discussion_r995099738) would be appreciated @Fryguy

We may also need a CR migration, but it might not matter if we just don't honor it?  cc @bdunne

Fryguy:
@agrare That would explain why that image hasn't been updated in years 😆

nasark:
Returning to this, perhaps it is easier to just default `deployMessagingService` to `true` instead of removing it from the CR altogether. Let me know what you think and how we can proceed with this @Fryguy @bdunne

nasark:
I've split some of this up and moved into other PRs based on issues I have found in my testing:
- https://github.com/ManageIQ/manageiq-pods/pull/935
- https://github.com/ManageIQ/manageiq-pods/pull/936

manageiq-operator/api/v1alpha1/helpers/miq-components/network_policies.go
```diff
@@ -217,44 +217,6 @@ func NetworkPolicyAllowPostgres(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme
 	return networkPolicy, f
 }
 
-func NetworkPolicyAllowKafka(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme, c *client.Client) (*networkingv1.NetworkPolicy, controllerutil.MutateFn) {
```
Fryguy, line 220: Why does this go away?

nasark:
@miq-bot remove_label stale

nasark:
@miq-bot remove_label stale

nasark:
@miq-bot add_label quinteros/yes?

manageiq-operator/api/v1alpha1/manageiq_types.go
```diff
@@ -64,8 +64,7 @@ type ManageIQSpec struct {
 	// +optional
 	DatabaseVolumeCapacity string `json:"databaseVolumeCapacity,omitempty"`
 
-	// Flag to indicate if Kafka and Zookeeper should be deployed (default: false)
-	// +optional
```
bdunne, line 68: I think we need to keep this comment

nasark: I've removed the deprecated bit, do you want to keep the `optional` comment too? Technically it will be used by default due to the `true` value but don't think that makes it optional?

manageiq-operator/api/v1alpha1/manageiq_types.go
```diff
@@ -64,8 +64,7 @@ type ManageIQSpec struct {
...
+	// Flag to indicate if Kafka and Zookeeper should be deployed (default: true)
```
bdunne, line 67: The deprecated part was fine.  We need to keep the `+optional` otherwise the CR validation will require users to fill in that field.  See `+optional` in https://book.kubebuilder.io/reference/markers/crd-validation
```suggestion
	// Deprecated: Flag to indicate if Kafka and Zookeeper should be deployed (default: true)
	// +optional
```

manageiq-operator/api/v1alpha1/helpers/miq-components/cr.go
```diff
@@ -51,7 +51,7 @@ func databaseVolumeCapacity(cr *miqv1alpha1.ManageIQ) string {
 
 func deployMessagingService(cr *miqv1alpha1.ManageIQ) bool {
 	if cr.Spec.DeployMessagingService == nil {
-		return false
+		return true
```
bdunne, line 54: I think we can remove the conditional completely and always return true unless we add some conditional logic to the service checking in the entrypoint script.  As is it, the orchestrator will wait foerver for the kafka if the user sets this to false, right?

Fryguy:
Backported to `quinteros` in commit fe79478dc8a65610d2743801b4973825fffe5985.

```text
commit fe79478dc8a65610d2743801b4973825fffe5985
Date:   Wed Jan 3 12:06:32 2024 -0500

    Merge pull request #896 from nasark/require_kafka
    
    Enforce and default Kafka configuration
    
    (cherry picked from commit 09da4c7f3083fad1b56f345d7e6e8c0063a37f58)
```

PR #906, opened 2022-10-25
Remove manageiq-appliance-build clone and customization hook

Fryguy:
Updated/Rebased version of #636 

manageiq-appliance-build repo.

With this change, everything is built in manageiq-rpm_build, and no need for source customization in pods, which obsoletes CORE_REPO_NAME, GIT_AUTH and GIT_HOST.


Fryguy:
@bdunne This is ready to go.

PR #907, opened 2022-10-27
Revert back to UBI8 for now.

bdunne:
UBI9 requires Power9 or above and Z14 or above.
I don't think we want to force users to upgrade hardware at this time.


Fryguy:
Should we use proper revert commits or is that not necessary / not possible?

PR #923, opened 2022-12-13
[WIP] Move server certificates to /etc/pki

bdunne:
Part of: https://github.com/ManageIQ/manageiq-pods/pull/770


Fryguy:
If we have to merge the container-postgresql one, then we'll have to backport this PR to all expected release versions, right?

bdunne:
Correct

kbrock:
While I like the dry nature of this PR.
My main push for consolidating the keys is so we aren't writing keys to `/root/`. (so user `manageiq` has access to them.)

So if keeping this key in with postgres is less work, I'm ok keeping it the way it is.
Guess the same would go for httpd encryption key.

PR #928, opened 2023-01-10
Apply a default security context to the containers

bdunne:
- AllowPrivilegeEscalation: false
- Privileged: false
- RunAsNonRoot: true
- Capabilities: Drop all


bdunne:
I'm going to open a similar PR on core for the workers.

PR #930, opened 2023-01-19
Read secrets from an ENV file if mounted

bdunne:
Having secrets in ENV vars may be less secure than having files on the filesystem.  Move sensitive information into files.
- [x] v2_key
- [x] database.yml
- [x] [postgres container](https://github.com/ManageIQ/container-postgresql/pull/49)

Depends on:
- [x] https://github.com/ManageIQ/manageiq-pods/pull/932
- [x] https://github.com/ManageIQ/container-postgresql/pull/49

### DIFF
--- a/images/manageiq-base/container-assets/container_env
+++ b/images/manageiq-base/container-assets/container_env
@@ -4,4 +4,45 @@
 
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 
-[ ! -f /var/www/miq/vmdb/config/database.yml ] && ln -s /var/run/secrets/manageiq/config/database.yml /var/www/miq/vmdb/config/
+if [[ ! -f /var/www/miq/vmdb/config/database.yml ]]; then
+  [[ -f /run/secrets/postgresql/POSTGRESQL_HOSTNAME ]] && database_hostname_file=$(cat /run/secrets/postgresql/POSTGRESQL_HOSTNAME)
+  [[ -f /run/secrets/postgresql/POSTGRESQL_PASSWORD ]] && database_password_file=$(cat /run/secrets/postgresql/POSTGRESQL_PASSWORD)
+  [[ -f /run/secrets/postgresql/POSTGRESQL_PORT ]] && database_port_file=$(cat /run/secrets/postgresql/POSTGRESQL_PORT)
+  [[ -f /run/secrets/postgresql/POSTGRESQL_USER ]] && database_username_file=$(cat /run/secrets/postgresql/POSTGRESQL_USER)
+  database_hostname=${DATABASE_HOSTNAME:-$database_hostname_file}
+  database_password=${DATABASE_PASSWORD:-$database_password_file}
+  database_port=${DATABASE_PORT:-$database_port_file}
+  DATABASE_SSL_MODE=${DATABASE_SSL_MODE:-prefer}
+  database_username=${DATABASE_USERNAME:-$database_username_file}
+
+  echo "== Writing database config =="
+  cat > /var/www/miq/vmdb/config/database.yml << KEY
+---
+production:
+  adapter: postgresql
+  username: ${database_username}
+  password: ${database_password}
+  host: ${database_hostname}
+  database: vmdb_production
+  port: ${database_port}
+  encoding: utf8
+  pool: 5
+  wait_timeout: 5
+  ssl_mode: ${DATABASE_SSL_MODE}
+KEY
+
+  [[ -f /.postgresql/root.crt ]] && echo "  sslrootcert=/.postgresql/root.crt" >> /var/www/miq/vmdb/config/database.yml
+fi
+
+if [[ ! -f /var/www/miq/vmdb/certs/v2_key ]]; then
+  [[ -f /run/secrets/manageiq/application/encryption_key ]] && encryption_key_file=$(cat /run/secrets/manageiq/application/encryption_key)
+  encryption_key=${ENCRYPTION_KEY:-$encryption_key_file}
+  echo "== Writing encryption key =="
+  cat > /var/www/miq/vmdb/certs/v2_key << KEY
+---
+:algorithm: aes-256-cbc
+:key: ${encryption_key}
+KEY
+fi
+
+echo "${GUID}" > ${APP_ROOT}/GUID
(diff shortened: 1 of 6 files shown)


Fryguy:
As discussed, let's try this direction (unless you think of the reason why you needed to create the files).

- Mount the secrets directly and make it "look like" the env var (e.g. for encryption-key it would mount as `/run/secrets/manageiq/ENCRYPTION_KEY` or if you can't share the mount point, something like `/run/secrets/ENCRYPTION_KEY/value`
- The entrypoint scripts would look for the file first, and then if not present use an ENV var.  This allows ENV vars to continue to work, particularly for dev use cases. Alternatively you can check the ENV var first and then check the file...this one is a little easier because you can use bash string methods:

   ```bash
   encryption_key=${ENCRYPTION_KEY:-$(cat /run/secrets/manageiq/ENCRYPTION_KEY 2>/dev/null)}
   ```
- For the container-postgresql, maybe mount the same way as the previous bullet, but then create a custom entrypoint that maps those to ENV vars, and then `execs` the original entrypoint.  We'd have to check that the ENV vars stay local after exec and aren't exported (it's possible they are exported anyway).
- Let's go back to creating the database.yml file in the entrypoint script.  This way we don't need to expose DATABASE_URI as an env var at all.  Since you are creating a database.yml, then we also don't need to convert to a URI, which gets rid of those CGI.escape hacks.

images/manageiq-base/container-assets/container_env
```diff
@@ -4,8 +4,45 @@
...
+  encoding: utf8
+  pool: 5
+  wait_timeout: 5
+  ssl_mode: ${DATABASE_SSL_MODE}
+KEY
+
+  [[ -f /.postgresql/root.crt ]] && echo "  sslrootcert=/.postgresql/root.crt" >> /var/www/miq/vmdb/config/database.yml
+fi
+
+if [[ ! -f /var/www/miq/vmdb/certs/v2_key ]]; then
```
Fryguy, line 37: super-minor, but we should probably use ${APP_ROOT} anywhere we have /var/www/miq/vmdb, if possible

images/manageiq-base/container-assets/container_env
```diff
@@ -4,8 +4,45 @@
 
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 
-DATABASE_USERINFO=""
-[[ -n "$DATABASE_USER" && -n "$DATABASE_PASSWORD" ]] && DATABASE_USERINFO="$(ruby --disable-gems -rcgi -e "puts \"#{CGI.escape(ENV['DATABASE_USER'])}:#{CGI.escape(ENV['DATABASE_PASSWORD'])}@\"")"
-DATABASE_QUERY="encoding=utf8&pool=5&wait_timeout=5&sslmode=${DATABASE_SSL_MODE:-prefer}"
-[[ -f /.postgresql/root.crt ]] && DATABASE_QUERY="${DATABASE_QUERY}&sslrootcert=/.postgresql/root.crt"
-export DATABASE_URL="postgresql://${DATABASE_USERINFO}${DATABASE_HOSTNAME:-localhost}:${DATABASE_PORT:-5432}/${DATABASE_NAME:-db_unknown}?${DATABASE_QUERY}"
+if [[ ! -f /var/www/miq/vmdb/config/database.yml ]]; then
```
Fryguy, line 7: if you don't have a database.yml file then this section runs but if you _do_ have a database.yml file, then database-hostname will never be set.  This means later in the check_svc_status in the entrypoint, you won't have a value.  I think, technically, you need to pull the values from the database.yml if present.

I won't let this hold up the PR, since in reality we never have a database.yml, but then it does bring into question why have this check at all.

images/manageiq-base/container-assets/container_env
```diff
@@ -4,8 +4,45 @@
 
 [[ -s /etc/default/evm ]] && source /etc/default/evm
 
-DATABASE_USERINFO=""
-[[ -n "$DATABASE_USER" && -n "$DATABASE_PASSWORD" ]] && DATABASE_USERINFO="$(ruby --disable-gems -rcgi -e "puts \"#{CGI.escape(ENV['DATABASE_USER'])}:#{CGI.escape(ENV['DATABASE_PASSWORD'])}@\"")"
```
Fryguy, line 8: Nice - this should improve the container_env startup time.

images/manageiq-base/container-assets/container_env
```diff
@@ -4,8 +4,45 @@
...
-DATABASE_QUERY="encoding=utf8&pool=5&wait_timeout=5&sslmode=${DATABASE_SSL_MODE:-prefer}"
-[[ -f /.postgresql/root.crt ]] && DATABASE_QUERY="${DATABASE_QUERY}&sslrootcert=/.postgresql/root.crt"
-export DATABASE_URL="postgresql://${DATABASE_USERINFO}${DATABASE_HOSTNAME:-localhost}:${DATABASE_PORT:-5432}/${DATABASE_NAME:-db_unknown}?${DATABASE_QUERY}"
+if [[ ! -f /var/www/miq/vmdb/config/database.yml ]]; then
+  [[ -f /run/secrets/postgresql/POSTGRESQL_HOSTNAME ]] && database_hostname_file=$(cat /run/secrets/postgresql/POSTGRESQL_HOSTNAME)
+  [[ -f /run/secrets/postgresql/POSTGRESQL_PASSWORD ]] && database_password_file=$(cat /run/secrets/postgresql/POSTGRESQL_PASSWORD)
+  [[ -f /run/secrets/postgresql/POSTGRESQL_PORT ]] && database_port_file=$(cat /run/secrets/postgresql/POSTGRESQL_PORT)
+  [[ -f /run/secrets/postgresql/POSTGRESQL_USER ]] && database_username_file=$(cat /run/secrets/postgresql/POSTGRESQL_USER)
+  database_hostname=${DATABASE_HOSTNAME:-$database_hostname_file}
```
Fryguy, line 12: For local use, we used to have defaults for hostname as localhost, port as 5432, and database as db_unknown.  I think we need to bring this over for that use case.

I won't let this one hold up merge either, since it's a pure dev-diagnostic use case.

kbrock: Not sure how this works with HA, but some cloud providers have all apps setup to just hit `localhost:5432`, and the os deals with the redirect / health.

kbrock:
Sorry I'm way late to the game, but I have to say:

@bdunne this is really good stuff

PR #932, opened 2023-01-23
Ensure that ManageAppSecret is managing the app secret

bdunne:
This bug was introduced during the operator-sdk upgrade when changing from a create only to a create or update that could be called during reconcile. I was looking up the secret using the wrong name (probably a copy/paste error).


Fryguy:
@bdunne Does this need to be backported?

PR #955, opened 2023-04-10
Script to generate Kafka keystores for SSL

nasark:
- script to generate kafka truststore and keystore files using provided CA (such as `root.crt` created in cert_generator.rb) and keystore password determined by user

Needed for:
- https://github.com/ManageIQ/manageiq-documentation/pull/1728

@miq-bot assign @bdunne 
@miq-bot add_reviewer @Fryguy 

### DIFF
--- a/tools/keystore_generator.sh
+++ b/tools/keystore_generator.sh
@@ -2,9 +2,14 @@
 
 set -e
 
-CA_CERT_PATH=$CA_CERT_PATH
-CA_KEY_PATH=$CA_KEY_PATH
-KEYSTORE_PASS=$KEYSTORE_PASS
+echo "Enter CA cert path:"
+read CA_CERT_PATH
+
+echo "Enter CA key path:"
+read CA_KEY_PATH
+
+echo "Set Keystore password:"
+read KEYSTORE_PASS
 
 if [ ! -e "$CA_CERT_PATH" ] || [ ! -e "$CA_KEY_PATH" ]; then
   echo "CA does not exist, please provide the corrrect paths in CA_CERT_PATH and CA_KEY_PATH"
--- a/tools/keystore_generator.sh
+++ b/tools/keystore_generator.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+set -e
+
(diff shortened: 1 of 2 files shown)


tools/keystore_generator.sh
```diff
@@ -0,0 +1,46 @@
...
+  exit 1
+fi
+
+if [ -z "$KEYSTORE_PASS" ]; then
+  echo "Please provide a keystore password in KEYSTORE_PASS"
+  exit 1
+fi
+
+# Generate truststore containing CA
+keytool -keystore ./kafka.truststore.jks \
```
Fryguy, line 20: Where does `keytool` come from?

tools/keystore_generator.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if [ ! -e "$CA_CERT_PATH" ] || [ ! -e "$CA_KEY_PATH" ]; then
+  echo "CA does not exist, please provide the corrrect paths in CA_CERT_PATH and CA_KEY_PATH"
+  exit 1
+fi
+
+if [ -z "$KEYSTORE_PASS" ]; then
+  echo "Please provide a keystore password in KEYSTORE_PASS"
+  exit 1
+fi
```
Fryguy, line 17: For security reasons we may want to read the password from STDIN.

tools/keystore_generator.sh
nasark: `keytool` comes with any Java distribution. I guess this means that the script assumes the user has Java on their system. Would this be an issue?

nasark:
> Is it possible to go the other direction leave generation of the certificates in the current script and have an entrypoint script on the kafka container that creates a keystore and adds those certs to it?

https://github.com/ManageIQ/container-kafka/pull/2

kbrock:
Would have preferred this being in appliance_console{_cli} but if that does not create certificates, then that request is out of scope.

Did we want the bot to close this?
Seems it is still outstanding

nasark:
This can be closed as we are no longer moving forward with Bitnami and this will not be needed for Strimzi

PR #959, opened 2023-05-04
Add a new hotfix image build script

bdunne:
- Build on top of existing images by copying over local RPMs and updating only those RPMs (as we would do in a hotfix situation between releases)
- Allow for productized source image names

### DIFF
--- a/bin/build
+++ b/bin/build
@@ -4,13 +4,10 @@ TAG=latest
 
 ARCH=`uname -m`
 
-RPM_BUILD_OPTIONS=${RPM_BUILD_OPTIONS:-""}
-RPM_BUILD_IMAGE=${RPM_BUILD_IMAGE:-"manageiq/rpm_build:$TAG"}
-RPM_PREFIX=${RPM_PREFIX:-"manageiq"}
-
-while getopts "t:d:r:hblnops" opt; do
+while getopts "t:c:d:r:hblnops" opt; do
   case $opt in
     b) REBUILD_RPM="true" ;;
+    c) CONTAINER_PREFIX=$OPTARG ;;
     d) BASE_DIR=$OPTARG ;;
     l) LOCAL_RPM="true" ;;
     n) NO_CACHE="true" ;;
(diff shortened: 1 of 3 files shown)


bin/build
```diff
@@ -28,10 +25,11 @@ if [ -n "$LOCAL_RPM" ] && [ -n "$REBUILD_RPM" ]; then
   exit 1
 fi
 
-REPO=${REPO:-docker.io/manageiq}
 BASE_DIR=${BASE_DIR:-$PWD}
+CONTAINER_PREFIX=${CONTAINER_PREFIX:-manageiq}
```
Fryguy, line 29: I assume we'll start using this in the future?

bdunne: Yes, but for now we can at least have the same options.

bin/build
```diff
@@ -4,13 +4,10 @@ TAG=latest
 
 ARCH=`uname -m`
 
-RPM_BUILD_OPTIONS=${RPM_BUILD_OPTIONS:-""}
-RPM_BUILD_IMAGE=${RPM_BUILD_IMAGE:-"manageiq/rpm_build:$TAG"}
-RPM_PREFIX=${RPM_PREFIX:-"manageiq"}
```
Fryguy, line 9: I dont't understand removing these as they are used down below in https://github.com/ManageIQ/manageiq-pods/pull/959/files#diff-1fba65931954c73107694f0b7b964148febc5ba35da6812e1e117dfb37436e77R42-R47

Fryguy: That being said I never understood building the RPMs as part of this script - always felt like if that's what's wanted the rpms should be built separately and then passed in.

Fryguy:
Backported to `petrosian` in commit 0c9072308a5a161cb0185a95ac461445133dd811.

```text
commit 0c9072308a5a161cb0185a95ac461445133dd811
Date:   Fri May 5 12:03:43 2023 -0400

    Merge pull request #959 from bdunne/add_hotfix_build_script
    
    Add a new hotfix image build script
    
    (cherry picked from commit ddb752fd55af62b9132f979f7964f28cbbb80522)
```

PR #962, opened 2023-05-12
Fix failure if repodata already exists

bdunne:
Example:
```
  Directory walk started
  Directory walk done - 22 packages
  Temporary output repo path: /tmp/rpms/.repodata/
  Preparing sqlite DBs
  Pool started (with 5 workers)
  Pool finished
  Critical: Cannot rename /tmp/rpms/.repodata/ -> /tmp/rpms/repodata/: Directory not empty
```

PR #963, opened 2023-05-12
Followup to #962

bdunne:
When running a hotfix build, the create_local_yum_repo script might not already have the rm -rf if the image was built before the change was merged


bdunne:
Backported to petrosian in commit 2f54e37f78a3625b04537a4fb5edd71b96d771dc

PR #968, opened 2023-06-02
Log to stderr by turning off the log file collector

jrafanie:
Fixes https://github.com/ManageIQ/container-postgresql/issues/40

Drop the no longer used options for logging to a file.

Related: https://github.com/CrunchyData/postgres-operator/issues/1141#issuecomment-629235277

Technically, it's an enhancement, but you can't look at pg logs in podified without visiting the pg pod so users would probably say this is a bug.

EDIT:  We saw some possibly no longer used/wrong settings so I opened an issue to start that discussion: https://github.com/ManageIQ/manageiq-pods/issues/969

### DIFF
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/postgresql_conf.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/postgresql_conf.go
@@ -38,8 +38,10 @@ hot_standby = on
 # ERROR REPORTING AND LOGGING
 #------------------------------------------------------------------------------
 
-log_filename = 'postgresql.log'
-log_rotation_age = 0
+log_destination = 'stderr'
+# This is used when logging to stderr:
+logging_collector = off
+
 log_min_duration_statement = 5000
 log_connections = on
 log_disconnections = on


manageiq-operator/api/v1alpha1/helpers/miq-components/postgresql_conf.go
```diff
@@ -38,8 +38,10 @@ hot_standby = on
 # ERROR REPORTING AND LOGGING
 #------------------------------------------------------------------------------
 
-log_filename = 'postgresql.log'
-log_rotation_age = 0
+log_destination = 'stderr'
```
jrafanie, line 41: Technically, this is the default... but I added `stderr` to ensure this is explicit.

manageiq-operator/api/v1alpha1/helpers/miq-components/postgresql_conf.go
```diff
@@ -38,8 +38,10 @@ hot_standby = on
...
+# This is used when logging to stderr:
+logging_collector = off
```
jrafanie, line 43: This is what actually turns off logging to files and logs to stderr.

manageiq-operator/api/v1alpha1/helpers/miq-components/postgresql_conf.go
```diff
@@ -38,8 +38,10 @@ hot_standby = on
...
+
 log_min_duration_statement = 5000
 log_connections = on
 log_disconnections = on
```
Fryguy, line 47: Should we drop these?  It seemed very chatty in the demo env.

jrafanie: Possibly?  But, not here... I opened https://github.com/ManageIQ/manageiq-pods/issues/969 to track the autovacuum settings but we can certainly review them all to see if we need those logged.

Fryguy (approved):

manageiq-operator/api/v1alpha1/helpers/miq-components/postgresql_conf.go
bdunne: Any reason not to use stdout?

Fryguy: That's not an option in the settings: https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-LOGGING-WHERE

Ultimately it's probably preferable to use jsonlog, but then I'm not sure you can do jsonlog _and_ do to stderr.

jrafanie: It's not an option (stderr, csvlog, jsonlog, and syslog) according to https://www.postgresql.org/docs/current/runtime-config-logging.html#RUNTIME-CONFIG-LOGGING-WHERE.  Perhaps it's to simplify things as `log_collector` "on" would need to redirect both stdout and stderr to logs.

```
log_collector

This parameter enables the logging collector, which is a background process that captures log messages sent to stderr and redirects them into log files
```

jrafanie: @Fryguy you beat me to it

Fryguy:
Backported to `petrosian` in commit 881c4fbe64b85e373c452038984e557a51b2756a.

```text
commit 881c4fbe64b85e373c452038984e557a51b2756a
Date:   Tue Jun 6 17:25:28 2023 -0400

    Merge pull request #968 from jrafanie/log_to_stderr
    
    Log to stderr by turning off the log file collector
    
    (cherry picked from commit be93f4155e4c2e09fbad85e10258cbbe94ee2860)
```

PR #974, opened 2023-06-09
Let sysv handle dynamic shared memory type for containers

jrafanie:
For posix, the default posix shared memory volume in openshift is 64 MB. We could solve this problem by mounting a larger volume and mounting that.

Or, we can just use sysv with a large enough shmall/shmmax, which we already have.

Since sysv is easier to implement, we're doing this here.

The details below show how to check the shared memory types, sizes, and possible solutions for both posix and sysv.

We originally had posix for dynamic shared memory:

```
postgres=# show shared_buffers;
 shared_buffers
----------------
 1GB
(1 row)

postgres=# show shared_memory_type;
 shared_memory_type
--------------------
 mmap
(1 row)

postgres=# show dynamic_shared_memory_type;
 dynamic_shared_memory_type
----------------------------
 posix
(1 row)
```

with /dev/shm of only the default: 64 MB:

```
sh-4.4$ df /dev/shm
Filesystem     1K-blocks  Used Available Use% Mounted on
shm                65536 10000     55536  16% /dev/shm
```

According to enterprisedb, you can solve this for each dynamic_shared_memory_type:

a) posix: by specifying a larger volume for "posix" (the default of 64 MB is too small)

   Add something like this:
```
        volumeMounts:
        - mountPath: /dev/shm
          name: shm
      volumes:
        - name: shm 
          emptyDir: 
            medium: Memory
            sizeLimit: 1Gi
```
b) sysv: or if your shmall/shmmax is large enough in the container, you can use "sysv" for your dynamic_shared_memory_type and you don't need to worry about the volume.

https://www.enterprisedb.com/docs/postgres_for_kubernetes/latest/postgresql_conf/ 
https://github.com/CrunchyData/postgres-operator/issues/2783 https://github.com/kubernetes/kubernetes/issues/28272 (posix shared memory was implemented in kubernetes here)

Since we had similarly enormous shmall values, we tried "sysv"

```
sh-4.4$ ipcs -lm

------ Shared Memory Limits --------
max number of segments = 4096
max seg size (kbytes) = 18014398509465599
max total shared memory (kbytes) = 18014398509481980
min seg size (bytes) = 1

sh-4.4$ cat /proc/sys/kernel/shmall
18446744073692774399
sh-4.4$ cat /proc/sys/kernel/shmmax
18446744073692774399
```

To do this manually on an existing podified installation, we edited the postgresql-configs ConfigMap.

We added a new file:

```
data:
  001_yolo_overrides.conf: >
    #------------------------------------------------------------------------------

    dynamic_shared_memory_type = sysv

    #------------------------------------------------------------------------------
  01_miq_overrides.conf: >
...
```

### DIFF
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/postgresql_conf.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/postgresql_conf.go
@@ -59,6 +59,19 @@ autovacuum_analyze_threshold = 500
 autovacuum_vacuum_scale_factor = 0.05
 
 #------------------------------------------------------------------------------
+# RESOURCE USAGE (except WAL)
+#------------------------------------------------------------------------------
+
+dynamic_shared_memory_type = sysv
+                                        # the default is the first option
+                                        # supported by the operating system:
+                                        #   posix
+                                        #   sysv
+                                        #   windows
+                                        #   mmap
+                                        # (change requires restart)
+
+#------------------------------------------------------------------------------
 # LOCK MANAGEMENT
 #------------------------------------------------------------------------------
 


Fryguy (approved):
LGTM - @bdunne Please also review.

jrafanie:
@Fryguy this is green and ready for merge/backport

Fryguy:
Backported to `petrosian` in commit 866fdfa38c0063bc5b95bfc7161ef384fbefca6f.

```text
commit 866fdfa38c0063bc5b95bfc7161ef384fbefca6f
Date:   Fri Jun 9 12:17:28 2023 -0400

    Merge pull request #974 from jrafanie/use_sysv_dynamic_shared_memory_type_for_containers
    
    Let sysv handle dynamic shared memory type for containers
    
    (cherry picked from commit 292eb40854dcfbc460c0b266d291ab7991a9547e)
```

Issue #988, opened 2023-09-07
OCP 4.14 support of heterogeneous cluster


kschanrtp:
In OCP 4.14, it is going to support heterogeneous cluster, that is, cluster can contain worker node that can run on either amd64 or ppc64le or s390x or arm64 platform. In order to allow OCP to select the correct worker node to run. The deployment for ManageIQ pods should contain nodeAffinity information. ManageIQ operator and ManageIQ orchestrator should provide the nodeAffinity information for the pods deployed in OCP environment.

Here is an example of affinity for supporting, amd64, ppc64le, and s390x platforms.
```
             spec:
               affinity:
                 nodeAffinity:
                   requiredDuringSchedulingIgnoredDuringExecution:
                     nodeSelectorTerms:
                     - matchExpressions:
                       - key: kubernetes.io/arch
                         operator: In
                         values:
                         - amd64
                         - ppc64le
                         - s390x


```

@miq-bot add-label enhancement

Fryguy:
@bdunne Please take a look.  We will need this for OCP 4.14

Fryguy:
I'm kind of surprised OCP can't "figure it out" on its own.  When it fetches the image, it knows, from the the docker manifest, which architectures the image can work on, and could determine it by itself.

Fryguy:
@bdunne This one might be challening, because technically we could support different sets of architectures depending on the build (thinking how upstream and downstream currently are different).

bdunne:
Fixed by #1003

PR #993, opened 2023-09-25
[WIP] Move util.go to miq-component/utils subpackage

nasark:
Currently if util functions are needed elsewhere, the entire `miq-components` package needs to be pulled. It would be nice to be able to pull in just the util methods, this is especially useful for the downstream operator

@miq-bot assign @Fryguy 
@miq-bot add_reviewer @bdunne 
@miq-bot add_labels enhancement, refactoring

PR #1003, opened 2023-10-17
Add Node affinity

bdunne:
For OpenShift 4.14 support

Part of #988

### DIFF
--- a/manageiq-operator/api/v1alpha1/miqutils/node_affinity.go
+++ b/manageiq-operator/api/v1alpha1/miqutils/node_affinity.go
@@ -1,0 +1,60 @@
+package miqutils
+
+import (
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"os"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func OperatorNodeAffinityArchValues(deployment *appsv1.Deployment, client client.Client) []string {
+	podName := os.Getenv("POD_NAME")
+	pod := FindPodByName(client, deployment.ObjectMeta.Namespace, podName)
+
+	if pod.Spec.Affinity == nil {
+		// In case we don't find the operator pod (local testing) or it doesn't have affinities
+		return []string{}
+	}
+	nodeSelectorTerms := pod.Spec.Affinity.NodeAffinity.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+
+	for _, selector := range nodeSelectorTerms {
+		for _, matchExpression := range selector.MatchExpressions {
+			if matchExpression.Key == "kubernetes.io/arch" {
+				return matchExpression.Values
+			}
+		}
+	}
+
+	// We should never get here, but the compiler requires it
+	return []string{}
+}
+
+func SetDeploymentNodeAffinity(deployment *appsv1.Deployment, client client.Client) {
+	operatorNodeAffinityArchValues := OperatorNodeAffinityArchValues(deployment, client)
+	if len(operatorNodeAffinityArchValues) == 0 {
+		// We're running local, can't find the operator pod, or it doesn't have any affinities to use as a template.  Skip it.
+		return
+	}
+
+	matchExpression := corev1.NodeSelectorRequirement{
+		Key:      "kubernetes.io/arch",
+		Operator: corev1.NodeSelectorOpIn,
+		Values:   operatorNodeAffinityArchValues,
+	}
+
+	matchExpressions := []corev1.NodeSelectorRequirement{matchExpression}
+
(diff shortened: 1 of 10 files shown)


bdunne:
@nasark Please also review.  The refactoring was inspired by your work in #993.  Do you think the `miqutils` code should continue to live in `api/v1alpha1` or should create a `api/v1` for it?

nasark:
> @nasark Please also review. The refactoring was inspired by your work in https://github.com/ManageIQ/manageiq-pods/pull/993. Do you think the miqutils code should continue to live in api/v1alpha1 or should create a api/v1 for it?

In my opinion since its technically just a refactor I would say to keep it in `v1alpha1`. Is there any benefit at the moment to creating a `v1` API? We would also need to support multiple versions in the CRD if thats the case

manageiq-operator/config/manager/manager.yaml
```diff
@@ -12,6 +12,15 @@ spec:
...
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: kubernetes.io/arch
+                operator: In
+                values:
+                - amd64
```
nasark, line 23: Do we also want `s390x` and `ppc64le` like in downstream?

nasark: In downstream we have a separate setNodeAffinity function that adds the other arches to the Kafka CR

bdunne:
> > @nasark Please also review. The refactoring was inspired by your work in #993. Do you think the miqutils code should continue to live in api/v1alpha1 or should create a api/v1 for it?
> 
> In my opinion since its technically just a refactor I would say to keep it in `v1alpha1`. Is there any benefit at the moment to creating a `v1` API? We would also need to support multiple versions in the CRD if thats the case

- I was thinking of `v1` just for the `miqutils`, not the CRD and everything else yet. (That's much more involved)
- v1 implies stable library, but maybe this isn't there yet.

manageiq-operator/config/manager/manager.yaml
bdunne: > Do we also want s390x and ppc64le like in downstream?

No, we only build x86_64 upstream

>  I'm not sure how this will work in downstream though if we have additional arches.

We don't use this file downstream.  The downstream operator will deploy the manageiq operator with the correct values.

manageiq-operator/api/v1alpha1/miqutils/node_affinity.go
```diff
@@ -0,0 +1,61 @@
...
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	"os"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func OperatorNodeAffinityArchValues(deployment *appsv1.Deployment, client client.Client) []string {
+	podName := os.Getenv("POD_NAME")
+	pod := FindPodByName(client, deployment.ObjectMeta.Namespace, podName)
+	values := []string{"amd64"}
```
bdunne, line 13: I added a commit to default to `amd64`

nasark (approved):
LGTM, just a note that Strimzi Kafka will require a different method of setting node affinity (similar to downstream). But that can be handled in https://github.com/ManageIQ/manageiq-pods/pull/1005 I suppose

Fryguy (approved):
LGTM, also want @nasark approval before merge.

Fryguy:
@bdunne How do we set the node affinities for the workloads? (workers and now also the automaton worker's jobs?)

bdunne:
> @bdunne How do we set the node affinities for the workloads? (workers and now also the automaton worker's jobs?)
> 
> I also updated the OP to say this is part of #988, but wasn't sure if it "completed" #988 or if we need other PRs (such as the node affinities for the workloads)

This will discover the node affinities applied to the operator and apply the same affinities to httpd, postgres, memcached and the orchestrator.  I'll be opening a PR shortly that does the same tihng inside the orchestrator for the deployments that it creates.

Fryguy:
@bdunne Can we merge this now or do we have to wait for the core orchestrator change?

Second question.  If we run on an older OpenShift, will these nodeAffinities cause issues, or are they just ignored on older OpenShifts

bdunne:
This can be merged before the orchestrator changes.

My test cluster is 4.13.4 and it's working fine there. I can't find when `kubernetes.io/arch` label was introduced, but I see references to it over 3 years ago.

Fryguy:
Skipping backport to `quinteros`, because it is already in the branch.

Fryguy:
Skipping backport to `quinteros`, because it is already in the branch.

PR #1005, opened 2023-10-23
Replace Bitnami with Strimzi Kafka

nasark:
- install Strimzi by creating a subscription from the `community-operators` OLM CatalogSource
- move `miq-components/kafka.go` to its own package `miq-components/kafka/kafka.go`, this will allow the CRs to be used elsewhere (i.e. downstream)

Depends on:
- [x] https://github.com/ManageIQ/manageiq-pods/pull/1008
- [ ] https://github.com/ManageIQ/manageiq-documentation/pull/1769
- [x] https://github.com/ManageIQ/manageiq-pods/pull/1020
- [x] https://github.com/ManageIQ/manageiq/pull/22773
- [x] https://github.com/ManageIQ/manageiq-pods/pull/896
- [x] https://github.com/ManageIQ/manageiq/pull/22775
- [x] https://github.com/ManageIQ/manageiq-pods/pull/1024

TODO:
- [x] Since Strimzi requires SSL, need to mount the `ca.crt` in orchestrator/workers regardless of whether `internal-certificate-secret` is present or not
- [x] set node affinity in Kafka CR
- [x] drop bitnami references in CR (marked as deprecated in CR)
- [x] topic management to be handled by orchestrator? (not needed for now)

@miq-bot assign @bdunne 
@miq-bot add_reviewer @Fryguy 
@miq-bot add_label enhancement

PR #1008, opened 2023-11-02
Add or update unique volume mount

nasark:
In kubernetes it is not possible to mount multiple volumes with the same mount path i.e. mount volume paths must be unique. Therefore added the edge case where if another volume is to be added that has the same mount path as an existing volume, the old volume is replaced

Needed for:
- https://github.com/ManageIQ/manageiq-pods/pull/1005

@miq-bot assign @bdunne


Fryguy:
Backported to `quinteros` in commit 8402f0188930d4330e439e6da3d21e9c29296979.

```text
commit 8402f0188930d4330e439e6da3d21e9c29296979
Date:   Mon Nov 6 09:56:51 2023 -0500

    Merge pull request #1008 from nasark/add_update_unique_vol_mount
    
    Add or update unique volume mount
    
    (cherry picked from commit 880f39ac19692166b101e93e63e90344e6ff6995)
```

PR #1041, opened 2024-01-26
Move messaging configuration to file and deprecate deployMessagingService flag

nasark:
- move messaging config from env vars to file
- the configs will be stored in secrets and mounted as a projected volume since the kafka user password is located in another secret created by strimzi
- kafka user secret naming convention normalized to match downstream `cr.Spec.AppName-user` -> `cr.Spec.AppName-kafka-admin` so that upstream and downstream can both use `func MessagingEnvSecret`
- moving the config to secrets allows us to check the hostname to see whether kafka resources should be deployed or not in the case of an external kafka. This allows us to fully deprecate `cr.Spec.DeployMessagingService`

Depends on:
- [ ] https://github.com/ManageIQ/manageiq/pull/22859

@miq-bot assign @bdunne
@miq-bot add_reviewer @Fryguy

### DIFF
--- a/images/manageiq-base/container-assets/container_env
+++ b/images/manageiq-base/container-assets/container_env
@@ -46,4 +46,35 @@ cat > ${APP_ROOT}/certs/v2_key << KEY
 :key: ${encryption_key}
 KEY
 
+
+[[ -f /run/secrets/messaging/MESSAGING_HOSTNAME ]] && messaging_hostname_file=$(cat /run/secrets/messaging/MESSAGING_HOSTNAME)
+[[ -f /run/secrets/messaging/MESSAGING_USERNAME ]] && messaging_username_file=$(cat /run/secrets/messaging/MESSAGING_USERNAME)
+[[ -f /run/secrets/messaging/MESSAGING_PASSWORD ]] && messaging_password_file=$(cat /run/secrets/messaging/MESSAGING_PASSWORD)
+[[ -f /run/secrets/messaging/MESSAGING_PORT ]] && messaging_port_file=$(cat /run/secrets/messaging/MESSAGING_PORT)
+[[ -f /run/secrets/messaging/MESSAGING_SASL_MECHANISM ]] && messaging_sasl_mechanism_file=$(cat /run/secrets/messaging/MESSAGING_SASL_MECHANISM)
+[[ -f /etc/pki/ca-trust/source/anchors/root.crt ]] && messaging_ca_path=/etc/pki/ca-trust/source/anchors/root.crt
+messaging_hostname=${MESSAGING_HOSTNAME:-$messaging_hostname_file}
+messaging_hostname=${messaging_hostname:-localhost}
+messaging_username=${MESSAGING_USERNAME:-$messaging_username_file}
+messaging_password=${MESSAGING_PASSWORD:-$messaging_password_file}
+messaging_port=${MESSAGING_PORT:-$messaging_port_file}
+messaging_port=${messaging_port:-9093}
+messaging_sasl_mechanism=${MESSAGING_SASL_MECHANISM:-$messaging_sasl_mechanism_file}
+messaging_ca_path=${messaging_ca_path:-/etc/pki/ca-trust/source/anchors/ca.crt}
+
+echo "== Writing messaging config =="
+cat > ${APP_ROOT}/config/messaging.yml << KEY
+---
+production:
+  host: ${messaging_hostname}
+  port: ${messaging_port}
+  protocol: Kafka
+  encoding: json
+  username: ${messaging_username}
+  password: ${messaging_password}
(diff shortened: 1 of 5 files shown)


nasark:

nasark:
@miq-bot add_label quinteros/yes?

manageiq-operator/internal/controller/manageiq_controller.go
```diff
@@ -526,6 +524,19 @@ func (r *ManageIQReconciler) generatePostgresqlResources(cr *miqv1alpha1.ManageI
...
 func (r *ManageIQReconciler) generateKafkaResources(cr *miqv1alpha1.ManageIQ) error {
+	secret, mutateFunc := miqkafka.MessagingEnvSecret(cr, r.Client, r.Scheme)
+	if result, err := controllerutil.CreateOrUpdate(context.TODO(), r.Client, secret, mutateFunc); err != nil {
+		return err
+	} else if result != controllerutil.OperationResultNone {
+		logger.Info("Secret has been reconciled", "component", "kafka", "result", result)
+	}
+
+	hostName := string(secret.Data["hostname"])
+	if hostName != "manageiq-kafka-bootstrap" {
```
nasark, line 535: I followed the same methodology we use for postgresql but in this case the hostname also depends on `cr.Spec.AppName`. Would this be an issue? For example if a user decides to change the default app name for some reason then this wouldn't work, however this allows us to not deploy upstream kafka in the downstream scenario

images/manageiq-base/container-assets/container_env
```diff
@@ -46,4 +46,35 @@ cat > ${APP_ROOT}/certs/v2_key << KEY
...
+[[ -f /run/secrets/messaging/MESSAGING_PORT ]] && messaging_port_file=$(cat /run/secrets/messaging/MESSAGING_PORT)
+[[ -f /run/secrets/messaging/MESSAGING_SASL_MECHANISM ]] && messaging_sasl_mechanism_file=$(cat /run/secrets/messaging/MESSAGING_SASL_MECHANISM)
+[[ -f /etc/pki/ca-trust/source/anchors/root.crt ]] && messaging_ca_path=/etc/pki/ca-trust/source/anchors/root.crt
+messaging_hostname=${MESSAGING_HOSTNAME:-$messaging_hostname_file}
+messaging_hostname=${messaging_hostname:-localhost}
+messaging_username=${MESSAGING_USERNAME:-$messaging_username_file}
+messaging_password=${MESSAGING_PASSWORD:-$messaging_password_file}
+messaging_port=${MESSAGING_PORT:-$messaging_port_file}
+messaging_port=${messaging_port:-9093}
+messaging_sasl_mechanism=${MESSAGING_SASL_MECHANISM:-$messaging_sasl_mechanism_file}
```
nasark, line 62: Since the env vars won't be used, should they be removed from these conditional assignments? I see that the database env vars are still checked in these conditional assignments https://github.com/ManageIQ/manageiq-pods/blob/master/images/manageiq-base/container-assets/container_env#L12-L20 so I followed suite but didn't understand the reasoning

Fryguy: We kept them more as a convenience for developers - if you run it locally, it's just a lot easier to pass them as `-e` on the command line

images/manageiq-base/container-assets/container_env
```diff
@@ -46,4 +46,35 @@ cat > ${APP_ROOT}/certs/v2_key << KEY
...
 
+
+[[ -f /run/secrets/messaging/MESSAGING_HOSTNAME ]] && messaging_hostname_file=$(cat /run/secrets/messaging/MESSAGING_HOSTNAME)
+[[ -f /run/secrets/messaging/MESSAGING_USERNAME ]] && messaging_username_file=$(cat /run/secrets/messaging/MESSAGING_USERNAME)
+[[ -f /run/secrets/messaging/MESSAGING_PASSWORD ]] && messaging_password_file=$(cat /run/secrets/messaging/MESSAGING_PASSWORD)
+[[ -f /run/secrets/messaging/MESSAGING_PORT ]] && messaging_port_file=$(cat /run/secrets/messaging/MESSAGING_PORT)
+[[ -f /run/secrets/messaging/MESSAGING_SASL_MECHANISM ]] && messaging_sasl_mechanism_file=$(cat /run/secrets/messaging/MESSAGING_SASL_MECHANISM)
+[[ -f /etc/pki/ca-trust/source/anchors/root.crt ]] && messaging_ca_path=/etc/pki/ca-trust/source/anchors/root.crt
+messaging_hostname=${MESSAGING_HOSTNAME:-$messaging_hostname_file}
+messaging_hostname=${messaging_hostname:-localhost}
```
Fryguy, line 57: Is this right?  I thought localhost didn't work?

bdunne: It should probably default to the name of the service used to reach kafka

PR #1052, opened 2024-02-06
Only return passwords that include letters.

bdunne:
Password decryption can fail if the database password is all numbers because ruby will read it as an integer instead of a string.

```
rake aborted!
NoMethodError: undefined method `empty?' for 8783875645714351:Integer 
.../gems/manageiq-password-1.2.0/lib/manageiq/password.rb:165:in `remove_erb' 
.../gems/manageiq-password-1.2.0/lib/manageiq/password.rb:90:in `try_decrypt' 
/var/www/miq/vmdb/lib/vmdb/settings_walker.rb:76:in `block in decrypt_passwords!'
```

### DIFF
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/generators.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/generators.go
@@ -4,6 +4,7 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"encoding/hex"
+	"regexp"
 )
 
 func randomBytes(n int) []byte {
@@ -20,6 +21,13 @@ func generateEncryptionKey() string {
 }
 
 func generatePassword() string {
-	buf := randomBytes(8)
-	return hex.EncodeToString(buf)
+	for {
+		buf := randomBytes(8)
+		password := hex.EncodeToString(buf)
+		if match, err := regexp.MatchString(`\D+`, password); err == nil && match {
+			// Only return if a letter is included.
+			// Password decryption can fail if the database password is all numbers because ruby will read it as an integer instead of a string.
+			return password
+		}
+	}
 }


manageiq-operator/api/v1alpha1/helpers/miq-components/generators.go
```diff
@@ -20,6 +21,13 @@ func generateEncryptionKey() string {
 }
 
 func generatePassword() string {
-	buf := randomBytes(8)
-	return hex.EncodeToString(buf)
+	for {
+		buf := randomBytes(8)
+		password := hex.EncodeToString(buf)
+		if match, err := regexp.MatchString(`\D+`, password); err == nil && match {
```
Fryguy, line 27: Will it return a blank password?  I was expecting it would try again in a loop?  A blank password might be worse if nothing is checking it.

Fryguy: Oh nevermind, I didn't see the `for` and forgot that's a free loop construct in go

bdunne: I don't think it can be blank, it's 8 random bytes, so 16 characters.

bdunne:
> This feels like it's bandaiding the underlying problem. Why is Ruby seeing it as an integer in the first place? I would have expected it to see a String. Makes me think we are not encoding into yaml somewhere properly (and that gets me worried about passwords starting with `!` for example)

The password is passed into the container via ENV or file mount [here](https://github.com/ManageIQ/manageiq-pods/blob/7bacf4f3bdbadb7bdcb337c948c769df2a564feb/images/manageiq-base/container-assets/container_env#L9) then written into database.yml.  So, in this case we saw:
```
---
production:
  adapter: postgresql
  username: root
  password: 8783875645714351
...
```
And when it was YAML loaded, password became an integer.

Fryguy:
Right, so that tells me we aren't writing the YAML correctly. The correct YAML should have been:

```
---
production:
  adapter: postgresql
  username: root
  password: "8783875645714351"
```

My concern is that there are other password characters that can cause similar issues, such as a leading `!`, `"`, or `|` character.  I realize we use hex now, so that can't happen in this case, but I also don't think we should be using hex moving forward, but instead something like base64 with a wider character set.

Fryguy:
Backported to `quinteros` in commit 743ce2ce751d476b3cdf3374749df21db3069312.

```text
commit 743ce2ce751d476b3cdf3374749df21db3069312
Date:   Thu Feb 8 10:35:25 2024 -0500

    Merge pull request #1052 from bdunne/database_password_integer
    
    Only return passwords that include letters.
    
    (cherry picked from commit 668e69924bd86df71c135251af091fe87c8d84ca)
```

PR #1069, opened 2024-03-13
Reorder DB container_env values for readability

Fryguy:
@bdunne Please review.  The old order was confusing me, and this seems easier to follow.


images/manageiq-base/container-assets/container_env
```diff
@@ -32,7 +37,7 @@ production:
   encoding: utf8
   pool: 5
   wait_timeout: 5
-  ssl_mode: ${DATABASE_SSL_MODE}
+  ssl_mode: ${database_ssl_mode}
 KEY
 
 [[ -f /.postgresql/root.crt ]] && echo "  sslrootcert=/.postgresql/root.crt" >> ${APP_ROOT}/config/database.yml
```
Fryguy, line 43: Completely unrelated, so I'll make a different PR, but I think this would be cleaner to append to the database_ssl_mode variable first and then interpolating it into the yml content as opposd to writing the yaml and then appending the file.

PR #1081, opened 2024-04-02
Add a CRD field to replace InsecureSkipVerify=true

bdunne:
~~Test image: docker.io/bdunne/manageiq-operator:ssl_verify~~
Add optional CR value `OIDCOAuthIntrospectionSSLVerify` to determine whether or not to verify SSL when fetching the introspection URL.  Default to SSLVerify for safety.


kbrock (approved):
LGTM

Are you thinking of passing a boolean parameter for passing in `InsecureSkipVerify:true` but defaulting to `false`?
Or are you thinking we can just drop it?

Fryguy:
I had the same q as @kbrock.  Do we need a setting for users that might need to set verify false?

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
```diff
@@ -659,9 +659,9 @@ func tlsSecretName(cr *miqv1alpha1.ManageIQ) string {
 	return secretName
 }
 
-func fetchIntrospectionUrl(providerUrl string) (string, error) {
+func fetchIntrospectionUrl(providerUrl string, sslVerify bool) (string, error) {
 	customTransport := http.DefaultTransport.(*http.Transport).Clone()
-	customTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	customTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: !sslVerify}
```
Fryguy, line 664: This is fine, but can we also just not set it?  Something like

```suggestion
	if !sslVerify {
		customTransport.TLSClientConfig = &tls.Config{InsecureSkipVerify: !sslVerify}
	}
```

Fryguy: Merging anyway.

Fryguy:
This PR has been effectively backported to `radjabov` via the merge of master into radjabov.

bdunne:
Unclean backport... Trying in https://github.com/ManageIQ/manageiq-pods/pull/1128 but tests aren't passing yet.

Fryguy:
@bdunne I moved this to morphy/yes? so it wouldn't get in the way of other backports.

PR #1093, opened 2024-04-25
update content-security-policy to match apache

kbrock:
see https://github.com/ManageIQ/manageiq-appliance/pull/383

This matches apache headers for rails, appliance, and pod


Fryguy:
See my comment in the appliance one, but I don't think we should do all of these.

kbrock:
We had agreed upon not matching these headers

PR #1101, opened 2024-05-08
Remove creation of manageiq.liveness-check topic

nasark:
We no longer use a push/pop approach as health check for kafka via the orchestrator and so this topic no longer needs to be created

Ref: https://github.com/ManageIQ/manageiq/pull/23020

@miq-bot assign @bdunne 
@miq-bot add_reviewers @bdunne, @Fryguy 
@miq-bot add_labels bug, radjabov/yes?

### DIFF
--- a/manageiq-operator/internal/controller/manageiq_controller.go
+++ b/manageiq-operator/internal/controller/manageiq_controller.go
@@ -572,7 +572,7 @@ func (r *ManageIQReconciler) generateKafkaResources(cr *miqv1alpha1.ManageIQ) er
 		logger.Info("Kafka User has been reconciled", "result", result)
 	}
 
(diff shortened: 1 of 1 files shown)


nasark:
@bdunne Added the CR migration, please review

manageiq-operator/api/v1alpha1/miqutils/find.go
```diff
@@ -59,6 +60,19 @@ func FindKafka(client client.Client, scheme *runtime.Scheme, namespace string, n
 	return kafka
 }
 
+func FindKafkaTopic(client client.Client, scheme *runtime.Scheme, namespace string, name string, group string) *unstructured.Unstructured {
```
nasark, line 63: Added `group` as a parameter so that this function can be reused downstream

PR #1106, opened 2024-05-17
Updated PodSelector label to opentofu-runner

sudhir-kelkar:
Updated network policy to consider updated label i.e. `service: opentofu-runner`


Fryguy (approved):
@bdunne Please review.

PR #1115, opened 2024-05-31
Generating token and putting int app-secrets

sudhir-kelkar:
Opentofu-runner need JWT token for service API authentication.
Creating this token using encryption key and setting it up in app-secrets

### DIFF
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/app-secret.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/app-secret.go
@@ -37,8 +37,11 @@ func ManageAppSecret(cr *miqv1alpha1.ManageIQ, client client.Client, scheme *run
 }
 
 func defaultAppSecret(cr *miqv1alpha1.ManageIQ) *corev1.Secret {
+	encryptionKey := generateEncryptionKey()
+	token, _ := generateJwtTokenToken(encryptionKey)
 	secretData := map[string]string{
-		"encryption-key": generateEncryptionKey(),
+		"encryption-key": encryptionKey,
+		"token":          token,
 	}
 
 	secret := &corev1.Secret{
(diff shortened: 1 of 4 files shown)


manageiq-operator/api/v1alpha1/helpers/miq-components/app-secret.go
```diff
@@ -37,8 +37,11 @@ func ManageAppSecret(cr *miqv1alpha1.ManageIQ, client client.Client, scheme *run
 }
 
 func defaultAppSecret(cr *miqv1alpha1.ManageIQ) *corev1.Secret {
+	encryptionKey := generateEncryptionKey()
+	token, _ := generateJwtTokenToken(encryptionKey)
 	secretData := map[string]string{
-		"encryption-key": generateEncryptionKey(),
+		"encryption-key": encryptionKey,
+		"token":          token,
```
agrare, line 44: Wondering should this be something like "opentofu-runner-token" since "token" is very general

sudhir-kelkar: Initially, I thought to name it opentofu-runner-token, but then I realized it is a JWT token that anybody can use. That's why I named it token. Although I couldn't think of any use case other than for OpenTofu-Runner, I can update it if necessary.

agrare: Yeah that's true, maybe the name should be more general then? https://github.com/ManageIQ/manageiq-pods/pull/1115/files#diff-9abe24d14f8b59d578f4a765fdb3c0e6dfaadb6d3d383555e550b1a11d01ea7fR39

sudhir-kelkar: Agree... I am changing it to `opentofu-runner-token. There is no other use case at this moment.

sudhir-kelkar: Changing it to `terraform-runner-token`. We are using it by that name in ManageIQ

bdunne (commented):
This is unexpected after our discussion last week.  I thought we were going to leverage the encryption key file mount that already exists on the client side and have the client connection generate the JWT.

agrare:
We're going with https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/56 instead

PR #1121, opened 2024-06-11
Fix bug in TLS secret detection

bdunne:
If someone were to successfully install the application then delete the TLSSecret, the operator would not recreate it because it detected that the Database Secret already existed.  This will cause errors during reconcile when the operator tries to update the CR status.  Instead, we should be checking the existence of the TLS Secret.


Fryguy:
Backported to `radjabov` in commit f1ec7d74193edfb22501d1967888c575e86556d4.

```text
commit f1ec7d74193edfb22501d1967888c575e86556d4
Date:   Tue Jun 11 11:50:02 2024 -0400

    Merge pull request #1121 from bdunne/tls_secret_error
    
    Fix bug in TLS secret detection
    
    (cherry picked from commit 6a9f9938ebe7752f2cda88243fec18398d4536d4)
```

PR #1139, opened 2024-07-31
Switch to ubuntu-24.04

bdunne:
`ubuntu-latest` still points to 22.04 which has podman 3.4.4 and run in that version does not support bind mounting volumes.  Upgrading to 24.04 gets us podman 4.9.3

Error:
```
error resolving mountpoints for container "4349440300a7f09982c00e0a8d5a010141b5eb912483ea6646bacaf54c080194": invalid mount type "bind"
```

### DIFF
--- a/.github/workflows/build_pods.yaml
+++ b/.github/workflows/build_pods.yaml
@@ -7,7 +7,7 @@ on:
 jobs:
   build_pods:
     if: github.repository_owner == 'ManageIQ'
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
     - uses: actions/checkout@v4
     - name: Set up Go


Fryguy:
LGTM - I wonder if there's a way we can know when latest bumps to this version.  I feel like we need something to trigger us to undo this change at some point or we're going to forget (maybe it's as simple as just opening an issue and letting it go stale)

Fryguy (approved):

Fryguy:
@bdunne Do we also need changes in https://github.com/ManageIQ/manageiq/blob/80ff67c710d7d081622a6b9b7bd0f70a2e929fca/.github/workflows/build_monolith.yaml#L10 ?

bdunne:
> @bdunne Do we also need changes in https://github.com/ManageIQ/manageiq/blob/80ff67c710d7d081622a6b9b7bd0f70a2e929fca/.github/workflows/build_monolith.yaml#L10 ?

I don't think so.  There's no bind mount there.

Fryguy:
Backported to radjabov via merge of master into radjabov branch

PR #1148, opened 2024-08-14
Update go to v1.23

bdunne:
Drop the toolchain and move minimum to 1.23

"If the toolchain line is omitted, the module or workspace is considered to have an implicit toolchain goV line, where V is the Go version from the go line." -https://go.dev/doc/toolchain


Fryguy (approved):

bdunne:
Also interesting that when using go 1.22 (`go 1.22` in go.mod) if you `go mod tidy` it will add the toolchain line for you, but it doesn't do that on 1.23.

PR #1163, opened 2024-09-04
Detect the latest version of dumb-init and use that

Fryguy:
This is what the output looks like

```
[2/2] STEP 10/16: RUN ARCH=$(uname -m) &&     URL=$(curl -s https://api.github.com/repos/Yelp/dumb-init/releases/latest | grep "browser_download_url.*_${ARCH}" | grep -o 'https://[^"]*') &&     echo $URL &&     curl -L -o /usr/bin/dumb-init $URL &&     chmod +x /usr/bin/dumb-init
https://github.com/Yelp/dumb-init/releases/download/v1.2.5/dumb-init_1.2.5_aarch64
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 71256  100 71256    0     0   233k      0 --:--:-- --:--:-- --:--:--  233k
```

### DIFF
--- a/images/manageiq-base/Dockerfile
+++ b/images/manageiq-base/Dockerfile
@@ -33,7 +33,9 @@ RUN chmod -R g+w /etc/pki/ca-trust && \
 
 # Install dumb-init to be used as the entrypoint
 RUN ARCH=$(uname -m) && \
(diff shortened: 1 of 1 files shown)


Fryguy:
I think we can treat it like we do other dependencies in that if there's a problem we can lock down (and then unlock when we figure out why it's broken). We could also potentially lock down harder on release branches if we think that's needed.

PR #1169, opened 2024-09-18
Watch routes and ingress, ensure we're only using one

bdunne:
- If routes are installed in the cluster, add a watch for them
- If we are using routes, ensure that we do not have an ingress.
- Remove manually set certificates on routes and use the cluster default

CP4AIOPS-6311

### DIFF
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
@@ -73,6 +73,10 @@ func Route(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme, client client.Clien
 		}
 
 		route.Spec.Host = cr.Spec.ApplicationDomain
+		// This removes the certificate that we previously set on the route and prevents anyone from setting their own.
+		// Removing the certificate on our route will cause it to use the cluster default certificate.
+		route.Spec.TLS.Certificate = ""
+		route.Spec.TLS.Key = ""
 
 		if internalCerts := InternalCertificatesSecret(cr, client); internalCerts.Data["httpd_crt"] != nil {
 			route.Spec.TLS.DestinationCACertificate = string(internalCerts.Data["root_crt"])
--- a/manageiq-operator/internal/controller/manageiq_controller.go
+++ b/manageiq-operator/internal/controller/manageiq_controller.go
@@ -479,6 +479,11 @@ func (r *ManageIQReconciler) generateHttpdResources(cr *miqv1alpha1.ManageIQ) er
 		} else if result != controllerutil.OperationResultNone {
 			logger.Info("Route has been reconciled", "component", "httpd", "result", result)
 		}
+
+		ingress := &networkingv1.Ingress{}
+		if err := r.Client.Get(context.TODO(), types.NamespacedName{Namespace: cr.Namespace, Name: "httpd"}, ingress); err == nil {
+			r.Client.Delete(context.TODO(), ingress)
+		}
 	} else {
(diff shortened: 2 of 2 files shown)


manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
```diff
@@ -73,6 +73,8 @@ func Route(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme, client client.Clien
 		}
 
 		route.Spec.Host = cr.Spec.ApplicationDomain
+		route.Spec.TLS.Certificate = ""
+		route.Spec.TLS.Key = ""
```
bdunne, line 77: I don't know if people would want to manually set certificates on routes, but this would remove them and force the use of the cluster certificates.

Fryguy: Can you comment as such right on the code?

kschanrtp: What happens if customer provides a custom tls cert secret in the IMInstall CR? Will custom cert still work?

bdunne: Done

bdunne: > What happens if customer provides a custom tls cert secret in the IMInstall CR? Will custom cert still work?

No, this would force using the certificates on the cluster.
Wouldn't they install that certificate on the cluster?

manageiq-operator/internal/controller/manageiq_controller.go
```diff
@@ -271,8 +273,21 @@ func (r *ManageIQReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+	if cfg, err := config.GetConfig(); err == nil {
+		if client, err := client.New(cfg, client.Options{Scheme: r.Scheme}); err == nil {
+			// Watch Routes if we are running in OpenShift
+			if err := client.List(context.TODO(), &routev1.RouteList{}); err == nil {
+				logger.Info("Adding watch for Routes!")
+				controller = controller.Owns(&routev1.Route{})
+			} else {
+				logger.Info(fmt.Sprintf("Skipping watch for Routes! %s", err))
+			}
```
Fryguy, line 286: Should this conditional also be on the other 3 `if` clauses if they fail?

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
kschanrtp: Not sure how this works. Probably. I am ok with the fix

kschanrtp (approved):
LGTM

manageiq-operator/internal/controller/manageiq_controller.go
bdunne: I don't think those would fail, but I added some anyway.

Fryguy (approved):

Fryguy:
Backported to `radjabov` in commit ee1772cd3270d6e65c9aa29c47b0474c12ac3784.

```text
commit ee1772cd3270d6e65c9aa29c47b0474c12ac3784
Date:   Fri Sep 20 19:48:59 2024 -0400

    Merge pull request #1169 from bdunne/watch_routes_ingress
    
    Watch routes and ingress, ensure we're only using one
    
    (cherry picked from commit fc227fa4a4b0e80ee67e37b7d1a6c03b49328e8b)
```

PR #1170, opened 2024-09-19
Replace EMS_IDS with EMS_ID in container entrypoint

agrare:
Workers having a comma separated list of EMS IDs is no longer used and this can be simplified to a single EMS_ID.

Co-depends:
* https://github.com/ManageIQ/manageiq/pull/23195


jrafanie (approved):
LGTM

jrafanie (approved):

PR #1173, opened 2024-10-01
Prefer UBI and go download over library/golang

bdunne:
- speed up build since there is less to download
- reduce external dependencies
- build and run on the same base image

```
REPOSITORY                TAG  IMAGE ID     CREATED         SIZE
my                        test 422e3c099772 10 seconds ago  366 MB
docker.io/library/golang  1.23 361d8b5c9aa5 3 weeks ago     862 MB
```

### DIFF
--- a/manageiq-operator/Dockerfile
+++ b/manageiq-operator/Dockerfile
@@ -4,12 +4,9 @@ FROM registry.access.redhat.com/ubi9/ubi-minimal as builder
 COPY go.mod /go.mod
 
 RUN microdnf -y install gzip jq tar && \
-    arch=$(uname -m) && \
-    if [ ${arch} == "x86_64" ] ; then \
-      arch="amd64" \
-    ; fi && \
+    arch=$(uname -m | sed 's/x86_64/amd64/') && \
     goversion=$(sed -nr 's/^go\s+(.*)/go\1/p' /go.mod) && \
-    file=$(curl -s 'https://go.dev/dl/?mode=json&include=all' | jq -r --arg GOVERSION "${goversion}" --arg ARCH "${arch}" 'map(select(.version | startswith($GOVERSION)))[0].files[] | select(.arch == $ARCH and .os == "linux").filename') && \
+    file=$(curl -s 'https://go.dev/dl/?mode=json&include=all' | jq -r --arg GOVERSION "${goversion}" --arg ARCH "${arch}" 'first(.[] | select(.version | startswith($GOVERSION))).files[] | select(.arch == $ARCH and .os == "linux").filename') && \
     curl -s -f -L https://go.dev/dl/${file} | tar -C / -xzf -
 
 ENV PATH=/go/bin:/root/.local/bin:/root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
--- a/manageiq-operator/Dockerfile
+++ b/manageiq-operator/Dockerfile
@@ -1,8 +1,16 @@
 # Build the manager binary
 FROM registry.access.redhat.com/ubi9/ubi-minimal as builder
 
-RUN microdnf -y install gzip tar && \
-    curl -f -L https://go.dev/dl/go1.23.1.linux-amd64.tar.gz | tar -C / -xzf -
+COPY go.mod /go.mod
+
+RUN microdnf -y install gzip jq tar && \
+    arch=$(uname -m) && \
+    if [ ${arch} == "x86_64" ] ; then \
+      arch="amd64" \
+    ; fi && \
+    goversion=$(sed -nr 's/^go\s+(.*)/go\1/p' /go.mod) && \
+    file=$(curl -s 'https://go.dev/dl/?mode=json&include=all' | jq -r --arg GOVERSION "${goversion}" --arg ARCH "${arch}" 'map(select(.version | startswith($GOVERSION)))[0].files[] | select(.arch == $ARCH and .os == "linux").filename') && \
+    curl -s -f -L https://go.dev/dl/${file} | tar -C / -xzf -
 
 ENV PATH=/go/bin:/root/.local/bin:/root/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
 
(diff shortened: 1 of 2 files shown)


manageiq-operator/Dockerfile
```diff
@@ -1,5 +1,10 @@
 # Build the manager binary
-FROM docker.io/library/golang:1.23 as builder
+FROM registry.access.redhat.com/ubi9/ubi-minimal as builder
+
+RUN microdnf -y install gzip tar && \
+    curl -f -L https://go.dev/dl/go1.23.1.linux-amd64.tar.gz | tar -C / -xzf -
```
Fryguy, line 5: This has to be made arch aware for multi-arch builds right?

Fryguy: The version number is gonna be a pain.  For example 1.23.2 is already out so this PR is out of date.

bdunne: Done

manageiq-operator/Dockerfile
```diff
@@ -1,5 +1,18 @@
 # Build the manager binary
-FROM docker.io/library/golang:1.23 as builder
+FROM registry.access.redhat.com/ubi9/ubi-minimal as builder
+
+COPY go.mod /go.mod
```
bdunne, line 4: This is used to detect the desired Go major.minor version, but could be changed to an ENV var if we want to extract this to a script for reuse.

manageiq-operator/Dockerfile
```diff
@@ -1,5 +1,18 @@
...
+
+RUN microdnf -y install gzip jq tar && \
+    arch=$(uname -m) && \
```
kschanrtp, line 7: I cannot find the download link https://go.dev/dl/ if the platform is
ppc64le 
s390x

Wonder if you can just use amd64 instead for both power and s390x.

manageiq-operator/Dockerfile
kschanrtp, line 4: Do we need  go.sum?

manageiq-operator/Dockerfile
Fryguy: https://go.dev/dl/go1.23.2.linux-ppc64le.tar.gz
https://go.dev/dl/go1.23.2.linux-s390x.tar.gz

manageiq-operator/Dockerfile
```diff
@@ -1,5 +1,18 @@
...
+    if [ ${arch} == "x86_64" ] ; then \
+      arch="amd64" \
+    ; fi && \
```
Fryguy, line 10: You can inline this and I think it's will look nicer

```suggestion
    if [ ${arch} == "x86_64" ] ; then arch="amd64"; fi && \
```

manageiq-operator/Dockerfile
```diff
@@ -1,5 +1,18 @@
...
+    goversion=$(sed -nr 's/^go\s+(.*)/go\1/p' /go.mod) && \
+    file=$(curl -s 'https://go.dev/dl/?mode=json&include=all' | jq -r --arg GOVERSION "${goversion}" --arg ARCH "${arch}" 'map(select(.version | startswith($GOVERSION)))[0].files[] | select(.arch == $ARCH and .os == "linux").filename') && \
```
Fryguy, line 12: Curious why you used map[0] over first? (Same in the end, but I figured first was clearer)

manageiq-operator/Dockerfile
kschanrtp: Forgot to expand the tab. Looks good. Thanks

Fryguy:
I am curious what's in the golang image that causes it to be so much larger

manageiq-operator/Dockerfile
bdunne: No, I was just reading the `go` line from the file to figure out which version we should download.

kschanrtp: Good. Thanks.

manageiq-operator/Dockerfile
bdunne: `first` doesn't appear to do anything here...
`curl -s 'https://go.dev/dl/?mode=json&include=all' | jq -r 'map(select(.version | startswith("go1.23")))'`
and
`curl -s 'https://go.dev/dl/?mode=json&include=all' | jq -r 'first(map(select(.version | startswith("go1.23"))))'`
both return an array.

```
$ curl -s 'https://go.dev/dl/?mode=json&include=all' | jq -r 'first(map(select(.version | startswith("go1.23"))))'
[
  {
    "version": "go1.23.2",
    "stable": true,
    "files": [
      {
        "filename": "go1.23.2.src.tar.gz",
        "os": "",
        "arch": "",
        "version": "go1.23.2",
        "sha256": "36930162a93df417d90bd22c6e14daff4705baac2b02418edda671cdfa9cd07f",
        "size": 28171276,
        "kind": "source"
      },
.....
```

Fryguy: Oh it's not first/map, it's

```
first(.[] | select(.version | startswith("go1.23.")))
```

bdunne: Done

manageiq-operator/Dockerfile
```diff
@@ -1,5 +1,16 @@
 # Build the manager binary
-FROM docker.io/library/golang:1.23 as builder
+FROM registry.access.redhat.com/ubi9/ubi-minimal as builder
+
+COPY go.mod /go.mod
+
+RUN microdnf -y install gzip jq tar && \
+    arch=$(uname -m) && \
+    if [ ${arch} == "x86_64" ]; then arch="amd64"; fi && \
```
Fryguy, line 8: Not saying to do this, but we can trim up this line into a single line with

```suggestion
    arch=$(uname -m | sed 's/x86_64/amd64/') && \
```

Thoughts?

Fryguy (approved):

manageiq-operator/Dockerfile
bdunne: Less is less ✂️ ✂️ ✂️

Fryguy (approved):

Fryguy:
Backported to `radjabov` in commit 0b3b3aaa74fea39dc1f7bd1ebf87964be410605a.

```text
commit 0b3b3aaa74fea39dc1f7bd1ebf87964be410605a
Date:   Mon Oct 14 18:00:32 2024 -0400

    Merge pull request #1173 from bdunne/ubi_golang
    
    Prefer UBI and go download over library/golang
    
    (cherry picked from commit 07b676821120b374b9d1e3d450b36a54b9a0802f)
```

PR #1192, opened 2025-01-07
[morphy] Upgrade protobuf to resolve CVE-2024-24786

bdunne:
CP4MCM-142


manageiq-operator/go.sum
```diff
@@ -1140,8 +1141,9 @@ google.golang.org/protobuf v0.0.0-20200228230310-ab0ca4ff8a60/go.mod h1:cfTl7dwQ
 google.golang.org/protobuf v1.20.1-0.20200309200217-e05f789c0967/go.mod h1:A+miEFZTKqfCUM6K7xSMQL9OKL/b6hQv+e19PK+JZNE=
 google.golang.org/protobuf v1.21.0/go.mod h1:47Nbq4nVaFHyn7ilMalzfO3qCViNmqZ2kzikPIcrTAo=
 google.golang.org/protobuf v1.23.0/go.mod h1:EGpADcykh3NcUnDUJcl1+ZksZNG86OlYog2l/sGQquU=
-google.golang.org/protobuf v1.26.0-rc.1 h1:7QnIQpGRHE5RnLKnESfDoxm2dTapTZua5a0kS0A+VXQ=
 google.golang.org/protobuf v1.26.0-rc.1/go.mod h1:jlhhOSvTdKEhbULTjvd4ARK9grFBp09yW+WbY/TyQbw=
```
bdunne, line 1144: I'm not sure if this PR will resolve the issue since this line is still here.  I don't really understand the dependency graph here, but it seems to be:

something requires `google.golang.org/protobuf@v1.36.2` which requires `github.com/golang/protobuf@v1.5.0`
and `github.com/golang/protobuf@v1.5.0` requires `google.golang.org/protobuf@v1.26.0-rc.1`

Also... `go: module github.com/golang/protobuf is deprecated: Use the "google.golang.org/protobuf" module instead.`

bdunne: If necessary, we may be able to do a module replace to remove it.

PR #1213, opened 2025-02-21
Fix issue with a shared reference for the deploymentLabels

Fryguy:
This fix handles an issue where the spec.selector.matchLabels and the spec.template.metadata.labels contained a reference to the same map. Some labels only are supposed to appear in the template labels and not the selector, such as the `manageiq.org/backup` label and the `ssl-certificate-resource-version` label. These labels were accidentally being copied into the spec selector as well.

In Kubernetes, template labels are modifiable while selector labels are immutable. Another rule is that template labels must match the selector labels, but it's a subset (i.e. you can have extra template labels). When the extra labels are "shared" into the spec selector, they will initially not exhibit any problems because they match, but as soon as the value is changed in the template labels, such as updating the `ssl-certificate-resource-version` to a new version on a cert update, then you get an error like the following:

```
Deployment.apps "postgresql" is invalid: spec.template.metadata.labels: Invalid value: map[string]string
{"app":"manageiq-application", "manageiq.org/backup":"t", "name":"postgresql", "ssl-certificate-resource-version":"12271329"}
: `selector` does not match template `labels`
```

@bdunne Please review.
cc @jrafanie @kschanrtp 

Note that I didn't directly test this yet, cause I don't know how LOL 😅


manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
```diff
@@ -17,7 +19,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	intstr "k8s.io/apimachinery/pkg/util/intstr"
-	"net/http"
```
jrafanie, line 20: You just can't leave it alone.  It has to be sorted by proper scope/context.  🤣

manageiq-operator/api/v1alpha1/helpers/miq-components/orchestrator.go
```diff
@@ -12,8 +16,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"strconv"
-	"strings"
```
jrafanie, line 16: 🤣 Sort-all-the-things!

Fryguy: I think gofmt did this for me 😅

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
Fryguy: It wasn't me haha

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
```diff
@@ -414,6 +415,8 @@ func HttpdDeployment(client client.Client, cr *miqv1alpha1.ManageIQ, scheme *run
 		"app":  cr.Spec.AppName,
 		"name": "httpd",
 	}
+	deploymentSelectorLabels := map[string]string{}
+	maps.Copy(deploymentLabels, deploymentSelectorLabels)
```
jrafanie, line 419: Is this only on the initial deployment? A function name like `NewMemcachedDeployment` imply only the initial one but wasn't sure if the copying is ever happening after the initial deployment creation.

Fryguy: So the operator sdk is weird. This method sets up a bunch of objects and then it returns a mutation function that is run every time a reconcile needs to happen. As such, this code outside of the function is in a closure bound to that function. For all intents and purposes it's like a global object and a function that manipulates the global object. It's super weird.

PR #1215, opened 2025-02-25
Update the maps.Copy operands.

kschanrtp:
The maps.Copy function syntax is maps.Copy(dest, src). Currently we have the wrong order. Fixing it.
https://pkg.go.dev/maps#Copy


jrafanie:
Fixes a bug from #1213

jrafanie (approved):

PR #1252, opened 2025-05-19
Only pick the major and minor go version to ensure the latest is pulled

bdunne:
Handle cases where go.mod contains "go 1.24.0" or "go 1.24" and pick the latest patch version for that major.minor.

Previously this was picking up the whole of "go1.24.0" and locking us down to that minimum version


Fryguy (approved):

Fryguy:
Backported to spassky via merge of master into spassky

PR #1264, opened 2025-06-30
Run with the latest go version

bdunne:
Unfortunately it looks like we need to specify the z-version of go now... this will keep it up to date.
https://github.com/ManageIQ/manageiq-pods/actions/runs/15949770344/job/45082004161

### DIFF
--- a/.github/workflows/build_pods.yaml
+++ b/.github/workflows/build_pods.yaml
@@ -15,7 +15,7 @@ jobs:
       with:
         cache-dependency-path: 'manageiq-operator/go.sum'
         check-latest: true
-        go-version-file: 'manageiq-operator/go.mod'
+        go-version: 'stable'
     - name: Docker login
       run: echo ${{ secrets.DOCKER_REGISTRY_PASSWORD }} | docker login docker.io --password-stdin --username ${{ secrets.DOCKER_REGISTRY_USERNAME }}
     - name: Build pods containers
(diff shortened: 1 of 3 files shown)


Fryguy (dismissed):

Fryguy:
Maybe we can open an issue/feature request on setup-go - there's gotta be a way to do it through there

Fryguy:
@bdunne I have a thought - Can you manually change the go.mod to say 1.24 instead of 1.24.0 in this PR?  I wonder if that .0 is causing it to think that's the version we want.

Fryguy:
@bdunne Nice - that actually set the version to 1.24.4 https://github.com/ManageIQ/manageiq-pods/actions/runs/16053488241/job/45301834742?pr=1264#step:5:15

Fryguy:
I noticed go mod tidy has a `-go` flag and a `-compat` flag - Not sure if those help in maintaining the go version or not.

bdunne:
This is all so frustrating...

If I drop the `.0` patch version in go.mod, `go mod tidy` will create the following diff which causes tests to fail.
```diff
 module github.com/ManageIQ/manageiq-pods/manageiq-operator
 
-go 1.24
+go 1.24.0
+
+toolchain go1.24.4
 
 require (
```

bdunne:
I'm not sure what will happen when 1.25.0 is released, but this is doing what we want right now...

Fryguy:
> which causes tests to fail.

What tests are failing?  Is it just our internal Ruby tests, or actual go tests failing?

PR #1378, opened 2026-05-07
Initial bootc image

bdunne:
Part of: https://github.com/ManageIQ/manageiq/issues/23737
- [x] Containerfiles
- [x] bin/build
- [x] config.toml(s)
- [x] disk image build added to bin/build
- [x] Full end-to-end test

Helpful links:
- https://github.com/osbuild/bootc-image-builder/
- https://osbuild.org/docs/user-guide/partitioning/
- https://osbuild.org/docs/user-guide/blueprint-reference/#additional-users
- https://github.com/redhat-cop/redhat-image-mode-actions/blob/main/.github/workflows/build_rhel_bootc.yml
- https://gitlab.com/redhat/cop/rhel/rhel-image-mode-cicd
- https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/using_image_mode_for_rhel_to_build_deploy_and_manage_operating_systems/creating-bootc-compatible-base-disk-images-with-bootc-image-builder_using-image-mode-for-rhel-to-build-deploy-and-manage-operating-systems#creating-vmdk-images-by-using-bootc-image-builder_creating-bootc-compatible-base-disk-images-with-bootc-image-builder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,5 @@ manageiq-operator/build/_output/*
 manageiq-operator/vendor/*
 tools/certs/*.crt
 tools/certs/*.key
+*.qcow2
 *.swp
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ images/manageiq-base-worker/container-assets/VMware-*
 images/manageiq-rpms/rpms
 manageiq-operator/build/_output/*
 manageiq-operator/vendor/*
+output/*
 tools/certs/*.crt
 tools/certs/*.key
 *.qcow2
--- a/bin/build
+++ b/bin/build
@@ -85,6 +85,17 @@ if [ -z "$NO_OPERATOR" ]; then
   popd
 fi
 
+if APPLIANCES; then
+  $CONTAINER_COMMAND pull quay.io/centos-bootc/centos-bootc:stream10
+  cd images/manageiq-appliance
+  podman build . -t $REPO/manageiq-appliance:$TAG
+  for i in $(ls Containerfile.*); do
+    extension=${i##*.}
+    echo $extension
+    podman build . -t $REPO/manageiq-appliance:$TAG-$extension
+  done
+fi
+
 if [ -n "$PUSH" ]; then
   push_images="manageiq-base manageiq-base-worker manageiq-orchestrator manageiq-webserver-worker manageiq-ui-worker"
   if [ -z "$NO_OPERATOR" ]; then
--- a/bin/build
+++ b/bin/build
@@ -30,6 +30,11 @@ REPO=${REPO:-docker.io/manageiq}
 
 CONTAINER_COMMAND="$(which podman &>/dev/null && echo "podman" || echo "docker")"
 
+if [[ -z "$NO_APPLIANCES" && -n "$PUSH" && -z "$IBM_CLOUD_API_KEY" ]]; then
+  echo "ERROR: Missing IBM_CLOUD_API_KEY"
+  exit 1
+fi
+
 set -e
 
 if [ -n "$REBUILD_RPM" ]; then
@@ -62,7 +67,7 @@ if [ -n "$RELEASE_BUILD" -a "$image" == "manageiq-base" ]; then
 fi
 
 pushd $IMAGE_DIR
-  # Pull the base image first
+  echo "Pulling base UBI image"
   $CONTAINER_COMMAND pull registry.access.redhat.com/ubi10/ubi
 
   build_images="manageiq-rpms manageiq-base manageiq-base-worker manageiq-orchestrator manageiq-webserver-worker manageiq-ui-worker"
@@ -84,28 +89,28 @@ fi
 
 appliance_images_to_push=""
 if [ -z "$NO_APPLIANCES" ]; then
+  echo "Pulling base appliance bootc image"
   $CONTAINER_COMMAND pull quay.io/centos-bootc/centos-bootc:stream10
   pushd images/manageiq-appliance
 
-  target_tag="$REPO/manageiq-appliance:$TAG"
-  cmd="$CONTAINER_BUILD_COMMAND . --tag $target_tag"
+  cmd="$CONTAINER_BUILD_COMMAND . --tag $REPO/manageiq-appliance-base:$TAG"
   echo "Building $image: $cmd"
   $cmd
 
   for containerfile in $(ls Containerfile.*); do
     target=${containerfile##*.}
 
-    target_tag="$REPO/manageiq-appliance:$TAG-$target"
-    cmd="$CONTAINER_BUILD_COMMAND . -f $containerfile --tag $target_tag"
+    cmd="$CONTAINER_BUILD_COMMAND . -f $containerfile --tag $REPO/manageiq-appliance-$target:$TAG"
     echo "Building $image: $cmd"
     $cmd
-    appliance_images_to_push+="$target_tag "
+    appliance_images_to_push+="manageiq-appliance-$target "
   done
   popd
 fi
 
 if [ -n "$PUSH" ]; then
   push_images="manageiq-base manageiq-base-worker manageiq-orchestrator manageiq-webserver-worker manageiq-ui-worker"
+  push_images+=" $appliance_images_to_push"
   if [ -z "$NO_OPERATOR" ]; then
     push_images="$push_images manageiq-operator"
   fi
@@ -121,18 +126,73 @@ if [ -n "$PUSH" ]; then
     echo "Pushing $image: $cmd"
     $cmd
   done
+fi
 
-  if [ -z "$NO_APPLIANCES" ]; then
-    for image in $appliance_images_to_push; do
-      cmd="$CONTAINER_COMMAND push $image"
+# Build the disk images
+if [ -z "$NO_APPLIANCES" ]; then
+  echo "Pulling bootc image builder image"
+  sudo podman pull quay.io/centos-bootc/bootc-image-builder:latest
 
-      # --format docker is needed for podman to ensure the builds are built with docker v2 manifests
-      if [ "$CONTAINER_COMMAND" == "podman" ]; then
-        cmd+=" --format docker"
+  for appliance_image in $appliance_images_to_push; do
+    echo "Building Appliance Disk for: $appliance_image"
+    target_platform="${appliance_image##*-}"
+    case "$target_platform" in
+      azure|ec2|hyperv|vagrant)
+        disk_type="vhd"
+        ;;
+      gce)
+        disk_type="gce"
+        ;;
+      vsphere)
+        disk_type="vmdk"
+        ;;
+      *)
+        # libvirt, openstack, ovirt
+        disk_type="qcow2"
+        ;;
+    esac
+
+    echo "Pulling appliance image: $REPO/$appliance_image:$TAG"
+    sudo podman pull "$REPO/$appliance_image:$TAG"
+
+    echo "Building appliance disk image"
+    sudo podman run \
+        --rm \
+        -it \
+        --privileged \
+        --security-opt label=type:unconfined_t \
+        -v ./images/manageiq-appliance/config.toml.$target_platform:/config.toml:ro \
+        -v ./output:/output \
+        -v /var/lib/containers/storage:/var/lib/containers/storage \
+        quay.io/centos-bootc/bootc-image-builder:latest \
+        --type $disk_type \
+        --use-librepo=True \
+        --chown $(id -u):$(id -g) \
+        "$REPO/$appliance_image:$TAG"
+
+    if [ -n "$PUSH" ]; then
+      if [ "$disk_type" == "vhd" ]; then
+        pushd output/vpc/
+          echo "Zipping VHD..."
+          zip -m -j disk.zip disk.vhd
+        popd
       fi
 
-      echo "Pushing $image: $cmd"
-      $cmd
-    done
-  fi
+      disk_image=$(find ./output/ -type f ! -name "*.json" ! -name ".gitkeep")
+      destination_name="$RPM_PREFIX-$target_platform-$TAG.$disk_type"
+
+      # Obtain bearer token
+      echo "Obtaining bearer token..."
+      response=$(curl --fail -s -X "POST" "https://iam.cloud.ibm.com/oidc/token" -H 'Accept: application/json' -H 'Content-Type: application/x-www-form-urlencoded' --data-urlencode "apikey=$IBM_CLOUD_API_KEY" --data-urlencode "response_type=cloud_iam" --data-urlencode "grant_type=urn:ibm:params:oauth:grant-type:apikey")
+      access_token=$(echo $response | jq -r ".access_token")
+
+      # Upload
+      echo
+      echo "Uploading..."
+      bucket="releases-manageiq-org"
+      curl -X "PUT" -H "Authorization: Bearer $access_token" -H "Content-Type: application/octet-stream" -H "x-amz-acl: public-read" -T "$disk_image" "https://s3.us-east.cloud-object-storage.appdomain.cloud/releases-manageiq-org/$destination_name"
+
+      rm -rf output/*
+    fi
+  done
 fi
--- a/bin/build
+++ b/bin/build
@@ -6,8 +6,9 @@ RPM_BUILD_OPTIONS=${RPM_BUILD_OPTIONS:-""}
 RPM_BUILD_IMAGE=${RPM_BUILD_IMAGE:-"manageiq/rpm_build:$TAG"}
 RPM_PREFIX=${RPM_PREFIX:-"manageiq"}
 
-while getopts "t:c:d:r:hbnops" opt; do
+while getopts "t:c:d:r:habnops" opt; do
   case $opt in
+    a) NO_APPLIANCES="true" ;;
     b) REBUILD_RPM="true" ;;
     c) CONTAINER_PREFIX=$OPTARG ;;
     d) BASE_DIR=$OPTARG ;;
@@ -17,7 +18,7 @@ while getopts "t:c:d:r:hbnops" opt; do
     r) REPO=$OPTARG ;;
     s) RELEASE_BUILD="true" ;;
     t) TAG=$OPTARG ;;
-    h) echo "Usage: $0 [-hbnops] [-c CONTAINER_PREFIX] [-d BASE_DIR] [-r IMAGE_REPOSITORY] [-t IMAGE_TAG]"; exit 1
+    h) echo "Usage: $0 [-habnops] [-c CONTAINER_PREFIX] [-d BASE_DIR] [-r IMAGE_REPOSITORY] [-t IMAGE_TAG]"; exit 1
   esac
 done
 
@@ -45,32 +46,28 @@ if [ -n "$REBUILD_RPM" ]; then
   $cmd
 fi
 
+CONTAINER_BUILD_COMMAND="$CONTAINER_COMMAND build --build-arg FROM_REPO=$REPO --build-arg FROM_TAG=$TAG --build-arg RPM_PREFIX=$RPM_PREFIX"
+if [ -n "$VERBOSE" ]; then
+  CONTAINER_BUILD_COMMAND+=" --progress plain"
+fi
+# --format docker is needed for podman to ensure the builds are built with docker v2 manifests
+if [ "$CONTAINER_COMMAND" == "podman" ]; then
+  CONTAINER_BUILD_COMMAND+=" --format docker"
+fi
+if [ -n "$NO_CACHE" ]; then
+  CONTAINER_BUILD_COMMAND+=" --no-cache"
+fi
+if [ -n "$RELEASE_BUILD" -a "$image" == "manageiq-base" ]; then
+  CONTAINER_BUILD_COMMAND+=" --build-arg RELEASE_BUILD=true"
+fi
+
 pushd $IMAGE_DIR
   # Pull the base image first
   $CONTAINER_COMMAND pull registry.access.redhat.com/ubi10/ubi
 
   build_images="manageiq-rpms manageiq-base manageiq-base-worker manageiq-orchestrator manageiq-webserver-worker manageiq-ui-worker"
   for image in $build_images; do
-    cmd="$CONTAINER_COMMAND build --build-arg FROM_REPO=$REPO --build-arg FROM_TAG=$TAG --build-arg RPM_PREFIX=$RPM_PREFIX --tag $REPO/$image:$TAG"
-
-    if [ -n "$VERBOSE" ]; then
-      cmd+=" --progress plain"
-    fi
-
-    # --format docker is needed for podman to ensure the builds are built with docker v2 manifests
-    if [ "$CONTAINER_COMMAND" == "podman" ]; then
-      cmd+=" --format docker"
-    fi
-
-    if [ -n "$NO_CACHE" ]; then
-      cmd+=" --no-cache"
-    fi
-
-    if [ -n "$RELEASE_BUILD" -a "$image" == "manageiq-base" ]; then
-      cmd+=" --build-arg RELEASE_BUILD=true"
-    fi
-
-    cmd+=" $image"
+    cmd="$CONTAINER_BUILD_COMMAND --tag $REPO/$image:$TAG $image"
 
     echo "Building $image: $cmd"
     $cmd
@@ -85,15 +82,26 @@ if [ -z "$NO_OPERATOR" ]; then
   popd
 fi
 
-if APPLIANCES; then
+appliance_images_to_push=""
+if [ -z "$NO_APPLIANCES" ]; then
   $CONTAINER_COMMAND pull quay.io/centos-bootc/centos-bootc:stream10
-  cd images/manageiq-appliance
-  podman build . -t $REPO/manageiq-appliance:$TAG
-  for i in $(ls Containerfile.*); do
-    extension=${i##*.}
-    echo $extension
-    podman build . -t $REPO/manageiq-appliance:$TAG-$extension
+  pushd images/manageiq-appliance
+
+  target_tag="$REPO/manageiq-appliance:$TAG"
+  cmd="$CONTAINER_BUILD_COMMAND . --tag $target_tag"
+  echo "Building $image: $cmd"
+  $cmd
+
+  for containerfile in $(ls Containerfile.*); do
+    target=${containerfile##*.}
+
+    target_tag="$REPO/manageiq-appliance:$TAG-$target"
+    cmd="$CONTAINER_BUILD_COMMAND . -f $containerfile --tag $target_tag"
+    echo "Building $image: $cmd"
+    $cmd
+    appliance_images_to_push+="$target_tag "
   done
+  popd
 fi
 
 if [ -n "$PUSH" ]; then
@@ -113,4 +121,18 @@ if [ -n "$PUSH" ]; then
     echo "Pushing $image: $cmd"
     $cmd
   done
+
+  if [ -z "$NO_APPLIANCES" ]; then
+    for image in $appliance_images_to_push; do
+      cmd="$CONTAINER_COMMAND push $image"
+
+      # --format docker is needed for podman to ensure the builds are built with docker v2 manifests
+      if [ "$CONTAINER_COMMAND" == "podman" ]; then
+        cmd+=" --format docker"
+      fi
+
+      echo "Pushing $image: $cmd"
+      $cmd
+    done
+  fi
 fi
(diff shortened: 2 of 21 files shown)


bin/build
```diff
@@ -85,6 +85,17 @@ if [ -z "$NO_OPERATOR" ]; then
   popd
 fi
 
+if APPLIANCES; then
```
bdunne: This is still WIP, but this is my general idea

images/manageiq-appliance/image_build/build
```diff
@@ -0,0 +1,15 @@
+sudo podman pull quay.io/centos-bootc/bootc-image-builder:latest
```
bdunne: I'll move this into `bin/build`, but this is how we build a qcow disk out of the container image

images/manageiq-appliance/Containerfile.fyre
```diff
@@ -0,0 +1,9 @@
+ARG FROM_REPO=docker.io/manageiq
+ARG FROM_TAG=latest
+FROM ${FROM_REPO}/manageiq-appliance:${FROM_TAG}
+
+# For development + Fyre
+RUN echo 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDLi6nZ+ijSs9NkBV0UkzoWyy50s1ZkJKl3u3ujT681d9gh43qeACLRawrqiIjZNWt2s8jM+ef833lV4ss7qoOyJL5QyqFKo0aS2ElPQIYRGtAc4y8feUOH/WexumpLe2f/Ddv0DFYjFlxtPXt42dN1Gq+MKKLyOOui8IvnbBMuNL1hvvW1IoHlAHZgzSElf96rdVw3uwqwQKCl7UKycP6eL1hwgcPL4U2kQS53OcBWuuM2E4dFqW1dz4G29dd2Fa7BEkWUtWt6hFNo/xx5+EvGNY76BxzmU3abuW6Q0XmBWx5Me7jvcABxY/MRvV3I8E+wiZkNRiijwwvibI6CCH8o5LbCwA6LHjxdwBAr+46wsSJomj/zPo0uWCbjQDlVUirSdSgjiZ4nRvPl+Ib9Ar33rhbHexzH2cDWCyUzcAofWpCZPg8T9+7J9Y9TGa2+9NACDlhxGcfPSeAdK81WDZPVyZFe8Jw4x9EbCvimZSkAxijP5tOOYyWrkAh9hl/nyxU= bdunne@li-f0bdb6cc-2eb4-11b2-a85c-8c60b94a3d2a.ibm.com' >> /root/.ssh/authorized_keys && \
```
bdunne: I'll remove this, it was just for early testing.

images/manageiq-appliance/Containerfile.fyre
```diff
@@ -0,0 +1,9 @@
...
+    echo 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC+R80frAPBZ9tuQ5CKIzCKJ5wiRZFvboiCnLgzO+XLaYA8u7/ne2UwaeFPNz/4lDevWtBmYXqC/1UkVOw/i8WlH91Ux7DYjd89xl+znsj8cn3QE38jfgWqoxklLjRBno0Q8ll4MeLrn3qLcJPXh11Yty7HQ/p0XfMgo2j3e7o/Dq+hdJNp7T1olWHRknrkWko/TRZIeuxCxIRS5azBkl0WwgeyoSGMBFxNVta2yDfl0zmIg20of2A/TFnBAQ74C3BRPONkJFMz32hRjxDSj8mbQSqbU2TftyYSGRVqJwOAcKbD7Wyt/m1G24kLivxJgzCLIHAc1kvibjJJDAvH6xP1 fyre_security' >> /root/.ssh/authorized_keys && \
+    su manageiq -c "echo 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABgQDLi6nZ+ijSs9NkBV0UkzoWyy50s1ZkJKl3u3ujT681d9gh43qeACLRawrqiIjZNWt2s8jM+ef833lV4ss7qoOyJL5QyqFKo0aS2ElPQIYRGtAc4y8feUOH/WexumpLe2f/Ddv0DFYjFlxtPXt42dN1Gq+MKKLyOOui8IvnbBMuNL1hvvW1IoHlAHZgzSElf96rdVw3uwqwQKCl7UKycP6eL1hwgcPL4U2kQS53OcBWuuM2E4dFqW1dz4G29dd2Fa7BEkWUtWt6hFNo/xx5+EvGNY76BxzmU3abuW6Q0XmBWx5Me7jvcABxY/MRvV3I8E+wiZkNRiijwwvibI6CCH8o5LbCwA6LHjxdwBAr+46wsSJomj/zPo0uWCbjQDlVUirSdSgjiZ4nRvPl+Ib9Ar33rhbHexzH2cDWCyUzcAofWpCZPg8T9+7J9Y9TGa2+9NACDlhxGcfPSeAdK81WDZPVyZFe8Jw4x9EbCvimZSkAxijP5tOOYyWrkAh9hl/nyxU= bdunne@li-f0bdb6cc-2eb4-11b2-a85c-8c60b94a3d2a.ibm.com' >> ~/.ssh/authorized_keys" && \
```
bdunne: same here

images/manageiq-appliance/Containerfile.gce
```diff
@@ -0,0 +1,30 @@
...
+
+COPY <<EOF /etc/yum.repos.d/google-cloud.repo
+[google-compute-engine]
+name=Google Compute Engine
+baseurl=https://packages.cloud.google.com/yum/repos/google-compute-engine-el10-x86_64-stable
+enabled=1
+gpgcheck=1
+repo_gpgcheck=0
+gpgkey=https://packages.cloud.google.com/yum/doc/rpm-package-key-v10.gpg
+EOF
```
Fryguy, line 13: Not for this PR, but this feels like it could be a committed asset that gets copied instead of inlined. Something like a container-assets-gce/google-cloud.repo and a normal COPY command.

bin/build
```diff
@@ -85,12 +86,37 @@ if [ -z "$NO_OPERATOR" ]; then
   popd
 fi
 
+appliance_images_to_push=""
+if [ -z "$NO_APPLIANCES" ]; then
+  $CONTAINER_COMMAND pull quay.io/centos-bootc/centos-bootc:stream10
```
Fryguy, line 93: This needs some indicator before pulling (in case it gets stuck we know waht's going on)

```suggestion
  echo "Pulling base appliance image"
  $CONTAINER_COMMAND pull quay.io/centos-bootc/centos-bootc:stream10
```

bin/build
```diff
@@ -85,12 +86,37 @@ if [ -z "$NO_OPERATOR" ]; then
...
+fi
+
 if [ -n "$PUSH" ]; then
   push_images="manageiq-base manageiq-base-worker manageiq-orchestrator manageiq-webserver-worker manageiq-ui-worker"
   if [ -z "$NO_OPERATOR" ]; then
     push_images="$push_images manageiq-operator"
   fi
 
+  if [ -z "$NO_APPLIANCES" ]; then
+    push_images="$push_images $appliance_images_to_push"
```
Fryguy:
```suggestion
    push_images+=" $appliance_images_to_push"
```

bin/build
```diff
@@ -85,12 +86,37 @@ if [ -z "$NO_OPERATOR" ]; then
...
+  cmd="$CONTAINER_BUILD_COMMAND . --tag $target_tag"
+  echo "Building $image: $cmd"
+  $cmd
+
+  for containerfile in $(ls Containerfile.*); do
+    target=${containerfile##*.}
+
+    target_tag="$REPO/manageiq-appliance:$TAG-$target"
+    cmd="$CONTAINER_BUILD_COMMAND . -f $containerfile --tag $target_tag"
+    echo "Building $image: $cmd"
```
Fryguy, line 104: Not for this PR, but we may want to consider splitting out appliance image building into a separate bin file so we can run all these in parallel with different build workers. Base image gets built first, then build the rest in parallel.

bin/build
```diff
@@ -85,12 +86,37 @@ if [ -z "$NO_OPERATOR" ]; then
...
+
+  target_tag="$REPO/manageiq-appliance:$TAG"
+  cmd="$CONTAINER_BUILD_COMMAND . --tag $target_tag"
+  echo "Building $image: $cmd"
+  $cmd
+
+  for containerfile in $(ls Containerfile.*); do
+    target=${containerfile##*.}
+
+    target_tag="$REPO/manageiq-appliance:$TAG-$target"
```
Fryguy: So, what this is showing is that we have 

`manageiq-appliance:latest` as the base image, then, for example, `manageiq-appliance:latest-openstack` as the actual appliance. I have 3 concerns with that structure.

1. The base layer should say base, so it's not to be confused as being the actual appliance. I'm thinking `manageiq-appliance-base:latest`
2. The actual appliance images are all manageiq-appliance then differentiated by tag. I think icr.io and possibly other registries do some weird purging rules after a certain number of tags.  Instead I'm wondering if these should be separate images with the tags just being the version...for example, `manageiq-appliance-openstack:latest` and `manageiq-appliance-ovirt:latest`. Aside from the registry concerns I also think of tags as version numbers of a particular image, and in this case these really are different images with different contents.

bin/build
```diff
@@ -103,3 +130,58 @@ if [ -n "$PUSH" ]; then
     $cmd
   done
 fi
+
+# Build the disk images
+if [ -z "$NO_APPLIANCES" ]; then
+  sudo podman pull quay.io/centos-bootc/bootc-image-builder:latest
```
Fryguy, line 134: Also add an echo before pulling here as well

bin/build
```diff
@@ -103,3 +130,58 @@ if [ -n "$PUSH" ]; then
...
+        disk_type="vhd"
+        ;;
+      gce)
+        disk_type="gce"
+        ;;
+      vsphere)
+        disk_type="vmdk"
+        ;;
+      *)
+        # libvirt, openstack, ovirt, vagrant???
```
Fryguy: Did you determine if this was vagrant?

Fryguy: Note that the final name for vagrant is `.box`, so you might need a second variable here.

bin/build
```diff
@@ -103,3 +130,58 @@ if [ -n "$PUSH" ]; then
...
+      # Obtain bearer token
+      echo "Obtaining bearer token..."
+      response=$(curl --fail -s -X "POST" "https://iam.cloud.ibm.com/oidc/token" -H 'Accept: application/json' -H 'Content-Type: application/x-www-form-urlencoded' --data-urlencode "apikey=$IBM_CLOUD_API_KEY" --data-urlencode "response_type=cloud_iam" --data-urlencode "grant_type=urn:ibm:params:oauth:grant-type:apikey")
+      access_token=$(echo $response | jq -r ".access_token")
+
+      # Upload
+      echo
+      echo "Uploading..."
+      bucket="releases-manageiq-org"
+      curl -X "PUT" -H "Authorization: bearer $access_token" -H "Content-Type: application/octet-stream" -H "x-amz-acl: public-read" --data-binary "@$disk_image" "https://s3.us-east.cloud-object-storage.appdomain.cloud/releases-manageiq-org/$destination_name"
```
Fryguy: I don't think it actuall matter, but typically this is capitalized


```suggestion
      curl -X "PUT" -H "Authorization: Bearer $access_token" -H "Content-Type: application/octet-stream" -H "x-amz-acl: public-read" --data-binary "@$disk_image" "https://s3.us-east.cloud-object-storage.appdomain.cloud/releases-manageiq-org/$destination_name"
```

images/manageiq-appliance/config.toml.azure
```diff
@@ -0,0 +1,55 @@
+[[customizations.user]]
+name = "manageiq"
+password = "smartvm"
+groups = ["wheel"]
+
+[customizations.kernel]
```
Fryguy, line 6:
```suggestion
[[customizations.kernel]]
```

images/manageiq-appliance/config.toml.azure
```diff
@@ -0,0 +1,55 @@
...
+name = "lv_var_www_miq_tmp"
+mountpoint = "/var/www/miq_tmp"
+fs_type = "xfs"
+minsize = "10 GiB"
+
+[[customizations.disk.partitions.logical_volumes]]
+name = "lv_home"
+mountpoint = "/var/home"
+fs_type = "xfs"
+minsize = "1 GiB"
```
Fryguy, line 55: Should this one have the swap volume?

images/manageiq-appliance/config.toml.azure
Fryguy: Also all the other ones below?

images/manageiq-appliance/config.toml.openstack
```diff
@@ -0,0 +1,60 @@
+[[customizations.user]]
+name = "manageiq"
+password = "smartvm"
+groups = ["wheel"]
+
+[customizations.kernel]
+append = "crashkernel=auto rhgb quiet console=tty0 console=ttyS0,115200n8"
```
Fryguy, line 7: Why is this one ordered different than the others?

images/manageiq-appliance/Containerfile
```diff
@@ -0,0 +1,67 @@
...
+ARG RELEASE_BUILD
+ARG RPM_PREFIX=manageiq
+
+LABEL name="manageiq-appliance" \
+      summary="ManageIQ appliance bootc image" \
+      vendor="ManageIQ" \
+      url="https://manageiq.org/" \
+      description="ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures."
+
+RUN mkdir -p /var/home
```
Fryguy, line 17: What's this for?

images/manageiq-appliance/Containerfile
```diff
@@ -0,0 +1,67 @@
...
+RUN --mount=type=bind,from=quay.io/manageiq/build_tools:v1,source=/tools,target=/usr/local/bin --mount=type=bind,from=rpms,source=/tmp/rpms,target=/tmp/rpms \
+    dnf config-manager --setopt=tsflags=nodocs --setopt=install_weak_deps=False --save && \
+    dnf -y update && \
+    dnf -y install 'dnf-command(config-manager)' && \
+    miq_enable_epel && \
+    dnf -y install https://rpm.manageiq.org/release/21-uhlmann/el10/noarch/manageiq-release-21.0-1.el10.noarch.rpm && \
+    # Currently the minimum version of qpid-proton in EPEL is 0.40.0, but we're still on 0.37, so pull it from the ManageIQ RPM repo instead https://github.com/ManageIQ/manageiq/blob/b5bf7bee569991d3f1a182a4792fae8cb6265863/Gemfile#L220
+    dnf config-manager --setopt=epel.exclude=*qpid-proton* --save && \
+    if [[ "$RELEASE_BUILD" != "true" ]]; then dnf config-manager --enable manageiq-21-uhlmann-nightly; fi && \
+    if [[ -n "$(ls /tmp/rpms/repodata)" ]]; then miq_prepare_local_yum_repo; fi && \
```
Fryguy: This change doesn't seem right. In the other images, we just checked for the presence of `/tmp/rpms`.

images/manageiq-appliance/Containerfile
```diff
@@ -0,0 +1,67 @@
...
+    dnf -y install \
+      firewalld \
+      fuse-libs \
+      python3-devel \
+      ${RPM_PREFIX}-appliance && \
+    miq_clean_dnf_rpm && \
+    dnf config-manager --set-disabled local-rpm
```
Fryguy: ❤️

images/manageiq-appliance/Containerfile
```diff
@@ -0,0 +1,67 @@
...
+    firewall-offline-cmd --zone=manageiq --add-service=postgresql && \
+    firewall-offline-cmd --zone=manageiq --add-service=dhcpv6-client && \
+    firewall-offline-cmd --zone=manageiq --add-service=cockpit && \
+    systemctl disable ipmi && \
+    systemctl enable cloud-ds-check cockpit.socket evminit memcached miqtop miqvmstat && \
+    ln -sf /dev/null /etc/systemd/system/ctrl-alt-del.target && \
+    sed -i s/root//g /etc/cockpit/disallowed-users && \
+    source /etc/default/evm && \
+    $APPLIANCE_SOURCE_DIRECTORY/manageiq-setup.sh && \
+    echo "smartvm" | passwd -s root && \
```
Fryguy, line 51: Is this needed? I thought the toml files had the user in there already,

bin/build
Fryguy: Another reason is presentation. In the registry UI, typically they only show the image as the top-level thing and statistics are aggregated at the top, then only the last X recent pushed tags are shown. If we split it out, we'd get statistics per image (not that people are pulling the images per se as they'd be pulling the appliances)

images/manageiq-appliance/Containerfile.vagrant
```diff
@@ -0,0 +1,27 @@
...
+    chmod 400 /etc/sudoers.d/vagrant-nopasswd && \
+    mkdir -m 0700 -p ~vagrant/.ssh && \
+    echo "ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEA6NF8iallvQVp22WDkTkyrtvp9eWW6A8YVr+kz4TjGYe7gHzIw+niNltGEFHzD8+v1I2YJ6oXevct1YeS0o9HZyN1Q9qgCgzUFtdOKLv6IedplqoPkcmF0aYet2PkEDo3MlTBckFXPITAMzF8dJSIFo9D8HfdOV0IAdx4O7PtixWKn5y2hMNG0zQPyUecp4pzC6kivAIhyfHilFR61RGL+GPXQ2MWZWFYbAGjyiYJnAmCP3NOTd0jMZEnDkbUvxhMmBYSdETk1rRgm+R4LOzFUGaHqHDLKLX+FIPKcF96hrucXzcWyLbIbEgE98OHlnVYCzRdK8jlqm8tehUc9c9WhQ== vagrant insecure public key" > ~vagrant/.ssh/authorized_keys && \
+    chmod 600 ~vagrant/.ssh/authorized_keys && \
+    chown -R vagrant:vagrant ~vagrant/.ssh/ && \
+    cat << EOM >> ~vagrant/.bashrc
+umask 022
+alias appliance_console='sudo -i appliance_console'
+alias ap=appliance_console
+EOM
```
Fryguy: This feels weird in the RUN command - feels like it should be a copy command in a separate layer (or we should pull it out into a container-assets-vagrant directory or something and use a normal COPY command.)

images/manageiq-appliance/Containerfile.vagrant
```diff
@@ -0,0 +1,27 @@
...
+# Note this is how Vagrant works. You are assumed not to use a Vagrant box on
+# a publicly acessible network.
+COPY <<EOM /etc/sudoers.d/vagrant-nopasswd
+Defaults:vagrant !requiretty
+vagrant ALL=(ALL) NOPASSWD:ALL
+EOM
+
+RUN --mount=type=bind,from=quay.io/manageiq/build_tools:v1,source=/tools,target=/usr/local/bin \
+    dnf -y install cloud-init && \
+    miq_clean_dnf_rpm && \
```
Fryguy: Personally I would split out the rpm install into it's own layer, then the rest of the stuff into a separate layer (or even individual RUN commands. The only reason we slam together all of the dnf stuff is so we can clean it at the end to keep the image layer as tiny as possible.

images/manageiq-appliance/Containerfile.ovirt
```diff
@@ -0,0 +1,8 @@
+ARG FROM_REPO=docker.io/manageiq
+ARG FROM_TAG=latest
+FROM ${FROM_REPO}/manageiq-appliance:${FROM_TAG}
+
+RUN --mount=type=bind,from=quay.io/manageiq/build_tools:v1,source=/tools,target=/usr/local/bin \
+    dnf -y install cloud-init qemu-guest-agent && \
+    miq_clean_dnf_rpm
+
```
Fryguy: Bonus blank line.

images/manageiq-appliance/Containerfile.ovirt
Fryguy, line 1: Since this file is identical to libvirt, does it make sense to just build `FROM ${FROM_REPO}/manageiq-appliance-libvirt:${FROM_TAG}` ?

bin/build
Fryguy: I think this would affect the directory layout to, which might actually be a good thing. You could then just name it Containerfile (and github would color it correctly). Instead of everything in one directory which are kind of not related to one another we'd have multiple directories.

```
images/manageiq-appliance-azure
  config.toml
  Containerfile
images/manageiq-appliance-base
  Containerfile
images/manageiq-appliance-gce
  container-assets
    google-cloud.repo
  config.toml
  Containerfile
```

bin/build
bdunne: They support vdi, vmdk, vhd and hdd.  I moved it to vhd.

images/manageiq-appliance/config.toml.azure
bdunne: No, this one is different.  It's the only one that doesn't.

images/manageiq-appliance/config.toml.openstack
bdunne: The console order matters.

images/manageiq-appliance/Containerfile
bdunne: It's not in the base image for some reason and that's where the home directories get mounted.  So, the image build will fail when we try to write to /home/manageiq later if we don't create it.

bin/build
bdunne: Yeah, I was thinking about that (especially for GHA), but didn't want to introduce that complexity in this PR.

images/manageiq-appliance/Containerfile
Fryguy: Oh is /home essentially a link into some /var/home/\<user\> ?

images/manageiq-appliance/Containerfile
```diff
@@ -0,0 +1,67 @@
...
+      vendor="ManageIQ" \
+      url="https://manageiq.org/" \
+      description="ManageIQ is a management and automation platform for virtual, private, and hybrid cloud infrastructures."
+
+RUN mkdir -p /var/home
+
+RUN --mount=type=bind,from=quay.io/manageiq/build_tools:v1,source=/tools,target=/usr/local/bin --mount=type=bind,from=rpms,source=/tmp/rpms,target=/tmp/rpms \
+    dnf config-manager --setopt=tsflags=nodocs --setopt=install_weak_deps=False --save && \
+    dnf -y update && \
+    dnf -y install 'dnf-command(config-manager)' && \
```
Fryguy, line 22: config-manager is used 2 lines up, so why do we have to install it again?

images/manageiq-appliance/Containerfile.hyperv
```diff
@@ -0,0 +1,8 @@
+ARG FROM_REPO=docker.io/manageiq
+ARG FROM_TAG=latest
+FROM ${FROM_REPO}/manageiq-appliance-base:${FROM_TAG}
+
+RUN --mount=type=bind,from=quay.io/manageiq/build_tools:v1,source=/tools,target=/usr/local/bin \
+    dnf -y install cloud-init hyperv-daemons && \
+    miq_clean_dnf_rpm
+
```
Fryguy: Bonus blank line.

images/manageiq-appliance/Containerfile.ovirt
```diff
@@ -0,0 +1 @@
+Containerfile.libvirt
```
Fryguy, line 1: ~~Missing trailing newline.~~ nvm this is a symlink - didn't realize it in the presentation.

.gitignore
Fryguy, line 1: I think you should also add /output directory to the .gitignore.

images/manageiq-appliance/Containerfile
```diff
@@ -0,0 +1,67 @@
...
+    miq_enable_epel && \
+    dnf -y install https://rpm.manageiq.org/release/21-uhlmann/el10/noarch/manageiq-release-21.0-1.el10.noarch.rpm && \
+    # Currently the minimum version of qpid-proton in EPEL is 0.40.0, but we're still on 0.37, so pull it from the ManageIQ RPM repo instead https://github.com/ManageIQ/manageiq/blob/b5bf7bee569991d3f1a182a4792fae8cb6265863/Gemfile#L220
+    dnf config-manager --setopt=epel.exclude=*qpid-proton* --save && \
+    if [[ "$RELEASE_BUILD" != "true" ]]; then dnf config-manager --enable manageiq-21-uhlmann-nightly; fi && \
+    if [[ -n "$(ls /tmp/rpms)" ]]; then miq_prepare_local_yum_repo; fi && \
+    dnf -y install \
+      firewalld \
+      fuse-libs \
+      python3-devel \
```
Fryguy, line 32: I'm surprised this is needed. I thought the ansible-venv brought this in.

images/manageiq-appliance/Containerfile
```diff
@@ -0,0 +1,67 @@
...
+      ${RPM_PREFIX}-appliance && \
+    miq_clean_dnf_rpm && \
+    if [[ -f /etc/yum.repos.d/local_rpm.repo ]]; then dnf config-manager --set-disabled local-rpm; fi
```
Fryguy, line 35: Minor but we can probably move this to another tool alongside miq_prepare_local_yum_repo, something like miq_disable_local_yum_repo (not for this PR)

images/manageiq-appliance/Containerfile.azure
```diff
@@ -0,0 +1,18 @@
+ARG FROM_REPO=docker.io/manageiq
+ARG FROM_TAG=latest
+FROM ${FROM_REPO}/manageiq-appliance-base:${FROM_TAG}
+
+RUN --mount=type=bind,from=quay.io/manageiq/build_tools:v1,source=/tools,target=/usr/local/bin \
+    dnf -y install WALinuxAgent && \
+    miq_clean_dnf_rpm && \
+    systemctl disable cloud-ds-check && \
```
Fryguy: Similar thing here where we should split dnf install into its own layer.

images/manageiq-appliance/Containerfile.ec2
```diff
@@ -0,0 +1,18 @@
...
+users:
+  - default
+20_miq_user.cfg
+mounts:
+ - [ ephemeral0 ]
+20_miq_ephemeral_mount.cfg
+
+RUN --mount=type=bind,from=quay.io/manageiq/build_tools:v1,source=/tools,target=/usr/local/bin \
+    dnf -y install cloud-init dracut-config-generic && \
+    miq_clean_dnf_rpm && \
```
Fryguy: Similar thing here where we should split dnf install into its own layer.

images/manageiq-appliance/Containerfile.gce
```diff
@@ -0,0 +1,30 @@
...
+
+RUN --mount=type=bind,from=quay.io/manageiq/build_tools:v1,source=/tools,target=/usr/local/bin \
+    systemctl disable cloud-ds-check && \
+    dnf -y install \
+      google-compute-engine \
+      google-osconfig-agent && \
+    miq_clean_dnf_rpm && \
```
Fryguy: Similar thing here where we should split dnf install into its own layer.

images/manageiq-appliance/config.toml.azure
Fryguy: Did this get undone?

bdunne: The brackets are correct the way they are.  The `[[` creates a map which is required for users or disks/partitions/filesystems.  Here's the reference docs:
https://osbuild.org/docs/user-guide/blueprint-reference/#kernel
https://osbuild.org/docs/user-guide/partitioning/

images/manageiq-appliance/Containerfile
bdunne: Yes

images/manageiq-appliance/Containerfile
bdunne: hmm.....

images/manageiq-appliance/Containerfile
bdunne: If that's the case, we can drop it.  I copied it from manageiq-appliance-build

images/manageiq-appliance/Containerfile
bdunne: Or we can change it to be disabled by default and only enable it on the install commands that use it.

images/manageiq-appliance/Containerfile
bdunne: I haven't tried setting root in the TOML users.

bin/build
bdunne: I'm not a big fan of the layout in https://github.com/ManageIQ/manageiq-pods/pull/1378#discussion_r3213238861 since these are all siblings, but very different from the other pods images.  I almost put all of these in a single Containerfile and used something like `podman build --target vmware`, but I thought this would be cleaner.

bdunne:
End-to-end test looks good.

Fryguy (approved):

PR #1404, opened 2026-06-23
Ensure we set user to manageiq

bdunne:
When running in OpenShift, this isn't completely necessary because it will run as the random userid assigned to the namespace. However, when run in other kubernetes like k3s, the container is launched with the default userid (0) which conflicts with the securityContext.runAsNonRoot=true that we set on the deployment.  So, we need to change the default userid to a non-root user.

### DIFF
--- a/images/manageiq-base-worker/Dockerfile
+++ b/images/manageiq-base-worker/Dockerfile
@@ -3,6 +3,8 @@ ARG FROM_TAG=latest
 
 FROM ${FROM_REPO}/manageiq-base:${FROM_TAG} AS vddk
 
+USER 0
+
 COPY container-assets/ /vddk/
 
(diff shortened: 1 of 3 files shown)


Fryguy (approved):
Merging for now - I recommend putting a comment next to the 903 that it's the manageiq user, so we're not guessing.

Fryguy:
Backported to `tal` in commit c85e6c03a9262337131c4223b0ae0164951b2c28.

```text
commit c85e6c03a9262337131c4223b0ae0164951b2c28
Date:   Tue Jun 30 10:43:20 2026 -0400

    Merge pull request #1404 from bdunne/user_903
    
    Ensure we set user to manageiq
    
    (cherry picked from commit f8503e5fc39797da00cc51c113f40db09ce9dd3c)
```

PR #1415, opened 2026-07-23
When running Ingress, need https

kschanrtp:
Root cause:
On k3s (non-OpenShift), the Ingress terminates TLS via the ingress controller, then forwards traffic as plain HTTP to 
the httpd service on port 8080. This is architecturally correct — but the httpd application.conf never sets RequestHeader 
set X_FORWARDED_PROTO 'https' unless the optional ssl_config snippet is loaded (which requires internal certificates). 
Without this header, ManageIQ's application layer sees HTTP, not HTTPS.

Additionally, the Ingress is missing the standard annotation to tell the ingress controller to redirect HTTP->HTTPS and forward the correct 
protocol header. Without nginx.ingress.kubernetes.io/force-ssl-redirect or equivalent Traefik annotations, visiting the plain 
HTTP URL won't redirect to HTTPS.

The fix:
Two files changed, two targeted additions:
httpd_conf.go — RequestHeader set X_FORWARDED_PROTO 'https' added unconditionally to application.conf so ManageIQ always knows the client came in 
               over HTTPS, regardless of environment. Safe on OpenShift because when internal certs exist the ssl_config include sets the exact 
               same value again.
httpd.go — Two nginx annotations added to the Ingress() function:
   ssl-redirect: "true" -> forces HTTP -> HTTPS redirect at the nginx level
   use-forwarded-headers: "true" ->  nginx passes X-Forwarded-Proto to the httpd backend
This function is never called on OpenShift (the controller deletes the Ingress there), so there is zero risk of affecting OpenShift.

### DIFF
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
@@ -115,6 +115,11 @@ func Ingress(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*networkingv1.In
 		if err := controllerutil.SetControllerReference(cr, ingress, scheme); err != nil {
 			return err
 		}
+		if ingress.Annotations == nil {
+			ingress.Annotations = map[string]string{}
+		}
+		ingress.Annotations["nginx.ingress.kubernetes.io/ssl-redirect"] = "true"
+		ingress.Annotations["nginx.ingress.kubernetes.io/use-forwarded-headers"] = "true"
 		if len(ingress.Spec.TLS) == 0 {
 			ingress.Spec.TLS = append(ingress.Spec.TLS, networkingv1.IngressTLS{})
 		}
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go
@@ -39,6 +39,7 @@ Options SymLinksIfOwnerMatch
   ProxyPreserveHost on
   RequestHeader set Host %[1]s
   RequestHeader set X-Forwarded-Host %[1]s
+  RequestHeader set X_FORWARDED_PROTO 'https'
   Header always unset Strict-Transport-Security
   Header always set Strict-Transport-Security "max-age=631138519"
   Header always unset X-Content-Type-Options
--- a/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go
+++ b/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go
@@ -39,7 +39,7 @@ Options SymLinksIfOwnerMatch
   ProxyPreserveHost on
   RequestHeader set Host %[1]s
   RequestHeader set X-Forwarded-Host %[1]s
-  RequestHeader set X_FORWARDED_PROTO 'https'
+  RequestHeader set X-Forwarded-Proto 'https'
   Header always unset Strict-Transport-Security
   Header always set Strict-Transport-Security "max-age=631138519"
   Header always unset X-Content-Type-Options
@@ -606,7 +606,7 @@ func httpdSslConfig() string {
 SSLEngine on
 SSLCertificateFile "/root/server.crt"
 SSLCertificateKeyFile "/root/server.key"
-RequestHeader set X_FORWARDED_PROTO 'https'
+RequestHeader set X-Forwarded-Proto 'https'
 `
 }
 
@@ -615,7 +615,7 @@ func appHttpdSslConfig() string {
 SSLEngine on
 SSLCertificateFile "/etc/pki/tls/certs/server.crt"
 SSLCertificateKeyFile "/etc/pki/tls/private/server.key"
-RequestHeader set X_FORWARDED_PROTO 'https'
+RequestHeader set X-Forwarded-Proto 'https'
 `
 }
 


manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go
```diff
@@ -39,6 +39,7 @@ Options SymLinksIfOwnerMatch
   ProxyPreserveHost on
   RequestHeader set Host %[1]s
   RequestHeader set X-Forwarded-Host %[1]s
+  RequestHeader set X_FORWARDED_PROTO 'https'
```
bdunne: Does this still work when the user is using plain http to communicate inside the cluster? (httpd -> ui-worker)

bdunne: Also, from what I can tell, it should be `X-Forwarded-Proto` https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-Proto

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
```diff
@@ -115,6 +115,11 @@ func Ingress(cr *miqv1alpha1.ManageIQ, scheme *runtime.Scheme) (*networkingv1.In
 		if err := controllerutil.SetControllerReference(cr, ingress, scheme); err != nil {
 			return err
 		}
+		if ingress.Annotations == nil {
+			ingress.Annotations = map[string]string{}
+		}
+		ingress.Annotations["nginx.ingress.kubernetes.io/ssl-redirect"] = "true"
+		ingress.Annotations["nginx.ingress.kubernetes.io/use-forwarded-headers"] = "true"
```
Fryguy, line 122: Does this presume the ingress controller is nginx? What happens if traefik is the controller?

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go
Fryguy: @kschanrtp I think there are other places we need to set this as well for consistency (I could be wrong on this one). cc @jrafanie

Also I agree with `X-Forwarded-Proto` for visual consistency

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
kschanrtp: Correct. This assumes Nginx Ingress Controller. The problem with Ingress is that every implementation has its own Annotations. In this case, I just pick Nginx.

Fryguy: That's annoying. Can we figure out the ingress controller and if/else here for the different ones? (Ideally we'd also support traefik since that's the k3s default I believe.

Note, I'm ok merging this as claming nginx only support on the first pass.

Fryguy:
> Additionally, the Ingress is missing the standard annotation to tell the ingress controller to redirect HTTP->HTTPS and forward the correct
protocol header. Without nginx.ingress.kubernetes.io/force-ssl-redirect or equivalent Traefik annotations, visiting the plain
HTTP URL won't redirect to HTTPS.

Interesting - I thought that was the whole point of HSTS

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
Fryguy: Or, can we just annotate it with all of them? If we don't need to check then don't bother and just list them all

Fryguy: Oh wow - check this out

https://doc.traefik.io/traefik/migrate/nginx-to-traefik/

Seems the nginx annotation work for _BOTH_ nginx and traefik.

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go
kschanrtp: Does it work with plain HTTP (httpd → ui-worker)? — Yes, the proxy destinations (%[4]s://ui:3000) are correctly set to http when no internal TLS certs are present (see [httpd.go:184](vscode-webview://0jafdp8pp450c9eg03jc481hdl7ar197v7653o3n7kilskt22rlp/manageiq-pods/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go:184)), so traffic flows correctly. The X-Forwarded-Proto: https value is still correct on the hardcoded lines because it reflects the external protocol (the incoming TLS connection to httpd at port 8080), not the internal hop.

X-Forwarded-Proto (bdunne's comment) — Correct. All three occurrences of X_FORWARDED_PROTO were using underscores instead of the standard dash-separated form. Fixed across httpdApplicationConf, httpdSslConfig, and appHttpdSslConfig.

Fryguy: There are also files outside of this repo, such as for appliances and in the guides repo - want @jrafanie 's input cause he's going through all that now for a different thing.

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
kschanrtp: Good.

kschanrtp:
> > Additionally, the Ingress is missing the standard annotation to tell the ingress controller to redirect HTTP->HTTPS and forward the correct
> > protocol header. Without nginx.ingress.kubernetes.io/force-ssl-redirect or equivalent Traefik annotations, visiting the plain
> > HTTP URL won't redirect to HTTPS.
> 
> Interesting - I thought that was the whole point of HSTS

Reply:
On the HSTS debate — both sides are partially right:

HSTS (Strict-Transport-Security) only protects browsers that have already visited the site once over HTTPS. On the very first visit, or any client that isn't a browser (curl, API clients, etc.), HSTS does nothing. A plain http:// request would just hang or get a connection refused because the httpd pod only listens on port 8080 (not port 80) — it never terminates plain HTTP at all. So HTTP → HTTPS redirection is the ingress controller's job, and it's already configured.

The annotation nginx.ingress.kubernetes.io/force-ssl-redirect (which the commenter suggested) would enforce the redirect even when no TLS section is configured on the Ingress, but since the TLS section is populated (lines 123–130), ssl-redirect: "true" is sufficient.

No code change is needed for this concern — the annotations are already in place at [httpd.go:121-122](vscode-webview://0jafdp8pp450c9eg03jc481hdl7ar197v7653o3n7kilskt22rlp/manageiq-pods/manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go:121).

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd_conf.go
jrafanie: > @kschanrtp I think there are other places we need to set this as well for consistency (I could be wrong on this one). cc @jrafanie

Repo | File | Notes
-- | -- | --
manageiq-pods | httpd_conf.go | This is the patch being reviewed
product operator | .../httpd.go | Missing — add in global/VirtualHost scope
manageiq-appliance | COPY/etc/httpd/conf.d/manageiq-https-application.conf | Missing — plain flat file, safe to add directly
guides | external_auth/oidc-httpd-configs/application.conf | Missing — required to keep guides in sync

kschanrtp:
@miq-bot add-reviewer @jrafanie

Fryguy:
@kschanrtp Ok thanks for that detail - I didn't realize that's how it worked. That makes me wonder if we have a similar issue on the appliance side, but that is a different discussion.

Fryguy (approved):

Fryguy:
LGTM - will wait on @jrafanie and @bdunne's review.

manageiq-operator/api/v1alpha1/helpers/miq-components/httpd.go
bdunne: From https://doc.traefik.io/traefik-hub/api-gateway/reference/routing/kubernetes/http/routers/ref-ingress-annotations `traefik.ingress.kubernetes.io/router.tls = true`

jrafanie:
We'll do the other repos as followups:

- [x] https://github.com/ManageIQ/manageiq-appliance/pull/414
- [ ] https://github.com/ManageIQ/guides/pull/610

jrafanie (approved):

